PR #63, opened 2020-02-20
Reading immutable history only if necessary

nchamo:
We are now checking if we need to read and parse the immutable history before actually doing so. Previously, we would always read it.

### DIFF
--- a/content/src/service/history/HistoryManagerImpl.ts
+++ b/content/src/service/history/HistoryManagerImpl.ts
@@ -7,20 +7,21 @@ import { happenedBeforeTime, happenedBefore, sortFromOldestToNewest } from "../t
 
 export class HistoryManagerImpl implements HistoryManager {
 
-    /** This history is sorted from newest to oldest */
-    private tempHistory: DeploymentHistory
-
-    private immutableHistorySize: number | undefined
-
-    private constructor(private storage: HistoryStorage, tempHistory: DeploymentHistory) {
-        this.tempHistory = tempHistory
-    }
+    private constructor(private readonly storage: HistoryStorage,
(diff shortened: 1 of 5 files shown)


content/src/service/history/HistoryManagerImpl.ts
```diff
@@ -7,20 +7,21 @@ import { happenedBeforeTime, happenedBefore, sortFromOldestToNewest } from "../t
...
 
-    /** This history is sorted from newest to oldest */
-    private tempHistory: DeploymentHistory
-
-    private immutableHistorySize: number | undefined
-
-    private constructor(private storage: HistoryStorage, tempHistory: DeploymentHistory) {
-        this.tempHistory = tempHistory
-    }
+    private constructor(private readonly storage: HistoryStorage,
```
marcosnc: nit: move the first variable to next line

content/src/service/history/HistoryManagerImpl.ts
```diff
@@ -102,10 +91,16 @@ export class HistoryManagerImpl implements HistoryManager {
...
     }
 
-    private async getImmutableHistory(): Promise<DeploymentHistory> {
-        const immutableHistory = await this.storage.getImmutableHistory()
-        // Sort from newest to oldest
-        return immutableHistory.reverse()
+
+    /** Before loading the immutable history, we check if it is in fact necessary */
+    private async retrieveNecessaryHistory(from: Timestamp | undefined): Promise<DeploymentHistory> {
+        if (!from || from <= this.immutableTime) {
```
marcosnc: I like it!
I think we can improve it, even more, to avoid reading the immutable history when history is called without parameters.

PR #73, opened 2020-02-26
Changes to sync time and caching

nchamo:
**Syncing**
We are now syncing every 2 minutes. It used to be evert 45 seconds
**Caching**
1. We are now caching more scene entities.
2. Caches sizes can be set via environment variables
3. We now have a centralised place (the `CacheManager`) that generates all caches

### DIFF
--- a/content/src/Environment.ts
+++ b/content/src/Environment.ts
@@ -25,6 +25,7 @@ import { AuthenticatorFactory } from "./service/auth/AuthenticatorFactory";
 import { AccessCheckerImplFactory } from "./service/access/AccessCheckerImplFactory";
 import { FailedDeploymentsManagerFactory } from "./service/errors/FailedDeploymentsManagerFactory";
 import { FetchHelperFactory } from "./helpers/FetchHelperFactory";
+import { CacheManagerFactory } from "./service/caching/CacheManagerFactory";
 
 export const CURRENT_CONTENT_VERSION: EntityVersion = EntityVersion.V3
 const DEFAULT_STORAGE_ROOT_FOLDER = "storage"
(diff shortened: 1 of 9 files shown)


content/src/service/ServiceImpl.ts
```diff
@@ -39,7 +39,7 @@ export class ServiceImpl implements MetaverseContentService, TimeKeepingService,
         private failedDeploymentsManager: FailedDeploymentsManager,
         private ignoreValidationErrors: boolean,
         private network: string) {
-        this.entities = Cache.withCalculation((entityId: EntityId) => this.storage.getEntityById(entityId), 1000)
+        this.entities = Cache.withCalculation((entityId: EntityId) => this.storage.getEntityById(entityId), 5000)
```
marcosnc: Should we add a config parameter here?

content/src/Environment.ts
```diff
@@ -180,11 +183,13 @@ export class EnvironmentBuilder {
         this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.FILE_DOWNLOAD_REQUEST_TIMEOUT, () => process.env.FILE_DOWNLOAD_REQUEST_TIMEOUT ?? ms('5m'))
         this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.USE_COMPRESSION_MIDDLEWARE, () => process.env.USE_COMPRESSION_MIDDLEWARE === "true");
         this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.PERFORM_MULTI_SERVER_ONBOARDING, () => process.env.PERFORM_MULTI_SERVER_ONBOARDING === "true");
+        this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.CACHE_SIZES               , () => new Map(Object.entries(process.env).filter(([name,]) => name.startsWith("CACHE"))));
```
marcosnc: nice!

PR #75, opened 2020-02-26
Now the content server can support name changes without problems

nchamo:
This PR adds to content servers the ability to support name changes on other servers. It does so by changing two things:
1. When syncing with the DAO (every 5 min by default), if we detect a new name, we update it
2. When syncing with other servers, if there is no new deployment, we won't update it's estimated immutable time if the name is different from the expected one.

### DIFF
--- a/content/src/service/synchronization/ContentCluster.ts
+++ b/content/src/service/synchronization/ContentCluster.ts
@@ -120,7 +120,7 @@ export class ContentCluster {
                         newClient = getClient(this.fetchHelper, address, newName, previousClient.getEstimatedLocalImmutableTime())
                     } else if (previousClient.getName() !== newName) {
                         // Update known name to new one
-                        ContentCluster.LOGGER.warn(`Server's name changed on ${address}. It was ${previousClient.getName()} and know it is ${newName}.`)
+                        ContentCluster.LOGGER.warn(`Server's name changed on ${address}. It was ${previousClient.getName()} and now it is ${newName}.`)
                         newClient = getClient(this.fetchHelper, address, newName, previousClient.getEstimatedLocalImmutableTime())
                     }
                 } else {
--- a/content/src/service/synchronization/ContentCluster.ts
+++ b/content/src/service/synchronization/ContentCluster.ts
@@ -109,7 +109,7 @@ export class ContentCluster {
                 let newClient: ContentServerClient | undefined
                 // Check if we already knew the server
                 const previousClient: ContentServerClient | undefined = this.serversInDAO.get(address);
-                if (previousClient && previousClient.getName() !== UNREACHABLE) {
+                if (previousClient && previousClient.getConnectionState() !== ConnectionState.NEVER_REACHED) {
                     if (newName === UNREACHABLE) {
                         // Create redirect client
                         newClient = getRedirectClient(this, previousClient.getName(), previousClient.getEstimatedLocalImmutableTime())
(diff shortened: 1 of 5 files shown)


content/src/service/synchronization/ContentCluster.ts
```diff
@@ -109,7 +109,7 @@ export class ContentCluster {
                 let newClient: ContentServerClient | undefined
                 // Check if we already knew the server
                 const previousClient: ContentServerClient | undefined = this.serversInDAO.get(address);
-                if (previousClient && previousClient.getName() !== UNREACHABLE) {
```
nchamo: No change in behaviour, but it is clearer now

content/src/service/synchronization/SynchronizationManager.ts
```diff
@@ -95,8 +100,10 @@ export class ClusterSynchronizationManager implements SynchronizationManager {
             this.synchronizationState = SynchronizationState.FAILED_TO_SYNC;
             ClusterSynchronizationManager.LOGGER.warn(`Failed to sync with servers. Reason:\n${error}`)
         } finally {
-            // Set the timeout again
```
nchamo: Without this flag, the sync might never stop

content/src/service/synchronization/ContentCluster.ts
```diff
@@ -118,6 +118,10 @@ export class ContentCluster {
                         // Create new client
                         ContentCluster.LOGGER.info(`Could re-connect to server ${newName} on ${address}`)
                         newClient = getClient(this.fetchHelper, address, newName, previousClient.getEstimatedLocalImmutableTime())
+                    } else if (previousClient.getName() !== newName) {
+                        // Update known name to new one
+                        ContentCluster.LOGGER.warn(`Server's name changed on ${address}. It was ${previousClient.getName()} and know it is ${newName}.`)
```
marcosnc: nit: know -> now

content/src/service/synchronization/ContentCluster.ts
```diff
@@ -118,6 +118,10 @@ export class ContentCluster {
...
+                        newClient = getClient(this.fetchHelper, address, newName, previousClient.getEstimatedLocalImmutableTime())
```
marcosnc: should we stop de previousClient? is that already being done?

content/src/service/synchronization/SynchronizationManager.ts
```diff
@@ -95,8 +100,10 @@ export class ClusterSynchronizationManager implements SynchronizationManager {
...
-            this.syncWithNodesTimeout = setTimeout(() => this.syncWithServers(), this.timeBetweenSyncs)
+            if (!this.stopping) {
```
marcosnc: Maybe you can also add this checking at the beginning of syncWithServers.

content/src/service/synchronization/clients/contentserver/ActiveContentServerClient.ts
```diff
@@ -82,8 +82,16 @@ class ActiveContentServerClient extends ContentServerClient {
...
+     * Check the current time on the status and report it. However, if the name is not as expected, then
+     * we shouldn't update the estimated immutable time. On the next DAO sync, the name will be updated correctly
+     */
     private async getCurrentTimestamp(): Promise<Timestamp> {
-        const { currentTime } = await this.fetchHelper.fetchJson(`${this.address}/status`)
-        return currentTime;
+        const { currentTime, name } = await this.getStatus()
+        if (name === this.getName()) {
+            return currentTime;
+        } else {
```
marcosnc: nit: else is redundant

content/test/integration/syncronization/name-change-handling.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+        server1 = await buildServer("Server1_", 6060, SYNC_INTERVAL, dao)
+        server2 = await buildServer("Server2_", 7070, SYNC_INTERVAL, dao)
+    })
+
+    afterEach(async () => {
+        await stopServers(server1, server2)
+        deleteServerStorage(server1, server2)
+    })
+
+    it('When a node starts, it get all the previous history', async () => {
```
marcosnc: should we add a test to validate the recurrent synchronization too? I'm assuming this test only checks the startup.

content/src/service/synchronization/SynchronizationManager.ts
nchamo: It shouldn't be necessary, since we are clearing the timeout.

content/src/service/synchronization/clients/contentserver/ActiveContentServerClient.ts
nchamo: You know I like my elses haha

content/src/service/synchronization/ContentCluster.ts
nchamo: No need to stop them, they don't do nothing by themselves

PR #79, opened 2020-02-27
Add many content server tests

nchamo:
Changes:
A. We made TTL backwards time configurable, but not by env variable. We did this so that we can modify it in tests
B. We added a few tests:
1. `When a server finds a new deployment with already known content, it can still deploy it successfully`
2. `When a user tries to fix an entity, it doesn't matter if there is already a newer entity deployed`
3. `When a pointer is denylisted, then the history is empty`
4. `When a pointer is not denylisted, then it reports the history correctly`


marcosnc:
You also packaged all the service validations inside a `Validations` class.

PR #80, opened 2020-03-02
Feat: Adding lambdas app metrics

mgagliardo:
Adding app metrics to Lambdas. Relates to https://github.com/decentraland/devops/issues/270


lambdas/src/Environment.ts
```diff
@@ -50,7 +50,8 @@ export const enum EnvironmentConfig {
   ENS_OWNER_PROVIDER_URL,
   COMMIT_HASH,
   USE_COMPRESSION_MIDDLEWARE,
-  LOG_LEVEL
+  LOG_LEVEL,
+  METRICS
```
nchamo: You need to add a default value to metrics, like it is done below for the other configuration. Otherwise, the env variable won't be set on this new configuration

nchamo:
Implemented by https://github.com/decentraland/catalyst/pull/261

PR #81, opened 2020-03-03
feat: use timestamp when checking for access

marcosnc:


content/test/unit/service/access/AccessCheckerImpl.spec.ts
```diff
@@ -1,47 +1,56 @@
...
 
         expect(errors.length).toBe(0)
     })
 
+    it(`When access URL is wrong it reports an error`, async () => {
+        const accessChecker = new AccessCheckerImpl(new ContentAuthenticator(), "Wrong URL");
+
+        const errors = await accessChecker.hasAccess(EntityType.SCENE, ["102,4"], Date.now(), "Some-address-without-permissions");
+
+        expect(errors.length).toBe(1)
```
nchamo: Could you assert the error message?

content/src/service/validations/Validations.ts
```diff
@@ -119,9 +119,9 @@ export class ValidatorInstance {
     }
 
     /** Validate that the pointers are valid, and that the Ethereum address has write access to them */
-    async validateAccess(entityType: EntityType, pointers: Pointer[], ethAddress: EthAddress, validationContext: ValidationContext): Promise<void> {
+    async validateAccess(entityType: EntityType, pointers: Pointer[], timestamp: Timestamp, ethAddress: EthAddress, validationContext: ValidationContext): Promise<void> {
```
nchamo: Nit: we could pass the entire entity, instead of passing many of its properties separately

content/src/Environment.ts
```diff
@@ -31,11 +31,10 @@ import { ValidationsFactory } from "./service/validations/ValidationsFactory";
 export const CURRENT_CONTENT_VERSION: EntityVersion = EntityVersion.V3
 const DEFAULT_STORAGE_ROOT_FOLDER = "storage"
 const DEFAULT_SERVER_PORT = 6969
-const DEFAULT_DCL_API_URL = "https://api.decentraland.zone/v1"
 export const DEFAULT_ETH_NETWORK = "ropsten"
+export const DEFAULT_DCL_PARCEL_ACCESS_URL = 'https://api.thegraph.com/subgraphs/name/nicosantangelo/watchtower'
```
nchamo: A few things about this change:
1. We need to update the .env on all servers. Ours (prod & dev) and the ones outside dcl
2. Is there a url that listens to ropsten? If so, should we use that as the default (as we did previously)?

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,90 +1,399 @@
...
+         *   - owner
+         *   - operator
+         *   - update operator
+         * at that time
+         */
+        const parcel = await this.getParcel(x, y, timestamp)
 
-        if (pointers.length!=1) {
-            errors.push(`Only one pointer is allowed when you create a Profile. Received: ${pointers}`)
+        if (parcel.estates?.length > 0 && parcel.estates[0].estateId) {
```
nchamo: What does `parcel.estates?.length > 0` return if `parcel.estates` is undefined?

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,90 +1,399 @@
...
+            // The parcel belongs to an estate
+            return this.isEstateUpdateAuthorized(
```
nchamo: It this ok? If the estate validations fails, shouldn't we still check the parcel access?

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,90 +1,399 @@
...
+            x, y,
+            timestamp: Math.floor(timestamp / 1000) // UNIX
+        }
+
+        try {
+            const response = await this.queryGraph<{ parcels: Parcel[] }>(
+                query,
+                variables
+            )
+            return response.parcels[0]
```
nchamo: Are we sure this will always return at least one element? If not, then this might return `undefined`

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,90 +1,399 @@
...
+            estateId,
+            timestamp: Math.floor(timestamp / 1000) // UNIX
+        }
+
+        try {
+            const response = await this.queryGraph<{ estates: Estate[] }>(
+                query,
+                variables
+            )
+            return response.estates[0]
```
nchamo: Same as before

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,90 +1,399 @@
...
-        if (pointer.startsWith("default")) {
-            if (!this.authenticator.isAddressOwnedByDecentraland(ethAddress)) {
-                errors.push(`Only Decentraland can add or modify default profiles`)
+        ethAddress = ethAddress.toLowerCase()
+        if (firstLevelAuthorities.includes(ethAddress)) {
+            return true
+        }
+
+        /* You also get access if you received:
+         *   - an auhtorization with isApproved and type Operator
```
nchamo: Typo "auhtorization"

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,90 +1,399 @@
...
+            .filter(addressSnapshot => addressSnapshot.address)
+            .map(addressSnapshot => addressSnapshot.address.toLowerCase())
+
+        ethAddress = ethAddress.toLowerCase()
+        if (firstLevelAuthorities.includes(ethAddress)) {
+            return true
+        }
+
+        /* You also get access if you received:
+         *   - an auhtorization with isApproved and type Operator
```
nchamo: Typo "auhtorization"

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,90 +1,399 @@
...
+            authorization => authorization.type === 'ApprovalForAll')
+
+        if (firstOperatorAuthorization?.isApproved || firstApprovalForAllAuthorization?.isApproved) {
+            return true
+        }
+
+        return false
+    }
+
+    private async isEstateUpdateAuthorized(
```
nchamo: I see a lot of repeated code, do you think there is an easy way to avoid it?

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,90 +1,399 @@
...
+                authorizations: Authorization[]
+            }>(query, variables)
+            return response.authorizations
+        } catch (error) {
+            AccessCheckerImpl.LOGGER.error(`Error fetching authorizations for ${owner}`, error)
+            throw error
+        }
+    }
+
+    // TODO: Move this to FetchHelper5
```
nchamo: I think there is an extra "5" in there

content/src/service/access/AccessCheckerImpl.ts
marcosnc: It returns `false`.

content/src/service/access/AccessCheckerImpl.ts
marcosnc: No, because if the parcel belongs to an estate, all the checking should be done based on the estate, not the parcel.

content/test/unit/service/access/AccessCheckerImpl.spec.ts
marcosnc: Done!

content/src/service/access/AccessCheckerImpl.ts
marcosnc: We're changing the logic here, granting the permission if it's provided by the estate but also if it's provided by the parcel.

content/src/service/access/AccessCheckerImpl.ts
marcosnc: fixed!

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -44,9 +44,12 @@ export class AccessCheckerImpl implements AccessChecker {
         return errors
     }
 
+    private SCENE_LOOKBACK_TIME = 5 * 60 * 1000;
```
nchamo: Me have a library called `ms`, so you can do  `SCENE_LOOKBACK_TIME = ms('5m');`

Also, maybe set to static?

content/test/integration/check-failed-deployments.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+import { AuditInfo } from "@katalyst/content/service/audit/Audit"
+import { EthAddress, AuthChain, ValidationResult } from "dcl-crypto"
+import { Timestamp } from "@katalyst/content/service/time/TimeSorting"
+import { AccessCheckerImpl } from "@katalyst/content/service/access/AccessCheckerImpl"
+import { ContentAuthenticator } from "@katalyst/content/service/auth/Authenticator"
+import { DEFAULT_DCL_PARCEL_ACCESS_URL } from "@katalyst/content/Environment"
+import fs from 'fs';
+import { httpProviderForNetwork } from "../../../contracts/utils"
+
+describe("Failed Deployments validations.", () => {
```
nchamo: Maybe we could separate from the automatic tests? Maybe move them to a new folder called tools/tooling or something similar?

failed-deployments-analysis-2.txt
```diff
@@ -0,0 +1,472 @@
+=> 1 of 401: #Fixed# QmPk63861E1DUwBrgaV4guQACrnhGM7NnyZGd7VHn2m8q5 0x8445165437cb63affb2f143378d0ec63e0249582 1583423166422 [131,18]
```
nchamo: Maybe we should avoid adding this?

failed-deployments-analysis-invalid.txt
```diff
@@ -0,0 +1,169 @@
+=> 1 of 401: Invalid QmPk63861E1DUwBrgaV4guQACrnhGM7NnyZGd7VHn2m8q5 0x8445165437cb63affb2f143378d0ec63e0249582 1583423166422 [131,18]
```
nchamo: Maybe we should avoid adding this?

failed-deployments-analysis.txt
```diff
@@ -0,0 +1,425 @@
+exec ${PAGER:-/usr/bin/less} "$0" || exit 1
```
nchamo: Maybe we should avoid adding this?

PR #83, opened 2020-03-09
Adding retries to many different situations

nchamo:


content/src/helpers/FetchHelper.ts
```diff
@@ -40,4 +40,18 @@ export class FetchHelper {
         }
     }
 
+}
+
+export async function retry<T>(execution: () => Promise<T>, attempts: number, description: string, waitTime: string = '1s'): Promise<T> {
+    while (attempts > 0) {
+        try {
+            return await execution()
+        } catch (error) { }
```
marcosnc: I think it's important to report the cause of the error, maybe at the last attempt.

content/src/service/synchronization/ClusterUtils.ts
```diff
@@ -1,38 +1,26 @@
...
-        // Wait a little before retrying
-        retries--;
-        if (retries >= 0) {
-            await delay(ms("1s"))
-            LOGGER.info("All calls to other servers failed. Going to retry.")
-        }
-    }
-
-    throw new Error(`Tried to execute request on all servers on the cluster, but they all failed`)
+        throw new Error(`Tried to ${description} on all servers on the cluster, but they all failed`)
```
marcosnc: Same here, we are not informing what's the error's cause.

content/src/service/synchronization/EventDeployer.ts
```diff
@@ -186,9 +186,9 @@ export class EventDeployer {
     }
 
     /** Execute an operation on the cluster, but return 'undefined' if it fails */
-    private async tryOnClusterOrUndefined<T>(execution: (server: ContentServerClient) => Promise<T>, cluster: ContentCluster, options?: { retries?: number, preferred?: ContentServerClient}): Promise<T | undefined> {
+    private async tryOnClusterOrUndefined<T>(execution: (server: ContentServerClient) => Promise<T>, cluster: ContentCluster, description: string, options?: { retries?: number, preferred?: ContentServerClient}): Promise<T | undefined> {
         try {
-            return await tryOnCluster(execution, cluster, options)
+            return await tryOnCluster(execution, cluster, description, options)
         } catch (error) {
             return undefined
```
marcosnc: Should we inform the error here?

content/src/helpers/FetchHelper.ts
nchamo: Done!

content/src/service/synchronization/ClusterUtils.ts
nchamo: We are now logging the error. Since we are trying on many servers, I'm not sure we should so more than logging.

content/src/service/synchronization/EventDeployer.ts
nchamo: There is  no need to do so, since the `tryOnCluster` already does that. Here, we are just changing from an error to an `undefined`

PR #88, opened 2020-03-16
feat: only servers on the DAO can deploy

nchamo:
With this change, only servers that are whitelisted by the DAO can accept deployments. To do this we:

1. Added the concept of a challenge. To figure out their own identity, servers generate a challenge (random text) when they start. Then, they will ask for every server's challenge text and check who matches their own text.
2. Added a flag called `ALLOW_DEPLOYMENTS_FOR_TESTING` to allow deployments even though the server was not whitelisted by the DAO. When this is done, the deployments will be saved on the history with the serverName: `"NOT_IN_DAO"`
3. We also added a retry mechanism to avoid the current situation where nginx would start after the content server, so the server couldn't reach itself


content/src/service/ServiceImpl.ts
```diff
@@ -81,12 +84,18 @@ export class ServiceImpl implements MetaverseContentService, TimeKeepingService,
     }
 
     deployEntity(files: ContentFile[], entityId: EntityId, auditInfo: AuditInfo, origin: string): Promise<Timestamp> {
-        return this.deployInternal(files, entityId, auditInfo, this.nameKeeper.getServerName(), ValidationContext.LOCAL, origin)
+        if (!this.allowDeploymentsWhenNotInDAO && !this.identityProvider.getOwnIdentity()) {
```
marcosnc: I'm not sure if it's clear enough that having something in `this.identityProvider.getOwnIdentity()` means you're inside the DAO.

content/src/service/ServiceImpl.ts
```diff
@@ -81,12 +84,18 @@ export class ServiceImpl implements MetaverseContentService, TimeKeepingService,
...
+            throw new Error(`Deployments are not allow since server is not in DAO`)
+        }
+        return this.deployInternal(files, entityId, auditInfo, this.getOwnName(), ValidationContext.LOCAL, origin)
     }
 
     deployToFix(files: ContentFile[], entityId: EntityId, auditInfo: AuditInfo, origin: string): Promise<Timestamp> {
+        if (!this.allowDeploymentsWhenNotInDAO && !this.identityProvider.getOwnIdentity()) {
```
marcosnc: I would prefer to allow fixes even when you're outside the DAO.

content/src/service/ServiceImpl.ts
```diff
@@ -283,6 +292,10 @@ export class ServiceImpl implements MetaverseContentService, TimeKeepingService,
         return !!entityIdDeployed.get(entityId)
     }
 
+    private getOwnName(): ServerName {
+        return this.identityProvider.getOwnIdentity()?.name ?? "NOT_IN_DAO"
```
marcosnc: nit: add a constant for this, please

content/src/service/synchronization/ContentCluster.ts
```diff
@@ -175,22 +186,50 @@ export class ContentCluster {
...
-            const serversWithMyName = serverNames.filter(({ name }) => name == this.nameKeeper.getServerName())
-
-            if (serversWithMyName.length > 1) {
-                ContentCluster.LOGGER.warn(`Expected to find only one server with my name '${this.nameKeeper.getServerName()}', but found ${serversWithMyName.length}`)
-            } else {
-                this.myIdentity = serversWithMyName[0]?.metadata
+            // Fetch server list from the DAO
+            const allServers: Set<ServerMetadata> = await this.dao.getAllContentServers()
+            const serversByAddresses: Map<ServerAddress, ServerMetadata> = new Map(Array.from(allServers).map(metadata => [metadata.address, metadata]))
+            const challengesByAddress: Map<ServerAddress, ChallengeText> = new Map()
```
marcosnc: I think this can be class cache. Challenges should not change, right?

content/src/service/synchronization/ContentCluster.ts
```diff
@@ -175,22 +186,50 @@ export class ContentCluster {
...
-
-            // Filter out other servers
-            const serversWithMyName = serverNames.filter(({ name }) => name == this.nameKeeper.getServerName())
-
-            if (serversWithMyName.length > 1) {
-                ContentCluster.LOGGER.warn(`Expected to find only one server with my name '${this.nameKeeper.getServerName()}', but found ${serversWithMyName.length}`)
-            } else {
-                this.myIdentity = serversWithMyName[0]?.metadata
+            // Fetch server list from the DAO
+            const allServers: Set<ServerMetadata> = await this.dao.getAllContentServers()
```
marcosnc: We should try to avoid execute this method as much as possible. Each call consume resources that we are paying.
Maybe we should add a cache for this and ensure to refresh it every 5 minutes, so we can call this method without problems.

content/src/service/synchronization/ContentCluster.ts
nchamo: Yes, they change when a server is restarted. Since it's only an HTTP call, I think we can make it again, right? Or are you worried about something in particular?

PR #92, opened 2020-03-18
fix: now we check if the address has access now or if it did 5 min ago

nchamo:
Now, when we validate for scene access, we check if the address has access to the parcel right now. It it doesn't, then we check 5 min into the past.

This is because someone might lose access to a parcel and still be able to deploy for 5 min, since the blockchain change hasn't propagated yet.


content/test/integration/service/history/client/HistoryClient.spec.ts
```diff
@@ -1,11 +1,11 @@
-import { TestServer } from "../../TestServer"
+import { TestServer } from "../../../TestServer"
```
nchamo: Just moved this test, so it was in the same folder as the code it is testing

content/test/integration/service/access/AccessCheckerImpl.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { AccessCheckerImpl } from "@katalyst/content/service/access/AccessCheckerImpl";
+import { EntityType } from "@katalyst/content/service/Entity";
+import { ContentAuthenticator } from "@katalyst/content/service/auth/Authenticator";
+
+describe("Integration - AccessCheckerImpl", function () {
```
nchamo: This test was connection to the internet, so I moved it out of the unit tests

PR #93, opened 2020-03-18
Replace WebSockets with HTTP to connect to infura

nchamo:
Stop using WebSockets to connect to infura, and use HTTP instead.

I manually verified that when the connections goes down, and then goes up again, the server can still sync correctly

### DIFF
--- a/contracts/utils.ts
+++ b/contracts/utils.ts
@@ -1,6 +1,6 @@
 import { Address } from "web3x/address";
 import { Eth } from "web3x/eth";
-import { WebsocketProvider, HttpProvider } from "web3x/providers";
+import { HttpProvider } from "web3x/providers";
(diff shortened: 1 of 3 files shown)


contracts/utils.ts
```diff
@@ -29,7 +29,8 @@ export const networks = {
 export function handlerForNetwork(networkKey: string, contractKey: string) {
   try {
     const network = networks[networkKey];
```
marcosnc: These 3 lines are exactly the same as in `httpProviderForNetwork`, may we reuse them or delete one of these methods?

moliva: I agree, let's reuse `httpProviderForNetwork` here to avoid duplication.

nchamo: Done

PR #94, opened 2020-03-18
Avoid so many calls to Infura

nchamo:
### Problem
Every 5 min the content server syncs with the DAO, to retrieve the server list.
Before this change, each server would:
1. Get the amount of servers on the list
2. For each index on the list, it would retrieve the entry id
3. With each id, it would retrieve the data (owner ethAddress and domain)

Now, this implies that every 5 min, each content server would make `1 + 2N` calls to Infura, even if there were no changes to the list (with N being the number of servers on the DAO).

### Solution
Since the id is unique (this means that even if a domain is removed and re-added its id will change), we will avoid fetching the data from already known servers. This should reduce the amount of calls almost by half. We would go from `1 + 2N` to `1 + N` calls.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,13 @@ jobs:
       - run:
           name: Run tests
           command: npx bazel test //lambdas:unit_test --test_output=all
+  build-commons:
+    <<: *base_env
+    steps:
+      - <<: *attach_root
+      - run:
+          name: Run unit tests
+          command: npx bazel test //commons/servers:unit_test --test_output=all
 
   publish-peer:
     <<: *base_env
@@ -174,6 +181,10 @@ workflows:
           requires:
             - checkout
       - build-lambdas:
+          <<: *all_branches_and_tags
+          requires:
+            - checkout
+      - build-commons:
           <<: *all_branches_and_tags
           requires:
             - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,8 +98,11 @@ jobs:
     steps:
       - <<: *attach_root
       - run:
-          name: Run unit tests
+          name: Run servers tests
           command: npx bazel test //commons/servers:unit_test --test_output=all
+      - run:
+          name: Run utils tests
+          command: npx bazel test //commons/utils:unit_test --test_output=all
 
   publish-peer:
     <<: *base_env
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -175,11 +175,11 @@ workflows:
       - build-lighthouse:
           <<: *all_branches_and_tags
           requires:
-            - checkout
+            - build-commons
       - build-content:
           <<: *all_branches_and_tags
           requires:
-            - checkout
+            - build-commons
       - build-lambdas:
           <<: *all_branches_and_tags
           requires:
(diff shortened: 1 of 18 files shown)


.circleci/config.yml
```diff
@@ -93,6 +93,13 @@ jobs:
       - run:
           name: Run tests
           command: npx bazel test //lambdas:unit_test --test_output=all
+  build-commons:
```
nchamo: Maybe others should depend on build-commons?

commons/servers/DAOContract.ts
```diff
@@ -0,0 +1,38 @@
+import { handlerForNetwork } from "decentraland-katalyst-contracts/utils";
+import { Catalyst } from "decentraland-katalyst-contracts/Catalyst";
+import { EthAddress } from "dcl-crypto";
+
+export class DAOContract {
```
nchamo: Maybe we could move this to the `contracts` module. Do you think it belongs there?

pablitar:
This sounds a little hacky, pretty coupled to a particular implementation of the contract.

What are the odds of the contract changing, turning this feature into a bug?

.circleci/config.yml
moliva: Let's add this job as part of the requirements of `accept-publish-docker` in line 206 to make sure tests are running properly for this package.

commons/servers/DAOClient.ts
```diff
@@ -21,31 +15,54 @@ export class DAOClient {
...
+      // Check if id is known
+      let metadata = this.servers.get(id)
 
-        if (domain.startsWith("http://")) {
-          console.warn(`Catalyst node domain using http protocol, skipping ${domain}`);
-          continue;
-        }
+      // If it isn't known, then calculate it
+      if (!metadata) {
+        const data = await this.contract.getServerData(id)
```
moliva: What if this fails? Do we want to skip the update as a whole until the next refresh or maybe continue with the next server?

commons/servers/DAOClient.ts
```diff
@@ -21,31 +15,54 @@ export class DAOClient {
...
+        metadata = this.toMetadata(data)
+      }
 
-        if (!domain.startsWith("https://")) {
-          domain = "https://" + domain;
-        }
+      // If metadata is defined, then store it
+      if (metadata) {
```
moliva: Could this be undefined/null? Let's remove it if not.

commons/servers/DAOClient.ts
nchamo: We should skip the update as a whole. Otherwise, we might think that the server was removed from the DAO

commons/servers/DAOClient.ts
nchamo: It can, that's why we are checking

.circleci/config.yml
pablitar: I just realized we need to add the same thing for commons/utils. I'll do it in #95

pablitar: Or... maybe we could include it as a step of this build. What do you think?

.circleci/config.yml
```diff
@@ -93,6 +93,13 @@ jobs:
...
+    <<: *base_env
+    steps:
+      - <<: *attach_root
+      - run:
+          name: Run unit tests
+          command: npx bazel test //commons/servers:unit_test --test_output=all
```
pablitar:
```suggestion
          name: Run unit tests
          command: npx bazel test //commons/servers:unit_test //commons/utils:unit_test --test_output=all
```

.circleci/config.yml
nchamo: Done :)

PR #105, opened 2020-03-27
Fix: allow legacy deployments in the past

nchamo:
We are making 3 small fixes in the same PR:
1. When we retry an action and fail, we were incorrectly logging that 0 attempts were made
2. We are allowing the "future" deployments 15 min into the future (it was 5 min previously)
3. We didn't allow legacy entities to be deployed if there was an entity with a newer version deployed. The problem is, that during sync, we should still allow these deployments. We fixed it

### DIFF
--- a/content/src/helpers/FetchHelper.ts
+++ b/content/src/helpers/FetchHelper.ts
@@ -43,6 +43,7 @@ export class FetchHelper {
 }
 
 export async function retry<T>(execution: () => Promise<T>, attempts: number, description: string, waitTime: string = '1s'): Promise<T> {
+    const initialAttempts = attempts
     while (attempts > 0) {
         try {
             return await execution()
@@ -52,7 +53,7 @@ export async function retry<T>(execution: () => Promise<T>, attempts: number, de
                 await delay(ms(waitTime))
                 LOGGER.info(`Failed to ${description}. Still have ${attempts} attempt/s left. Will try again in ${waitTime}`)
             } else {
-                LOGGER.warn(`Failed to ${description} after ${attempts} attempts. Error was ${error}`)
+                LOGGER.warn(`Failed to ${description} after ${initialAttempts} attempts. Error was ${error}`)
                 throw error
             }
         }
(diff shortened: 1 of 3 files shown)


content/src/service/validations/Validations.ts
```diff
@@ -127,7 +127,7 @@ export class ValidatorInstance {
     }
 
     /** Validate that the deployment is recent */
-    private static REQUEST_TTL_FORWARDS: number = ms('5m') // 5 minutes
+    private static REQUEST_TTL_FORWARDS: number = ms('15m')
```
nchamo: Change #2

content/src/helpers/FetchHelper.ts
```diff
@@ -52,7 +53,7 @@ export async function retry<T>(execution: () => Promise<T>, attempts: number, de
                 await delay(ms(waitTime))
                 LOGGER.info(`Failed to ${description}. Still have ${attempts} attempt/s left. Will try again in ${waitTime}`)
             } else {
-                LOGGER.warn(`Failed to ${description} after ${attempts} attempts. Error was ${error}`)
+                LOGGER.warn(`Failed to ${description} after ${initialAttempts} attempts. Error was ${error}`)
```
nchamo: Change #1

content/src/service/validations/Validations.ts
```diff
@@ -158,17 +158,20 @@ export class ValidatorInstance {
...
-                } else if (currentAuditInfo.version == auditInfoBeingDeployed.version && auditInfoBeingDeployed.originalMetadata) {
-                    if (!currentAuditInfo.originalMetadata) {
-                        this.errors.push(`Found an overlapping entity with a higher version already deployed.`)
-                    } else if (currentAuditInfo.originalMetadata.originalVersion > auditInfoBeingDeployed.originalMetadata.originalVersion) {
+            const currentAuditInfos: Map<Entity, AuditInfo | undefined> = new Map(await Promise.all(currentPointedEntities.map<Promise<[Entity, AuditInfo | undefined]>>(async entity => [entity, await auditInfoFetcher(entity.type, entity.id)])))
+            Array.from(currentAuditInfos.entries())
+                .filter(([, currentAuditInfo]) => !!currentAuditInfo)
+                .map<[Entity, AuditInfo]>(([currentEntity, currentAuditInfo]) => [currentEntity, currentAuditInfo!!])
+                .forEach(([currentEntity, currentAuditInfo]) => {
+                if (happenedBeforeEntities(currentEntity, entityToBeDeployed)) {
```
nchamo: Change #3

PR #106, opened 2020-04-03
chore: Comms performance test & more stats in peer

pablitar:


comms/peer/src/Peer.ts
```diff
@@ -143,6 +143,10 @@ export class Peer implements IPeer {
 
       this.pingTimeoutId = schedulePing();
     }
+    
+    this.stats = new GlobalStats(this.config.statsUpdateInterval ?? 1000);
```
nchamo, line 147: Super nit: maybe you could make this a constant somewhere, so it's easier to find/understand

comms/performance-test/client/index.ts
```diff
@@ -0,0 +1,361 @@
...
+import { randomBetween } from "decentraland-katalyst-utils/util";
+import { PeerMessageTypes } from "../../peer/src/messageTypes";
+import { Position3D, Quaternion } from "decentraland-katalyst-utils/Positions";
+import { PeerConfig } from "../../peer/src";
+import { PositionData, CommsMessage, ProfileData, ChatData } from "./protobuf/comms";
+import { util } from "../../peer/src/peerjs-server-connector/util";
+import { GlobalStats } from "../../peer/src/stats";
+import { Reader } from "protobufjs";
+
+function uuid(): string {
```
nchamo, line 11: We have a uuid library dependency already. Check the class `NameKeeper` on the content module

comms/performance-test/client/index.ts
```diff
@@ -0,0 +1,361 @@
...
+}
+
+(async () => {
+  if (testId) await testStarted();
+
+  console.log("Creating " + numberOfPeers + " peers");
+
+  const peers: SimulatedPeer[] = await Promise.all([...new Array(numberOfPeers).keys()].map(_ => createPeer()));
+
+  // function doSendMessage(container: SimulatedPeer) {
```
nchamo, line 285: Why are we keeping this?

PR #123, opened 2020-05-25
feat: adding deployments endpoint

nchamo:
We are adding a `/deployments` endpoint that exposes all deployments. Right now, they can only be filtered by: `fromLocalTimestamp` & `toLocalTimestamp`. Results are paginated similarly to the `/history` endpoint.

In the near future, we will add more filters to the endpoint and try to replace the `/entities` and `/audit` endpoints

### DIFF
--- a/content/src/denylist/DenylistServiceDecorator.ts
+++ b/content/src/denylist/DenylistServiceDecorator.ts
@@ -240,12 +240,12 @@ export class DenylistServiceDecorator implements MetaverseContentService {
     const entityToTarget: Map<Entity, DenylistTarget> = new Map(entities.map(entity => [entity, buildEntityTarget(entity.type, entity.id)]));
 
     // Check if targets are denylisted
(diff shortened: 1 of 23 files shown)


content/src/denylist/DenylistServiceDecorator.ts
```diff
@@ -140,12 +141,63 @@ export class DenylistServiceDecorator implements MetaverseContentService {
...
+        const entityTarget = buildEntityTarget(entityType, entityId);
+        const hashTargets: Map<ContentFileHash, DenylistTarget> = !content ? new Map() : new Map(Array.from(content.entries())
+          .map(([, hash]) => [hash, buildContentTarget(hash)]))
+        entityTargetsByEntity.set(entityId, entityTarget);
+        contentTargetsByEntity.set(entityId, hashTargets);
+        allTargets.push(entityTarget, ...hashTargets.values())
+      })
+
+      // Check which targets are denylisted
+      const queryResult = await this.denylist.areTargetsDenylisted(task.denylist, allTargets)
```
marcosnc, line 169: nit: queryResult is not self explanatory.

PR #126, opened 2020-05-29
test: verify that deltas are stored correctly

nchamo:


nchamo:
The `getDeltas` is only exposed for testing right now. When we start using it, we will add the appropriate filters

PR #129, opened 2020-06-04
feat: add filters to deployments endpoint

nchamo:
We are adding a few filters to the `/deployments` endpoint. Previously, we only had:
* fromLocalTimestamp
* toLocalTimestamp

Now, we are adding:
* deployedBy
* entityTypes
* entityIds
* pointers
* onlyCurrentlyPointed

Also, now, by default we won't show the `auditInfo` unless a query param called `showAudit` is set to true. In the near future, we will replace this query param with a `fields` parameter, where you can specify the fields you would like to retrieve.

### DIFF
--- a/content/src/controller/Controller.ts
+++ b/content/src/controller/Controller.ts
@@ -231,16 +231,16 @@ export class Controller {
         // Path: /deployments
         // Query String: ?fromLocalTimestamp={timestamp}&toLocalTimestamp={timestamp}&entityType={entityType}&entityId={entityId}&onlyCurrentlyPointed={boolean}&deployedBy={ethAddress}
 
-        const entityTypes:(EntityType|undefined)[] = this.asArray<string>(req.query.entityType).map(type => this.parseEntityType(type))
-        const entityIds:EntityId[]                 = this.asArray<EntityId>(req.query.entityId)
-        const fromLocalTimestamp                   = this.asInt(req.query.fromLocalTimestamp)
-        const toLocalTimestamp                     = this.asInt(req.query.toLocalTimestamp)
-        const onlyCurrentlyPointed                 = this.asBoolean(req.query.onlyCurrentlyPointed)
-        const showAudit                            = this.asBoolean(req.query.showAudit) ?? false
-        const deployedBy                           = this.asArray<EthAddress>(req.query.deployedBy)
-        const pointers                             = this.asArray<Pointer>(req.query.pointer)
-        const offset                               = this.asInt(req.query.offset)
-        const limit                                = this.asInt(req.query.limit)
+        const entityTypes:(EntityType|undefined)[] | undefined = this.asArray<string>(req.query.entityType).map(type => this.parseEntityType(type))
+        const entityIds:EntityId[] | undefined       = this.asArray<EntityId>(req.query.entityId)
+        const fromLocalTimestamp: number | undefined = this.asInt(req.query.fromLocalTimestamp)
+        const toLocalTimestamp: number | undefined   = this.asInt(req.query.toLocalTimestamp)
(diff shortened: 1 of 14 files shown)


content/src/controller/Controller.ts
```diff
@@ -229,21 +229,41 @@ export class Controller {
...
         // Method: GET
         // Path: /deployments
-        // Query String: ?fromLocalTimestamp={timestamp}&toLocalTimestamp={timestamp}
-        const fromLocalTimestamp = this.asInt(req.query.fromLocalTimestamp)
-        const toLocalTimestamp   = this.asInt(req.query.toLocalTimestamp)
-        const offset             = this.asInt(req.query.offset)
-        const limit              = this.asInt(req.query.limit)
+        // Query String: ?fromLocalTimestamp={timestamp}&toLocalTimestamp={timestamp}&entityType={entityType}&entityId={entityId}&onlyCurrentlyPointed={boolean}&deployedBy={ethAddress}
+
+        const entityTypes:(EntityType|undefined)[] = this.asArray<string>(req.query.entityType).map(type => this.parseEntityType(type))
```
marcosnc, line 234: Shouldn't be `EntityType[]|undefined` ?

content/src/controller/Controller.ts
```diff
@@ -229,21 +229,41 @@ export class Controller {
...
+        const entityIds:EntityId[]                 = this.asArray<EntityId>(req.query.entityId)
```
marcosnc, line 235: Shouldn't we support `undefined` here?

content/src/controller/Controller.ts
```diff
@@ -229,21 +229,41 @@ export class Controller {
...
+        const fromLocalTimestamp                   = this.asInt(req.query.fromLocalTimestamp)
+        const toLocalTimestamp                     = this.asInt(req.query.toLocalTimestamp)
+        const onlyCurrentlyPointed                 = this.asBoolean(req.query.onlyCurrentlyPointed)
+        const showAudit                            = this.asBoolean(req.query.showAudit) ?? false
+        const deployedBy                           = this.asArray<EthAddress>(req.query.deployedBy)
```
marcosnc, line 240: Nit: add the type, please.

content/src/storage/repositories/DeploymentsRepository.ts
```diff
@@ -74,19 +73,53 @@ export class DeploymentsRepository {
...
+        }
+
+        if (filters?.originServerUrl) {
+            values.originServerUrl = filters.originServerUrl
+            whereClause.push(`dep1.origin_server_url = $(originServerUrl)`)
+        }
+
+        if (filters?.deployedBy && filters.deployedBy.length > 0) {
+            values.deployedBy = filters.deployedBy
+            whereClause.push(`dep1.deployer_address IN ($(deployedBy:list))`)
```
marcosnc, line 103: Just to be sure we are safe against SQL injection: we're just setting a query parameter here, right?

content/src/service/deployments/DeploymentManager.ts
```diff
@@ -159,9 +159,17 @@ export type PartialDeploymentHistory = {
...
 
-export type DeploymentFilters = {
-    fromLocalTimestamp?: Timestamp
-    toLocalTimestamp?: Timestamp
+export type  DeploymentFilters = {
+    fromLocalTimestamp?: Timestamp,
+    toLocalTimestamp?: Timestamp,
+    fromOriginTimestamp?: Timestamp,
+    toOriginTimestamp?: Timestamp,
+    originServerUrl?: ServerAddress,
```
marcosnc, line 167: Maybe this should be an array too.

nchamo: This is currently not exposed. It is there so the `/history` endpoint can use it. If we decide to expose it, then we should discuss how to do so

PR #138, opened 2020-06-17
feat: Lambdas allow resize of images contents

pablitar:


lambdas/src/apis/images/controllers/images.ts
```diff
@@ -0,0 +1,119 @@
...
+    return true;
+  } catch (error) {
+    return false;
+  }
+}
+
+async function ensureDirectoryExists(directory: string): Promise<void> {
+  const alreadyExist = await existPath(directory);
+  if (!alreadyExist) {
+    await fs.promises.mkdir(directory);
```
nchamo, line 40: On the content server we had a few problems because we would get many requests and we would end up calling `mkdir` many times. I would add a `try/catch` to avoid failing on those situations

PR #144, opened 2020-06-22
refactor: use deployments internally

nchamo:
More cleanup!

We are now using the `getDeployments` method internally, and deleting:
* `getAuditInfo`
* `getEntitiesByIds`
* `getEntitiesByPointers`

By doing this we are getting rid of all code, and moving towards getting rid of entities, and only using them during deploys. For everything else, we will be using `deployments`.


content/src/controller/ControllerEntityFactory.ts
```diff
@@ -21,4 +22,9 @@ export class ControllerEntityFactory {
         return { id, type, timestamp, pointers, content, metadata }
     }
 
+    static maskDeployment(fullDeployment: Deployment, fields?: EntityField[]): ControllerEntity {
```
marcosnc, line 25: nit: I would prefer `deployment2ControllerEntity` as the name.

content/src/service/time/TimeSorting.ts
```diff
@@ -14,15 +15,29 @@ export function sortFromOldestToNewest<T extends EntityComparable> (comparableAr
...
-/** Return true if the first entity was created happened before the second one */
-export function happenedBeforeEntities(entity1: Entity, entity2: Entity): boolean {
-    const comparable1 = { entityId: entity1.id, timestamp: entity1.timestamp }
-    const comparable2 = { entityId: entity2.id, timestamp: entity2.timestamp }
-    return happenedBefore(comparable1, comparable2)
+/** Return true if the first deployments happened before the second one */
+export function happenedBefore(toBeComparable1: Deployment | Entity | EntityComparable, toBeComparable2: Deployment | Entity | EntityComparable): boolean {
+    let comparable1: EntityComparable
+    let comparable2: EntityComparable
+    if ('auditInfo' in toBeComparable1) {
```
marcosnc, line 26: nit: extract this behavior to a private function.

PR #145, opened 2020-06-23
feat: add fields query parameter

nchamo:


content/src/controller/Controller.ts
```diff
@@ -246,17 +246,37 @@ export class Controller {
...
             res.status(400).send({ error: `Found an unrecognized entity type` });
             return
         }
 
+        // TODO: Delete after one deployment
+        if (showAudit) {
+            if (fields && !fields.includes(DeploymentField.AUDIT_INFO)) {
+                fields += `,${DeploymentField.AUDIT_INFO}`
+            } else if (!fields) {
+                fields = `${DeploymentField.POINTERS},${DeploymentField.CONTENT},${DeploymentField.METADATA},${DeploymentField.AUDIT_INFO}`
```
marcosnc, line 262: I think this should be outside the `showAudit` clause.

PR #148, opened 2020-06-24
fix: add timeout when querying thegraph

nchamo:
We are now calling the `Fetcher` on `dcl-catalyst-commons` when querying `thegraph`. This fetcher already uses a timeout for requests

### DIFF
--- a/content/src/service/access/AccessCheckerImpl.ts
+++ b/content/src/service/access/AccessCheckerImpl.ts
@@ -1,6 +1,6 @@
 import log4js from "log4js"
 import ms from "ms";
-import { EntityType, Pointer, Timestamp } from "dcl-catalyst-commons";
+import { EntityType, Pointer, Timestamp, Fetcher } from "dcl-catalyst-commons";
 import { AccessChecker } from "./AccessChecker";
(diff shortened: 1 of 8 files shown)


content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -313,26 +314,13 @@ export class AccessCheckerImpl implements AccessChecker {
...
-        const res = await fetch(this.dclParcelAccessUrl, opts)
-        if (res.ok) {
-            const json = await res.json()
-            if (json.errors) {
-                throw new Error(
-                    `Error querying graph. Reasons: ${JSON.stringify(json.errors)}`
-                )
-            }
-            return json.data
+        const json = await this.fetcher.postForm(this.dclParcelAccessUrl, JSON.stringify({ query, variables }), { 'Content-Type': 'application/json' }, { attempts: 1 })
```
marcosnc, line 317: Should we perform more than 1 attempt?

nchamo: If this fails, there is an external retries that has 5 attempts

PR #151, opened 2020-06-26
fix: avoid out of memory

nchamo:
Previously, we used to load all deployments in-memory, to then process them with streams. This worked in the past, because we were using little information to sync with other servers. Now, after #139, we can avoid the extra requests, but we have more information. This is the reason why we started getting out-of-memory errors, and we needed to fix it.

Now, thanks to https://github.com/decentraland/catalyst-client/pull/9, we can get the deployments as streams, so we can avoid having all the information in-memory. The process now works like this:
1. We get a stream of deployments by using the `catalyst-client`.
2. We add a stream operation that lets the deployments through, but stores the newest localTimestamp
3. We then merge (using interleaving) all streams from all content servers into one
4. We have a stream operation that removes unnecessary information that we might get from content servers that don't have #145 yet. This means that we can remove this in the future
5. We then have another stream operation that removes already processed deployments, to avoid duplicates
6. We then have another stream operation that accumulates a number of deployments (currently set at 300) and then queries the database to verify if any of them have already been deployed. If they have, then they are filtered out.
7. We then have a stream operation that downloads all necessary files to perform the deployments. This operation handled 15 parallel **ordered** downloads, and was already in place. However, we managed to remove the order. Previously, if the deployment N wasn't ready, then the deployment N + 16 wouldn't start. Now, it can.
8. Finally, we will deploy the new entity. If the deployment isn't successful, then it is added to the list of `failedDeployments`

After this whole process if done, if there were no errors, we will use the timestamp calculated in step 2 for the next time we ask for deployments.


content/src/service/synchronization/EventDeployer.ts
```diff
@@ -18,45 +18,18 @@ export class EventDeployer {
...
 
+    async processAllDeployments(deployments: Readable[], options?: HistoryDeploymentOptions) {
         // Process history and deploy it
-        return this.eventProcessor.processDeployments(newDeployments, options)
+        return this.eventProcessor.processDeployments(deployments, options)
     }
 
     /** Download and prepare everything necessary to deploy an entity */
-    private async prepareDeployment(deployment: Deployment, source?: ContentServerClient): Promise<DeploymentExecution> {
+    private async prepareDeployment(deployment: DeploymentWithAuditInfo, source?: ContentServerClient): Promise<DeploymentExecution> {
```
nchamo, line 32: Changed this because we are not getting the full deployment now, we are only getting the base properties and the audit info

content/src/service/synchronization/streaming/EventStreamProcessor.ts
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * This method takes many deployment streams and tries to deploy them locally.
+     */
+    async processDeployments(deployments: Readable[], options?: HistoryDeploymentOptions) {
+        // Merge the streams from the different servers
+        const merged = mergeStreams(deployments);
+
+        // TODO: Remove on next deployment
+        // Delete information that is not yet excluded from the response
+        const reduceExtraInfo = streamMap(({ entityType, entityId, entityTimestamp, deployedBy, auditInfo }) => {
```
marcosnc, line 32: I think It would be better to define these transformations as constants outside the function, so it will be clearer they don't require local information to be created.

nchamo: This transformation you are point to is the only one that doesn't have state and wouldn't need to be instantiated on every sync. But it's also going away on the next deployment

content/src/service/synchronization/clients/ContentServerClient.ts
```diff
@@ -18,35 +20,44 @@ export class ContentServerClient {
...
-            }))
-        } catch (error) {
-            if (this.connectionState === ConnectionState.CONNECTED) {
-                this.connectionState = ConnectionState.CONNECTION_LOST
-                ContentServerClient.LOGGER.info(`Lost connection to '${this.address}'`)
+        // Wait for stream to end to update connection state
+        stream.once('end', () => {
+            if (!error) {
+                // Update connection state
+                if (this.connectionState !== ConnectionState.CONNECTED) {
```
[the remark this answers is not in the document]
nchamo: If I wasn't connected, then I log that I **could** connect

PR #153, opened 2020-06-29
feat: exposing deltas endpoint

nchamo:
We are now exposing the changes each deployment generates on pointers. Each change (or delta) will be exposed in the following way:
```
{
    "entityType": "scene",
    "entityId": "QmWV9Z42kqMF73F1qWK2a6oxdV69hRjMShbWmxdvbhfJGp",
    "localTimestamp": 1593446235668,
    "changes": [
        {
            "pointer": "52,-52",
            "before": "QmWdRSUiYSvXVHUMfKgPNLu61KqeDF89RBhbP4TdBD4SrF",
            "after": "QmWV9Z42kqMF73F1qWK2a6oxdV69hRjMShbWmxdvbhfJGp"
        },
        {
            "pointer": "52,-53",
            "after": "QmWV9Z42kqMF73F1qWK2a6oxdV69hRjMShbWmxdvbhfJGp"
        }
    ]
}
```


content/src/Server.ts
```diff
@@ -72,6 +72,7 @@ export class Server {
     this.registerRoute("/denylist/:type/:id", controller, controller.isTargetDenylisted, HttpMethod.HEAD);
     this.registerRoute("/failedDeployments", controller, controller.getFailedDeployments);
     this.registerRoute("/challenge", controller, controller.getChallenge);
+    this.registerRoute("/deltas/:type", controller, controller.getDeltas);
```
marcosnc, line 75: Should we avoid the type as a path parameter? In the past, this kind of parameterization shown to be very limiting, and I think we may want to receive deltas from different kinds of types.

PR #154, opened 2020-07-01
feat: adding content server snapshots

nchamo:
We are now adding snapshots. These can be defined as a list of all the active entities on the content server, at a given time. By active, we mean entities that are currently referenced by at least one pointer.

Every N deployments (N depends on the entity type) we will create a snapshot, and add it to the storage. Snapshots will be a text file with a JSON array with the following type: `[EntityId, Pointer[]][]`. The good part about this, is that you could easy read this file into an ES6 Map like this:
```
const fileContent: Buffer = fs.readFile(...)
const array: [EntityId, Pointer[]][] = JSON.parse(fileContent.toString())
const asMap: Map<EntityId, Pointer[]> = new Map(array)
```

Then, we will expose a `/snapshot` endpoint that displays the hash of the latest snapshot, and the local deployment timestamp of the last deployment that was included in the snapshot

### DIFF
--- a/content/src/controller/Controller.ts
+++ b/content/src/controller/Controller.ts
@@ -346,7 +346,7 @@ export class Controller {
         const metadata = this.snapshotManager.getSnapshotMetadata(type)
 
         if (!metadata) {
-            res.send({ error: 'Snapshot not yet created' })
+            res.status(503).send({ error: 'Snapshot not yet created' })
         } else {
             res.send(metadata)
         }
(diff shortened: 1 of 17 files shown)


content/src/controller/Controller.ts
```diff
@@ -329,6 +331,27 @@ export class Controller {
...
+
+        const type = this.parseEntityType(req.params.type)
+
+        // Validate type is valid
+        if (!type) {
+            res.status(400).send({ error: `Unrecognized type: ${req.params.type}` });
+            return
+        }
+
+        const metadata = this.snapshotManager.getSnapshotMetadata(type)
```
marcosnc, line 346: Nit: why "metadata"?

content/src/controller/Controller.ts
```diff
@@ -329,6 +331,27 @@ export class Controller {
...
+
+        if (!metadata) {
+            res.send({ error: 'Snapshot not yet created' })
```
marcosnc, line 349: Should we return something different than a 200-OK here?

PR #155, opened 2020-07-01
feat: Suspend Relay Protocol support

pablitar:
This change enables the peer library to use a control message to request relay suspension from a particular peer.

The idea is to reduce the unnecessary gossip and the amount of expired/duplicates packages that a peer may receive.

The mechanisms added are:
- Peers collect data on which connected peers they should suspend relay for each relayed peer. Basically, every time an expired or duplicate message is received, a relayed peer is marked to be suspended for that connection, trying to ensure that at least 1 connection for which the relayed peer is reachable always remains
- Peers send to other peers relay suspension messages on a regular basis, configurable using the relaySuspensionInterval config variable
- Relays suspension expire after some time, configured using the relaySuspensionDuration config variable. This is to account for the dynamic nature of the network.


comms/peer/src/Peer.ts
```diff
@@ -1120,7 +1278,7 @@ export class Peer implements IPeer {
             break;
           }
 
-          if (this.httpClient.lighthouseUrl !== payload.lighthouseUrl || this.currentLayer !== payload.layer) {
+          if (this.currentLayer !== payload.layer) {
```
moliva, line 1281: Why should we omit the lighthouse URL check?

pablitar: Ufff. Nice catch. I actually had changed this for a test in my local network :facepalm:

PR #161, opened 2020-07-06
fix: prevent concurrent deployments on the same pointers

nchamo:
We are now preventing concurrent deployments on the same pointers. These are EXTREMELY rare, but in the event they happen, the database could end up inconsistent.


content/src/service/ServiceImpl.ts
```diff
@@ -178,6 +190,9 @@ export class ServiceImpl implements MetaverseContentService, ClusterDeploymentsS
             await Promise.all(this.listeners.map(listener => listener({ entity, auditInfo: auditInfoComplete, origin })))
         }
 
+        // Update the current list of pointers being deployed
+        entity.pointers.forEach(pointer => this.pointersBeingDeployed.get(entity.type)!!.delete(pointer))
```
marcosnc, line 194: I think it's better to get the set before the forEach.

nchamo:
@marcosnc, how would you test this?

I've tried making concurrent deployments, but they were not "concurrent enough" for the new mechanism to kick in

PR #165, opened 2020-07-23
Added contract based lambda function to retrieve the catalyst servers list.

marcosnc:


lambdas/src/apis/contracts/TimeRefreshedDataHolder.ts
```diff
@@ -0,0 +1,19 @@
+export class TimeRefreshedDataHolder<T> {
+
+    private value: T
+
+    constructor(private readonly provider: ()=>Promise<T>, private readonly refreshTime: number) { }
```
nchamo, line 5: Super nit. Space before and after `=>`

lambdas/src/apis/contracts/DAOListsClient.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export class DAOListContractClient {
+
+  constructor(private readonly contract: DAOListContract) {  }
+
+  async getAllValues(): Promise<Set<string>> {
+    const count = await this.contract.getCount()
+    const values: Set<string> = new Set()
+    for (let i = 1; i <= count; i++) {
```
nchamo, line 14: Why are we starting with 1? Is this ok?

lambdas/src/apis/contracts/DAOCache.ts
```diff
@@ -0,0 +1,31 @@
+import { ServerMetadata } from "decentraland-katalyst-commons/ServerMetadata";
+import { TimeRefreshedDataHolder } from "./TimeRefreshedDataHolder";
+import { DAOContractClient } from "decentraland-katalyst-commons/DAOClient";
+import ms from "ms";
+import { DAOListClient } from "./DAOListsClient";
+
+const REFRESH_TIME: number = ms('15m')
```
nchamo, line 7: I think we could event set this higher, like `30m`. These values will rarely change

lambdas/src/apis/contracts/DAOListsClient.ts
marcosnc: Yes... :( 
The 0 position is a reserved value according to Nacho.

lambdas/src/apis/contracts/DAOCache.ts
marcosnc: You're right, I'll set it to 30m.

lambdas/src/apis/contracts/DAOListsClient.ts
marcosnc: I changed the code to use the `get`  method in the contract, it starts at 0.

PR #169, opened 2020-09-08
Filter only the names that are used in the profiles.

marcosnc:


lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -28,12 +28,11 @@ export async function getProfileById(env: Environment, fetcher: SmartContentServ
...
  */
 async function filterNonOwnedNames(theGraphBaseUrl: string, profileId: string, metadata: EntityMetadata): Promise<EntityMetadata> {
-    const isThereAnAvatarWithAName: boolean = metadata.avatars.map(profile => profile.name)
+    const avatarsNames: string[] = metadata.avatars.map(profile => profile.name)
         .filter(name => name && name !== '')
-        .length > 0
 
-    if (isThereAnAvatarWithAName) {
-        const ownedENS = await getOwnedENS(theGraphBaseUrl, profileId)
+    if (avatarsNames.length>0) {
```
nchamo, line 34:
```suggestion
    if (avatarsNames.length > 0) {
```

lambdas/src/apis/profiles/ensFiltering.ts
```diff
@@ -1,32 +1,32 @@
...
+        names: names } })
 })
 
-export async function getOwnedENS(theGraphBaseUrl: string, ethAddress: string): Promise<string[]> {
+export async function filterENS(theGraphBaseUrl: string, ethAddress: string, namesToFilter: string[]): Promise<string[]> {
     const totalAttempts = 5
     for(let attempt=0; attempt<totalAttempts; attempt++) {
         try {
-            const response = await fetch(theGraphBaseUrl, opts(ethAddress))
+            const response = await fetch(theGraphBaseUrl, opts(ethAddress, namesToFilter))
```
nchamo, line 26: Maybe it would be a good time to replace this with our fetcher? Our fetcher supports both retries & timeouts

You can take a look at https://github.com/decentraland/catalyst/blob/master/content/src/service/access/AccessCheckerImpl.ts#L317

nchamo (approved):
Great work!!!

Added just two minor comments

PR #170, opened 2020-09-08
Added the new Wearable entity type.

marcosnc:


content/src/Environment.ts
```diff
@@ -201,7 +201,7 @@ export class EnvironmentBuilder {
         this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.PSQL_PORT                      , () => process.env.POSTGRES_PORT ?? DEFAULT_DATABASE_CONFIG.port)
         this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.GARBAGE_COLLECTION             , () => process.env.GARBAGE_COLLECTION === 'true')
         this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.GARBAGE_COLLECTION_INTERVAL    , () => ms('6h'))
-        this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.SNAPSHOT_FREQUENCY             , () => new Map([[EntityType.SCENE, 100], [EntityType.PROFILE, 500]]))
+        this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.SNAPSHOT_FREQUENCY             , () => new Map([[EntityType.SCENE, 100], [EntityType.PROFILE, 500], [EntityType.WEARABLE, 100]]))
```
nchamo, line 204: Maybe 50 would be better? This will happen so rarely, that we might never make new snapshots

content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -0,0 +1,99 @@
...
+import { Pointer, Fetcher } from "dcl-catalyst-commons";
+import { EthAddress } from "dcl-crypto";
+import { ContentAuthenticator } from "../auth/Authenticator";
+
+export class AccessCheckerForWearables {
+
+    constructor(
+        private readonly authenticator: ContentAuthenticator,
+        private readonly fetcher: Fetcher,
+        private readonly dclParcelAccessUrl: string,
```
nchamo, line 11: Maybe we should rename this?

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,356 +1,37 @@
...
 
-    constructor(
-        private readonly authenticator: ContentAuthenticator,
-        private readonly fetcher: Fetcher,
-        private readonly dclParcelAccessUrl: string) { }
+    private readonly accessCheckerForScenes: AccessCheckerForScenes;
+    private readonly accessCheckerForProfiles: AccessCheckerForProfiles;
+    private readonly accessCheckerForWearables: AccessCheckerForWearables;
+
+    constructor(authenticator: ContentAuthenticator, fetcher: Fetcher, dclParcelAccessUrl: string) {
```
nchamo, line 18: Does the name `dclParcelAccessUrl` still make sense? Since we are still using it for wearables?

nchamo (commented):
Looking good!

Could we add an integration test that checks a valid and an invalid deployment? So we avoid the same problem we had with scenes

marcosnc:
> Looking good!
> 
> Could we add an integration test that checks a valid and an invalid deployment? So we avoid the same problem we had with scenes

Yes, we can. I'm adding a reminder here to locate the test file: `test/integration/service/access/AccessCheckerImpl.spec.ts`.
We can also divide that test file into different test files, one for each type.

content/src/service/access/AccessCheckerImpl.ts
marcosnc: In fact, it's a really good catch! It's a totally different URL!

content/src/service/access/AccessCheckerImpl.ts
```diff
@@ -1,356 +1,37 @@
...
 
-    constructor(
-        private readonly authenticator: ContentAuthenticator,
-        private readonly fetcher: Fetcher,
-        private readonly dclParcelAccessUrl: string) { }
+    private readonly accessCheckerForScenes: AccessCheckerForScenes;
+    private readonly accessCheckerForProfiles: AccessCheckerForProfiles;
+    private readonly accessCheckerForWearables: AccessCheckerForWearables;
+
+    constructor(authenticator: ContentAuthenticator, fetcher: Fetcher, dclParcelAccessUrl: string, dclCollectionsAccessUrl: string) {
```
nchamo, line 18: Same here with the name change (if we changed it on the ENV)

content/src/Environment.ts
```diff
@@ -189,7 +192,8 @@ export class EnvironmentBuilder {
...
         this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.BOOTSTRAP_FROM_SCRATCH         , () => process.env.BOOTSTRAP_FROM_SCRATCH === 'true');
         this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.REQUEST_TTL_BACKWARDS          , () => ms('20m'));
-        this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.DCL_PARCEL_ACCESS_URL          , () => process.env.DCL_PARCEL_ACCESS_URL ?? (env.getConfig(EnvironmentConfig.ETH_NETWORK) === 'mainnet' ? DEFAULT_DCL_PARCEL_ACCESS_URL_MAINNET : DEFAULT_DCL_PARCEL_ACCESS_URL_ROPSTEN))
+        this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.DCL_PARCEL_ACCESS_URL          , () => process.env.DCL_PARCEL_ACCESS_URL      ?? (env.getConfig(EnvironmentConfig.ETH_NETWORK) === 'mainnet' ? DEFAULT_DCL_PARCEL_ACCESS_URL_MAINNET      : DEFAULT_DCL_PARCEL_ACCESS_URL_ROPSTEN))
+        this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.DCL_COLLECTIONS_ACCESS_URL     , () => process.env.DCL_COLLECTIONS_ACCESS_URL ?? (env.getConfig(EnvironmentConfig.ETH_NETWORK) === 'mainnet' ? DEFAULT_DCL_COLLECTIONS_ACCESS_URL_MAINNET : DEFAULT_DCL_COLLECTIONS_ACCESS_URL_ROPSTEN))
```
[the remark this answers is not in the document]
marcosnc: Yes, it's going to be used for other types too (at least Emotes).

content/src/service/access/AccessCheckerImpl.ts
marcosnc: idem.

PR #173, opened 2020-09-14
Added a new lambda to return DCL collections (aka wearables for now) …

marcosnc:
…in the ERC 721 standard.


lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -0,0 +1,65 @@
...
+    "common"   : 100000,
+    "uncommon" : 10000,
+    "rare"     : 5000,
+    "epic"     : 1000,
+    "legendary": 100,
+    "mythic"   : 10,
+    "unique"   : 1
+}
+
+type WearableEntity = {
```
nchamo, line 46: Maybe you could use the entity from `dcl-catalyst-commons`? Something like type:
```
type WearableEntity = Entity
```

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -0,0 +1,65 @@
...
+    id: string
+    type: string
+    pointers: string[]
+    timestamp: number
+    content?: EntityContent[]
+    metadata?: WearableMetadata
+}
+
+type EntityContent = {
```
nchamo, line 55: This type is also available in `dcl-catalyst-commons`

lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -83,7 +76,7 @@ function addBaseUrlToSnapshots(baseUrl: string, metadata: EntityMetadata): Entit
     return { avatars }
 }
 
-interface V3ControllerEntity {
+type V3ControllerEntity = {
```
nchamo, line 79: This type and `V3ControllerEntityContent` are also available on `dcl-catalyst-commons`. There is no need to re-define them

lambdas/src/Server.ts
```diff
@@ -88,3 +100,56 @@ export class Server {
     }
   }
 }
+
+/*
+
+https://peers.decentraland.org/lambdas/wearables/erc721/<contract>/<option>/<emission>
```
nchamo, line 106: What is all this? Can we delete it? Some of it is in spanish

lambdas/src/SmartContentServerFetcher.ts
```diff
@@ -1,28 +1,28 @@
...
     private static LOGGER = log4js.getLogger('SmartContentServerFetcher');
 
     private contentServerUrl: string | undefined
 
-    constructor(private readonly externalContentServerUrl: string) { }
+    constructor(private readonly externalContentServerUrl: string) {
+        super()
+    }
 
     async getContentServerUrl(): Promise<string> {
```
nchamo, line 19: Can we make this private now?

lambdas/src/SmartContentServerFetcher.ts
```diff
@@ -1,28 +1,28 @@
 import log4js from "log4js"
-import fetch from "node-fetch"
```
nchamo, line 2: Can we remove the dependency to `node-fetch` from the project now? I mean the bazel file, not the package json

lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -1,35 +1,28 @@
...
     // Method: GET
     // Path: /:id
     const profileId:string = req.params.id
-    const v3Url = (await smartContentServerFetcher.getContentServerUrl()) + `/entities/profile?pointer=${profileId}`
-    const response = await fetch(v3Url)
     let returnProfile: EntityMetadata = { avatars:[] }
-    if (response.ok) {
-        const entities:V3ControllerEntity[] = await response.json()
+    try {
+        const entities:V3ControllerEntity[] = await fetcher.fetchJsonFromContentServer(`/entities/profile?pointer=${profileId}`)
```
nchamo, line 11: Can we start using the `/deployments` endpoint?
It would be something like `/deployments?entityType=profile&pointer=${profileId}&onlyCurrentlyPointed=true?fields=metadata`

In a next iteration, it would also be nice to start using the [client](https://github.com/decentraland/catalyst-client)

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -0,0 +1,65 @@
...
+    file: string,
+    hash: string,
+}
+
+type WearableMetadata = {
+    i18n: { code: string, text: string}[]
```
nchamo, line 61:
```suggestion
    i18n: { code: string, text: string }[]
```

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -0,0 +1,65 @@
...
+    content?: EntityContent[]
+    metadata?: WearableMetadata
+}
+
+type EntityContent = {
+    file: string,
+    hash: string,
+}
+
+type WearableMetadata = {
```
nchamo, line 60: Maybe we should get someone with more experience with wearables to validate this

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -0,0 +1,65 @@
+import { Request, Response } from 'express'
+import { SmartContentServerFetcher } from '../../../SmartContentServerFetcher'
+
+export async function getStandardErc721(fetcher: SmartContentServerFetcher, req: Request, res: Response) {
+    // Method: GET
+    // Path: /standard/erc721/:contract/:option/:emission
+    const { contract, option, emission } = req.params;
+
+    try {
+        const entities:WearableEntity[] = await fetcher.fetchJsonFromContentServer(`/entities/wearable?pointer=${contract}-${option}`)
```
nchamo, line 10: Can we start using the `/deployments` endpoint?
It would be something like `/deployments?entityType=wearable&pointer=${contract}-${option}&onlyCurrentlyPointed=true?fields=metadata`

In a next iteration, it would also be nice to start using the [client](https://github.com/decentraland/catalyst-client/)

nchamo (commented):
Great work!

Just a few comments

lambdas/src/apis/collections/controllers/collections.ts
marcosnc: In fact, I kept the type defined here because I wanted to enforce the type of metadata field.

lambdas/src/apis/profiles/controllers/profiles.ts
marcosnc: Same here, this entity is enforcing the metadata type.

lambdas/src/Server.ts
marcosnc: Ups...

lambdas/src/SmartContentServerFetcher.ts
marcosnc: No, because sometimes you want to know this externally (at leas to log the URL).

lambdas/src/SmartContentServerFetcher.ts
marcosnc: Not yet, I preferred to not change the content-v2 translator (that uses node-fetch).

lambdas/src/apis/profiles/controllers/profiles.ts
marcosnc: I love the /deployments endpoint for obvious reasons however, I think this endpoint is a lot clearer because is entity oriented.
Agree with you regarding the client, but I would prefer to make that change in another iteration.

lambdas/src/apis/collections/controllers/collections.ts
marcosnc: Good catch!

lambdas/src/apis/collections/controllers/collections.ts
marcosnc: This is the current structure for wearables as they are in: https://wearable-api.decentraland.org/v2/addresses/0x87956abC4078a0Cc3b89b419928b857B8AF826Ed/wearables

lambdas/src/apis/collections/controllers/collections.ts
marcosnc: Same here.

nchamo (approved):
Great work!!!

PR #179, opened 2020-09-22
chore: NPM link support for peer library & added packet to packet callback

pablitar:
To be able to develop Peer library with Explorer more easily, added a script to create an NPM link.

Also added the low-level packet to packet callback. This can be useful if there is some information that the user of the peer library needs to use in the callback (for instance, the sequence id).


nchamo (approved):
Great work!

PR #181, opened 2020-10-05
Fixed eth address comparison for wearables.

marcosnc:


content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -48,10 +48,11 @@ export class AccessCheckerForWearables {
 
     private async checkCollectionAccess(collection: string, itemId: number, ethAddress: EthAddress): Promise<boolean> {
         try {
-            const collectionItems: WerableCollectionItems = await this.getCollectionItems(collection, itemId, ethAddress)
-            return collectionItems.collections[0]?.creator === ethAddress
-                || collectionItems.collections[0]?.managers.includes(ethAddress)
-                || collectionItems.items[0]?.managers.includes(ethAddress)
+            const ethAddressLowercase = ethAddress.toLocaleLowerCase()
+            const collectionItems: WerableCollectionItems = await this.getCollectionItems(collection, itemId, ethAddressLowercase)
+            return collectionItems.collections[0]?.creator === ethAddressLowercase
```
eordano, line 53: Why are we checking only the `[0]`th element?

eordano: What I wonder actually is what's the `getCollectionItems` contract interface, why does it need an address, and if it returns an array shouldn't we check all the items?

content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -48,46 +48,57 @@ export class AccessCheckerForWearables {
...
         }`
 
         try {
-            return await this.fetcher.queryGraph<WerableCollectionItems>(
+            const wearableCollectionsAndItems = await this.fetcher.queryGraph<WearableCollectionsAndItems>(
                 this.dclCollectionsAccessUrl,
                 query,
                 { collection: collection, itemId: itemId})
+            return {
+                collectionCreator : wearableCollectionsAndItems.collections[0]?.creator,
```
nchamo, line 82: Is it possible that `wearableCollectionsAndItems.collections` or `wearableCollectionsAndItems.items` is empty? If so, then we are not handing the `undefined` in `checkCollectionAccess`, and it would fail

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -30,10 +31,67 @@ export async function getStandardErc721(fetcher: SmartContentServerFetcher, req:
...
+
+async function internalContents(fetcher: SmartContentServerFetcher, res: Response, contract: string, option: string, selector: (WearableMetadata: WearableMetadata) => string |undefined) {
+    try {
+        let contentBuffer: Buffer | undefined = undefined
+        const entities:Entity[] = await fetcher.fetchJsonFromContentServer(`/entities/wearable?pointer=${contract}-${option}`)
+        if (entities && entities.length > 0 && entities[0].metadata) {
+            const wearableMetadata: WearableMetadata = entities[0].metadata
+            const relativeContentUrl = createRelativeContentUrl(fetcher, entities[0], selector(wearableMetadata))
+            if (relativeContentUrl) {
+                contentBuffer = await fetcher.fetchBufferFromContentServer(relativeContentUrl) // TODO: fetch a stream instead of a Buffer
```
nchamo, line 63: Could you create a github issue for this TODO?

nchamo (approved):
Great work!

content/src/service/access/AccessCheckerForWearables.ts
marcosnc: It's covered by the try/catch, but I'm including an extra check just to avoid the error log.

lambdas/src/apis/collections/controllers/collections.ts
marcosnc: Sure!

PR #204, opened 2020-10-27
Added an ownedName property but keeping not owned names.

marcosnc:


lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -23,7 +23,7 @@ export async function getProfileById(fetcher: SmartContentServerFetcher, ensOwne
 /**
  * We filter ENS to avoid send an ENS that is no longer owned by the user
```
nchamo, line 24: Could you also update the comment?

nchamo (approved):
Great work!

Issue #203, opened 2020-10-27
[Refactor] Remove hardcoded listeners from the service


nchamo:
In https://github.com/decentraland/catalyst/pull/154, we added the concept `listeners` to the content server service. When a deployment is made, listeners are called so that they can execute some action.

We have some hardcoded listeners into the service, for Segment and SQS. It would be best to remove that from the service itself, and let those elements just subscribe to the service.

PR #215, opened 2020-11-18
Refactor/report deployment on history manager

agusaldasoro:
This PR fixes an issue in https://github.com/decentraland/catalyst/issues/203


content/src/Server.ts
```diff
@@ -134,6 +134,7 @@ export class Server {
     await this.snapshotManager.start();
     await this.synchronizationManager.start();
     await this.garbageCollectionManager.start();
+    await this.service.start();
```
nchamo, line 137: Let's move this before 
```this.httpServer = this.app.listen(this.port);```

That way, the deployments count is made before HTTP requests can be made, and therefore before new deployments can be made

File: content/src/service/ServiceImpl.ts [hunk 2]
```diff
@@ -172,8 +177,7 @@ export class ServiceImpl implements MetaverseContentService, ClusterDeploymentsS
                     await this.storeEntityContent(hashes, alreadyStoredContent)
 
                     // Since we are still reporting the history size, add one to it
-                    await this.historyManager.reportDeployment(transaction.deployments)
-
+                    this.historySize++
```
nchamo, line 180: Let's move this outside the transaction, since it could potentially fail for some reason and we would still be counting  the new deployment. We can add it inside the same `if` that calls the listeners

content/src/service/history/HistoryManager.ts
```diff
@@ -2,7 +2,5 @@ import { Timestamp, LegacyPartialDeploymentHistory, ServerName } from "dcl-catal
 import { DeploymentsRepository } from "@katalyst/content/storage/repositories/DeploymentsRepository"
 
 export interface HistoryManager {
-    reportDeployment(deploymentsRepository: DeploymentsRepository): Promise<void>;
```
nchamo, line 5: Love this!

nchamo (approved):
Great work!

PR #225, opened 2020-11-30
fix: Fixing issues when submitting weird lighthouse assigned ids in urls

pablitar:


comms/lighthouse/BUILD.bazel
```diff
@@ -110,7 +110,10 @@ ts_library(
         ":lighthouse",
         "//commons/servers",
         "//commons/utils",
+        "@npm//@types/express",
         "@npm//@types/jasmine",
+        "@npm//isomorphic-fetch",
```
nchamo, line 115: I would recommend using [cross-fetch](https://www.npmjs.com/package/cross-fetch) instead. On the `catalyst-client` we had some problems with `isomorphic-fetch`.

nchamo (approved):
Looks good!

Small recommendation about which fetch lib to use

comms/lighthouse/BUILD.bazel
pablitar: This is just for testing so it shouldn't matter that much

Issue #223, opened 2020-11-27
Make pointers case insensitive


agusaldasoro:
Currently, all pointers are being saved as lowercase. So, we need to modify the code so when setting a pointer through a query param, then the values retrieved are the ones that match case insensitive with the pointer param.

PR #226, opened 2020-11-30
feat: pointers case insensitive in queries

agusaldasoro:
This fixes https://github.com/decentraland/catalyst/issues/223


nchamo (approved):
Great work!

Issue #221, opened 2020-11-24
Add linter and formatter to project


agusaldasoro:
We need to agree on formatting and linter standards for the repository, then configure them and modify all the files to comply them. This issue was started in https://github.com/decentraland/catalyst/pull/220

###  For format we will be using [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)

1. Configure Prettier acording the standard that the team has decided.
2. Configure VSCode to automatically format a file on save.
3. Run prettier over all the files in the repository.
4. Add Prettier as a requirement when setting up the project as a developer.
5. [nice to have] Create a pre commit that validates format with prettier over the modified files. This step must have a way to skip it.

### Linter

1. Configure the linter in the repository.
2. Run the linter and resolve all the issues found.
3. Add running the linter as part of the pipeline.

PR #227, opened 2020-12-03
chore: use eslint

agusaldasoro:
This PR fixes https://github.com/decentraland/catalyst/issues/221

## Changes
- Removed `tslinter` in order to use `eslinter`, as the first one was deprecated.
- Configured `eslinter` to run with `prettier` as a dependency. This means that when running `eslinter`, `prettier` will be executed first.
- Configured the project to run `eslinter` as a pre-commit hook.
- Configured the project so if using VS Code IDE, then the linter will run on save.
- Configured the CI to run `yarn lint:check` as a dependency when checking out, to avoid errors by committing code with `--no-verify` flag.

### Note
Please check that I've disabled some validations for a few entire files inside the `comms` folder, as those errors were repeated.


.circleci/config.yml
```diff
@@ -61,6 +61,8 @@ jobs:
             # fallback to using the latest cache if no exact match is found
             - v1-dependencies-
 
+      - run: yarn lint:check
```
nchamo, line 64: This probably shouldn't be a part of the checkout. It should be its own job, and run between checkout and other `builds-*`

.eslintignore
```diff
@@ -0,0 +1,2 @@
+comms/peer/src/Peer.ts
```
nchamo, line 1: How can we fix this file and avoid the need to ignore it?

comms/lighthouse/src/handlers.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/ban-types */
```
nchamo, line 1: What does this do?

comms/lighthouse/src/server.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
nchamo, line 1: What does this do?

nchamo (commented):
Just a few questions

comms/lighthouse/src/handlers.ts
agusaldasoro: This comments avoids `eslint` to fail when a value is cast to object

comms/lighthouse/src/server.ts
agusaldasoro: This comments avoids `eslint` to fail when the annotation `@ts-ignore` is used

nchamo (approved):
Great work!

Issue #56, opened 2020-02-12
[monitoring] segment payload size larger than supported


abarmat:
The content server is exhibiting a warning related to tracking of events.

```
Your message must be < 32kb. This is currently surfaced as a warning to allow clients to update. Versions released after August 1, 2018 will throw an error instead. Please update your code before then. {
 userId: '0xdc13378dafca7fe2306368a16bcfac38c80bfcad',
  event: 'Catalyst Content Upload',
  properties:
   { server: '2b235332-2e10-4d07-9966-b95efb6146ec',
     type: 'scene',
     cid: 'QmUNpSvuqLVXfRgSsXXSKeRsixu8SnK7tmEimJo9ncQJTg',
     pointers:
```

agusaldasoro:
Hi @abarmat, I've just closed this issue as https://github.com/decentraland/catalyst/pull/229 was merged.

The payload from the report information sent to Segment was bigger than 32kb, so that was triggering this error.

The solution for this issue was reducing the payload sent to Segment by removing the pointers and file information from it, as those items were arrays they were responsible for increasing the size of the request when an entity hasd many pointers and/or many fields.

PR #229, opened 2020-12-08
Remove pointers and files from segment reporting

agusaldasoro:
This PR Fixes https://github.com/decentraland/catalyst/issues/56

### Description
As the payload sent to [Segment](https://segment.com/) was bigger than 32kb, then the error shown in issue #56 was being triggered. 

### Solution
Remove pointers and files from the reporting to Segment.


nchamo (approved):
Great work!

PR #230, opened 2020-12-08
fix: avoid flakiness in test

nchamo:
This test that was modified was flaky. It was flaky because we hoped that servers 1 and 3 didn't learn about the updates in server 2 before it was shut down. Now, we:
1. Start only server 2
1. Deploy E2 on it
1. Shut it down
1. Start servers 1 and 3. 

By doing this, there is no way that servers 1 and 3 learn about E2 until server 2 is restarted


agusaldasoro (approved):
Nice!

PR #232, opened 2020-12-11
fix: make pointers and deployedBy filters case insensitive

nchamo:
In https://github.com/decentraland/catalyst/pull/226 we attempted to make pointers case insensitive. However, we missed to do so, because there was a check on `DenylistServiceDecorator` that was still case sensitive. 

Now, we fixed it 😃


agusaldasoro (approved):
Great! 👍

Issue #228, opened 2020-12-04
Modify User-Agent between catalyst requests


nchamo:
Currently, when content servers sync with each other, the [user agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) is set to `node-fetch/1.0 (+https://github.com/bitinn/node-fetch)`

It would be better if we were able to modify the user agent header, so that it was clearer that the requests come from catalyst servers, and not some generic node-fetch library.

This will probably requiere changes to `catalyst-client` and `catalyst-commons` repositories. Ideally, this would be done in a way that if we want to add future headers to requests, we won't need to change those other 2 repositories

PR #235, opened 2020-12-14
feat: add user-agent header

agusaldasoro:
This PR fixes https://github.com/decentraland/catalyst/issues/228

### Changes
- Updated `catalyst-commons` and `catalyst-client` libraries to support headers on every request.
- Configured ContentServer to send its own `User-Agent` header


nchamo (approved):
Great work!

Issue #113, opened 2020-04-19
Message to block should include an indication of the action to execute


eordano:
This line:

https://github.com/decentraland/catalyst/blob/f700ca5ec100cd8fb8d6ef4d522b9aa886ea2973/content/src/denylist/Denylist.ts#L75

Could be abused if the signature for a different action would be taken.

The correct signed message should look like:

`block-{type}-{entity or pointer}-{timestamp}`

PR #236, opened 2020-12-14
Added action to block and unblock messages.

marcosnc:
Fixes https://github.com/decentraland/catalyst/issues/113


nchamo (commented):
Great work!

I just realized that we don't have a test that sends a wrong signature and verifies that the request failed. Could you add one?

Most tests are here https://github.com/decentraland/catalyst/blob/197fa44cf0eb75d8750ffd28bfcd20d86688396a/content/test/integration/denylist/Denylist.spec.ts#L22

marcosnc:
> Great work!
> 
> I just realized that we don't have a test that sends a wrong signature and verifies that the request failed. Could you add one?
> 
> Most tests are here
> 
> https://github.com/decentraland/catalyst/blob/197fa44cf0eb75d8750ffd28bfcd20d86688396a/content/test/integration/denylist/Denylist.spec.ts#L22

Sure!

nchamo (approved):
Great work!

agusaldasoro (approved):
Nice!

Issue #190, opened 2020-10-15
Return a BAD_REQUEST with validation errors


nchamo:
As of now, when an entity is deployed and fails, all errors return a [500](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) (Internal server error). 

Validation errors should return a [400](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) instead.

PR #241, opened 2020-12-17
feat: return 400 error code when not authorised endpoint /denylist/{type}/{id}

agusaldasoro:
This PR fixes https://github.com/decentraland/catalyst/issues/190

### DIFF
--- a/content/src/controller/Controller.ts
+++ b/content/src/controller/Controller.ts
@@ -18,7 +18,7 @@ import { AuthChain, Authenticator, AuthLink, EthAddress, Signature } from 'dcl-c
 import express from 'express'
 import fs from 'fs'
 import log4js from 'log4js'
-import { Denylist, OperationResult, OperationStatus } from '../denylist/Denylist'
+import { Denylist, isSuccessfulOperation, OperationResult } from '../denylist/Denylist'
 import { parseDenylistTypeAndId } from '../denylist/DenylistTarget'
 import { CURRENT_COMMIT_HASH, CURRENT_CONTENT_VERSION } from '../Environment'
 import { ContentAuthenticator } from '../service/auth/Authenticator'
@@ -541,7 +541,7 @@ export class Controller {
 
     try {
       const result: OperationResult = await this.denylist.addTarget(target, { timestamp, authChain })
-      if (result.status === OperationStatus.OK) {
+      if (isSuccessfulOperation(result)) {
         res.status(201).send()
       } else {
         res.status(400).send(result.message)
@@ -569,7 +569,7 @@ export class Controller {
 
     try {
       const result: OperationResult = await this.denylist.removeTarget(target, { timestamp, authChain })
-      if (result.status === OperationStatus.OK) {
+      if (isSuccessfulOperation(result)) {
         res.status(200).send()
       } else {
         res.status(400).send(result.message)
--- a/content/src/controller/Controller.ts
+++ b/content/src/controller/Controller.ts
@@ -18,7 +18,7 @@ import { AuthChain, Authenticator, AuthLink, EthAddress, Signature } from 'dcl-c
 import express from 'express'
 import fs from 'fs'
 import log4js from 'log4js'
-import { Denylist, isSuccessfulOperation, OperationResult } from '../denylist/Denylist'
+import { Denylist, DenylistSignatureValidationResult, isSuccessfulOperation } from '../denylist/Denylist'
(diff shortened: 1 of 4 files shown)


content/src/denylist/Denylist.ts
```diff
@@ -116,3 +129,17 @@ export enum DenylistAction {
   ADDITION = 'addition',
   REMOVAL = 'removal'
 }
+
+export enum OperationStatus {
```
nchamo, line 133: Could we give more context on the name? Maybe something like `DenylistSignatureValidationStatus` and `DenylistSignatureValidationResult`?

content/src/denylist/Denylist.ts
```diff
@@ -101,8 +113,9 @@ export class Denylist {
           this.network
         )
       })
+      return { status: OperationStatus.OK }
```
nchamo, line 116: I think we can avoid the try/catch here. Something like:
```
return new Promise((resolve) => {
validateSignature(
          metadata,
          messageToSign,
          resolve({ status: OperationStatus.OK }),
          errorMessage => resolve({ status: OperationStatus.ERROR, message: `Failed to authenticate the blocker. Error was: ${error}` }),
          (signer) => !!signer && (nodeOwner === signer || this.authenticator.isAddressOwnedByDecentraland(signer)),
          this.network
        )
})        
```

content/src/denylist/Denylist.ts
```diff
@@ -26,9 +26,12 @@ export class Denylist {
     return `${actionMessage}-${target.asString()}-${timestamp}`
   }
 
-  async addTarget(target: DenylistTarget, metadata: DenylistMetadata) {
+  async addTarget(target: DenylistTarget, metadata: DenylistMetadata): Promise<OperationResult> {
     // Validate blocker and signature
-    await this.validateSignature(DenylistAction.ADDITION, target, metadata)
+    const operationResult: OperationResult = await this.validateSignature(DenylistAction.ADDITION, target, metadata)
+    if (isSuccessfulOperation(operationResult)) {
```
nchamo, line 32: What? If the signature validation is successful then you should continue and write everything to the database

content/src/denylist/Denylist.ts
```diff
@@ -37,11 +40,15 @@ export class Denylist {
...
     })
+    return { status: OperationStatus.OK }
   }
 
-  async removeTarget(target: DenylistTarget, metadata: DenylistMetadata) {
+  async removeTarget(target: DenylistTarget, metadata: DenylistMetadata): Promise<OperationResult> {
     // Validate blocker and signature
-    await this.validateSignature(DenylistAction.REMOVAL, target, metadata)
+    const operationResult: OperationResult = await this.validateSignature(DenylistAction.REMOVAL, target, metadata)
+    if (isSuccessfulOperation(operationResult)) {
```
nchamo, line 49: Same as before

content/src/controller/Controller.ts
```diff
@@ -540,14 +540,18 @@ export class Controller {
     }
 
     try {
-      await this.denylist.addTarget(target, { timestamp, authChain })
-      res.status(201).send()
+      const result: OperationResult = await this.denylist.addTarget(target, { timestamp, authChain })
+      if (isSuccessfulOperation(result)) {
+        res.status(201).send()
+      } else {
+        res.status(400).send(result.message)
```
nchamo, line 547: Didn't you have to change existing tests with this change? Didn't we have a test that verified when a denylist action failed?

content/src/denylist/Denylist.ts
agusaldasoro: niceeee

content/src/controller/Controller.ts
agusaldasoro: It wasn't an API test, it just checked that the service failed.

content/test/integration/denylist/Denylist.spec.ts
```diff
@@ -100,6 +100,27 @@ describe('Integration - Denylist', () => {
...
+      contentPaths: ['content/test/integration/resources/some-binary-file.png']
+    })
+
+    // Deploy the entity
+    await server.deploy(deployData)
+
+    // Undenylist the entity
+    await server.undenylistEntity(entityBeingDeployed, decentralandIdentity)
+
+    // Assert that audit info marks the entity as denylisted
```
nchamo, line 116:
```suggestion
    // Assert that audit info marks the entity as not denylisted
```

nchamo (approved):
Great work!

Issue #243, opened 2020-12-18
Performance improvements to /profile lambda


nchamo:
Right now, our `/profile` lambda takes an ethAddress, fetches the profile from the content server, and sanitizes its name by querying the-graph. This last action is performed to verify whether the current name is actually an NFT or not.

Now, the problem is that it can take a few seconds to return the profile. In order to improve this, we can do 2 things:
1. We can add a cache similar to `Cache<EthAddress, Set<Name>>`, to avoid querying the-graph each time a profile is requested. This is specially useful since we consider the fact that when a user enters the world, most users around it will fetch the new user's profile from the catalyst at around the same time.
Notes: 
* This cache should have some kind of time invalidation. 
* We [had something similar](https://github.com/decentraland/catalyst/blob/4cabbfa1b4e9b7b3f206e916fb9f71aca3e8d0e6/content/src/service/caching/Cache.ts) in the content server that could help in this scenario.
2. We could add support to fetch more than one profile at a time. This would allow the explorer to avoid many requests, and at the same time we could group name queries to the-graph into one

PR #245, opened 2021-01-05
feat: Add cache to lambda profile

agusaldasoro:
Fixes https://github.com/decentraland/catalyst/issues/243

### Changes
- Added a cache to store the results of the requests to The Graph regarding name ownership. This should decrease the amount of time of getting the lambda profile.


lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -22,7 +35,7 @@ export async function getProfileById(
       returnProfile = addBaseUrlToSnapshots(fetcher.getExternalContentServerUrl(), returnProfile)
     }
   } catch {}
```
marcosnc, line 37: I'm not sure if we should keep the result in the cache in case of any error.

lambdas/src/Environment.ts
```diff
@@ -11,6 +13,8 @@ export const DEFAULT_ENS_OWNER_PROVIDER_URL_ROPSTEN =
 const DEFAULT_ENS_OWNER_PROVIDER_URL_MAINNET = 'https://api.thegraph.com/subgraphs/name/decentraland/marketplace'
 
 const DEFAULT_LAMBDAS_STORAGE_LOCATION = 'lambdas-storage'
+const DEFAULT_PROFILE_METADATA_CACHE_MAX = '500'
```
agusaldasoro, line 16: I've put an arbitrary number here, should we increase it?

marcosnc (approved):
Great work!!!

PR #247, opened 2021-01-12
Feat: files s3 buckets

agusaldasoro:


content/src/storage/ContentStorageFactory.ts
```diff
@@ -2,10 +2,34 @@ import path from 'path'
...
+      return this.local(env)
+    }
+  }
+
+  static shouldStoreInS3(env: Environment): boolean {
+    return (
+      !!env.getConfig(EnvironmentConfig.S3_STORAGE_ACCESS_KEY_ID) &&
+      !!env.getConfig(EnvironmentConfig.S3_STORAGE_SECRET_ACCESS_KEY) &&
+      !!env.getConfig(EnvironmentConfig.S3_STORAGE_BUCKET)
+    )
```
menduz, line 21: To enable S3 we should only validate if a `EnvironmentConfig.S3_STORAGE_BUCKET` is present. The good practice in deployments is to use [instanceProfile](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html) which do not require any access keys, it is like "the machine's role".. It is better in terms of security because it uses the origin (instance ip) of the requests authenticate against the role.

menduz (changes requested):
Please consider making `accessKeyId` and `secretAccessKey` optional to benefit deployments with instance profiles, which are the recommended good practice to authenticate inside AWS instances. It would also work for locally configured AWS command lines.

While access keys are the de-facto way for other environments.

PR #248, opened 2021-01-12
refactor: change how ens cache works

nchamo:
We are now changing the ENS Cache so that it looks like this: `Cache<Pair<EthAddress: string, Name: string>, Owned: boolean>`


agusaldasoro (approved):
I like it!!

PR #249, opened 2021-01-13
feat: enable fetching multiple profiles with one request

nchamo:
Right now, we only have one endpoint `/profile/:ethAddress` that fetches one profile. In order to improve the performance of the client, we want to enabling fetching more than one in one request.

The new endpoint can be used like this `/profiles?id=EthAddress&id=EthAddress...`. We also added a `/profiles/:ethAddress`, so we could deprecate `/profile` in the future

Now, in order to do that, we made two big changes:
**ENSCache**
Previously, the cache was something like `Cache<Pair<EthAddress, Name>, Owned: boolean>`. Now, we changed it to: `Cache<Name, owner: EthAddress | null>`. 

**Queries to the graph**
We want to group as many calls as possible to the graph, so that the API loads faster. To do so, we are now simply fetching by chunks of 1000 names. The result with return a pair `Pair<Name, Owner>`, which we will then use to populate the cache and return whether a given address owns a name or not.

### DIFF
--- a/lambdas/src/apis/profiles/controllers/profiles.ts
+++ b/lambdas/src/apis/profiles/controllers/profiles.ts
@@ -7,7 +7,7 @@ import { EnsOwnership } from '../EnsOwnership'
 
 const LOGGER = log4js.getLogger('profiles')
 
-export async function getProfileById(
+export async function getIndividualProfileById(
   client: SmartContentClient,
   ensOwnership: EnsOwnership,
   req: Request,
--- a/lambdas/src/apis/profiles/controllers/profiles.ts
+++ b/lambdas/src/apis/profiles/controllers/profiles.ts
@@ -1,10 +1,14 @@
-import { Entity } from 'dcl-catalyst-commons'
+import { Entity, EntityType } from 'dcl-catalyst-commons'
+import { EthAddress } from 'dcl-crypto'
 import { Request, Response } from 'express'
-import { SmartContentServerFetcher } from '../../../utils/SmartContentServerFetcher'
+import log4js from 'log4js'
+import { SmartContentClient } from '../../../utils/SmartContentClient'
 import { EnsOwnership } from '../EnsOwnership'
 
+const LOGGER = log4js.getLogger('profiles')
+
 export async function getProfileById(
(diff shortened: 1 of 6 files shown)


lambdas/test/apis/profiles/EnsOwnership.spec.ts
```diff
@@ -50,22 +56,22 @@ describe('Ensure ENS filtering work as expected', () => {
...
     const originalAddress = '0x079BED9C31CB772c4C156F86E1CFf15bf751ADd0'
     const ensOwnership = buildOwnership(fetcher)
 
-    await ensOwnership.areNamesOwned(originalAddress, ['marcosnc', 'invalid_name'])
-    await ensOwnership.areNamesOwned('anotherAddress', ['marcosnc'])
+    await ensOwnership.areNamesOwnedByAddress(originalAddress, ['marcosnc', 'invalid_name'])
+    await ensOwnership.areNamesOwnedByAddress('anotherAddress', ['marcosnc'])
 
-    verify(mockedFetcher.queryGraph(anything(), anything(), anything())).times(2)
+    verify(mockedFetcher.queryGraph(anything(), anything(), anything())).once()
```
nchamo, line 74: With the `Cache<Name, owner: Address | null>` approach, we don't have to check the graph twice! 🎉

agusaldasoro (approved):
Good job!

lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -1,83 +1,113 @@
...
+import { EthAddress } from 'dcl-crypto'
 import { Request, Response } from 'express'
-import { SmartContentServerFetcher } from '../../../utils/SmartContentServerFetcher'
+import log4js from 'log4js'
+import { SmartContentClient } from '../../../utils/SmartContentClient'
 import { EnsOwnership } from '../EnsOwnership'
 
+const LOGGER = log4js.getLogger('profiles')
+
 export async function getProfileById(
```
agusaldasoro, line 10: nit: change `getProfileById` to `getIndividualProfileById`

PR #250, opened 2021-01-14
fix: use slice correctly in profiles lambda

nchamo:
`slice` uses `(start, end)`, not `(start, count)`

Also, adding some logging to the content server start up process. Otherwise, it looks dead


pablitar (approved):
LGTM!!

Issue #199, opened 2020-10-27
Changes to Collections lambdas


marcosnc:
Currently, the lambdas for collections is downloading the image and thumbnail as a buffer instead of using streams.
The best way to fix this is by adding a new function to the `SmartContentServerFetcher` that returns a stream instead of a buffer (we're using `fetchBufferFromContentServer` now).

nchamo:
Update to this issue. 

1. This issue is now blocked by https://github.com/decentraland/catalyst-client/issues/18.
1. That change should be used [here](https://github.com/decentraland/catalyst/blob/f056edba08c0dbe878abe443b824531bd171d44f/lambdas/src/apis/collections/controllers/collections.ts#L55)
1. Since the issue was first created, we added a `SmartContentClient` to the lambdas project. Once https://github.com/decentraland/catalyst-client/issues/18 is done and the library reference is updated, lambdas will just be able to use the new method
1. Since we will be using `SmartContentClient`, it would make sense to also use it to find the entities in the content server
1. There has been an update to how the pointer needs to be fabricated. The logic should be the following

```
function buildPointer(contract: string, option: string) {
    let pointer = `${contract}-${option}`
    if (!contract.startsWith('0x')) { 
        pointer = `dcl://${pointer}`
    }
    return pointer
}
```

PR #251, opened 2021-01-18
Feat/pipe entities

agusaldasoro:
Fixes https://github.com/decentraland/catalyst/issues/199

When getting images and thumbnails instead of downloading them and sending them to the upstream, now the connection is through pipe streaming.

This PR is blocked by:
- https://github.com/decentraland/catalyst-commons/pull/18
- https://github.com/decentraland/catalyst-client/pull/30


lambdas/src/utils/SmartContentServerFetcher.ts
```diff
@@ -43,6 +43,15 @@ export class SmartContentServerFetcher extends Fetcher {
     return this.fetchBuffer(contentServerUrl + this.slash(relativeUrl), options)
   }
 
+  async fetchPipeFromContentServer(relativeUrl: string, responseTo: Response, options?: RequestOptions): Promise<any> {
```
nchamo, line 46: Why is this necessary? If it's not being used, then let's delete it

nchamo (approved):
Great work!

PR #254, opened 2021-01-21
fix: set timeout for sync fetch

nchamo:
The fetch used for syncing was set to 2m by default. Now, we are making sure that the timeout is much smaller.

This is necessary because some servers might be really slow, and we want to make sure that the sync doesn't hang, waiting to fetch everything from them. Take into account that this call is a simple json fetch, should not take more than 20s


agusaldasoro (approved):
👏

PR #255, opened 2021-01-21
chore: add logs on fallback to old content

nchamo:
We currently have a fallback in place where lambdas go to the old content server if they don't find something. 

We are not sure if this fallback is still being used, so we are adding some logs to it.

### DIFF
--- a/lambdas/src/apis/content-v2/controllers/translator.ts
+++ b/lambdas/src/apis/content-v2/controllers/translator.ts
@@ -181,6 +181,7 @@ export async function getContents(fetcher: SmartContentServerFetcher, req: Reque
       // Let's try on the old content server
       const legacyContentServerResponse = await fetch(`https://content.decentraland.org/contents/${cid}`)
       if (legacyContentServerResponse.ok) {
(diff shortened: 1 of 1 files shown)


lambdas/src/apis/content-v2/controllers/translator.ts
```diff
@@ -181,6 +181,7 @@ export async function getContents(fetcher: SmartContentServerFetcher, req: Reque
       // Let's try on the old content server
       const legacyContentServerResponse = await fetch(`https://content.decentraland.org/contents/${cid}`)
       if (legacyContentServerResponse.ok) {
+        LOGGER.info(`Failed to find '${cid}' on the content server, but found it on the legacy one.`)
```
agusaldasoro, line 184: Should this be a warning?

agusaldasoro (approved):

lambdas/src/apis/content-v2/controllers/translator.ts
nchamo: It's doing what it's supposed to be doing, I wouldn't consider the situation something that requires a warning

PR #257, opened 2021-01-28
Add documentation for content server API

agusaldasoro:
Created a MD file to store all the endpoints related to Content Server API.

The documentation is work in progress, only two endpoints are documented.

The template for the documentation is from: https://github.com/jamescooke/restapidocs

### DIFF
--- a/docs/content-server-api/CONTENT_SERVER_API.md
+++ b/docs/content-server-api/CONTENT_SERVER_API.md
@@ -2,22 +2,22 @@
 
 ## Endpoints
 
-* [E] (audit/:type/:entityId/get.md) : `GET /audit/:type/:entityId`
-* [E] (challenge/get.md) : `GET /challenge`
-* [C] (contents/hashId/get.md) : `GET /contents/:hashId`
-* [D] (contents/available-content/get.md) : `GET /available-content`
-* [E] (denylist/get.md) : `GET /denylist`
-* [E] (denylist/type/id/put.md) : `PUT /denylist/:type/:id`
-* [E] (denylist/type/id/delete.md) : `DELETE /denylist/:type/:id`
-* [E] (denylist/type/id/head.md) : `HEAD /denylist/:type/:id`
-* [E] (deployments/get.md) : `GET /deployments`
-* [B] (entities/post.md) : `POST /entities`
-* [A] (entities/type/get.md) : `GET /entities/:type`
-* [E] (failedDeployments/get.md) : `GET /failedDeployments`
-* [E] (history/get.md) : `GET /history`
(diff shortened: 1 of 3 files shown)


docs/content-server-api/deployments/get.md
```diff
@@ -0,0 +1,385 @@
...
+  - Description: This value is used as the field to order all the deployments in the collection. If no parameter is sent, then the default field to order with will be local_timestamp.
+
+- sortingOrder
+  - Format: string
+  - Value: [ASC,DESC]
+  - Example: sortingOrder=ASC
+  - Default value: DESC
+  - Description: This value is used as the field to order all the deployments in the collection. If no parameter is sent, then the default order will be desc.
+
+- showAudit
```
nchamo, line 72: Let's hide this. It was necessary for a migration, but we should even remove it from the code

docs/content-server-api/history/get.md
```diff
@@ -0,0 +1,73 @@
...
+  "pagination": {
+    "offset": 0,
+    "limit": 500,
+    "moreData": true
+  }
+}
+```
+
+
+## Notes
```
nchamo, line 73: I think we can remove this, right?

docs/content-server-api/deployments/get.md
```diff
@@ -0,0 +1,385 @@
...
+  ],
+  "filters": {},
+  "pagination": {
+    "offset": 0,
+    "limit": 500,
+    "moreData": true
+  }
+}
+
+## Notes
```
nchamo, line 385: I think we can remove this, right?

docs/content-server-api/CONTENT_SERVER_API.md
```diff
@@ -0,0 +1,23 @@
+# Content Server API
+
+## Endpoints
+
+* [Audit] (audit/:type/:entityId/get.md) : `GET /audit/:type/:entityId`
+* [Challenge] (challenge/get.md) : `GET /challenge`
+* [Contents per hashId] (contents/hashId/get.md) : `GET /contents/:hashId`
+* [Available Content] (contents/available-content/get.md) : `GET /available-content`
```
nchamo, line 8: Are these supposed to be links?

agusaldasoro: The idea of this schema is to have one file per method, in the future we should have all the endpoints+methods documented. Do you think that it's a good idea to create the files as empty and put "Work in progress" there?

nchamo: I see 2 options:
* Have links for all endpoints, and have a `Work in progress` like you said
* Have links for only the documented endpoints

I'm fine either way

nchamo (approved):
Great work!

PR #259, opened 2021-02-03
fix: use new urn format for wearables in lambda

nchamo:
We are doing 2 things in this PR:
* The `/profiles` will now always return wearable ids in urn format
* We make the wearables cache work like the ens cache. In that sense, they both now look like this: `Cache<EthAddress, NFTId | null>`. This allowed us to generalize NFT caching and remove a lot of code


lambdas/src/apis/profiles/EnsOwnership.ts
```diff
@@ -24,87 +20,17 @@ export class EnsOwnership {
...
-      offset += EnsOwnership.PAGE_SIZE
-    }
-
-    return result
+    super(maxSize, maxAge)
   }
 
   /** This method will take a list of names and return only those that are owned by the given eth address */
-  private async fetchOwnedNames(names: Name[]) {
+  protected async querySubgraph(names: Name[]) {
```
marcosnc, line 33: I don't understand the relationship between name and nft here. This function receives a list of names, but then calls those names nft. Could we keep only names or nfts? On the other hand, the old function's name was clearer in my opinion. This one is called "querySubgraph" but that doesn't give a good idea of what is it doing.

marcosnc (approved):
Great work!
I've just made a minor comment.

lambdas/src/apis/profiles/EnsOwnership.ts
nchamo: This was changed because `querySubgraph` is inherited from `NFTOwnership`. In `EnsOwnership`, those nfts are names. I can always call them nfts if that is what you prefer, but we can't avoid the method being more "generic"

PR #260, opened 2021-02-03
fix: update wearable access check with new pointer format

nchamo:
We are updating the content server access check by using the new wearable pointer format (urn)

### DIFF
--- a/content/src/service/access/AccessCheckerForWearables.ts
+++ b/content/src/service/access/AccessCheckerForWearables.ts
@@ -1,4 +1,4 @@
-import { parseUrn } from '@dcl/urn-resolver'
+import { BlockchainCollectionV2Asset, parseUrn } from '@dcl/urn-resolver'
 import { Fetcher, Pointer } from 'dcl-catalyst-commons'
 import { EthAddress } from 'dcl-crypto'
 import log4js from 'log4js'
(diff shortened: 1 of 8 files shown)


content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -19,47 +18,27 @@ export class AccessCheckerForWearables {
     }
 
     const pointer: Pointer = pointers[0].toLowerCase()
+    const parsed = await parseUrn<{ contractAddress: string; id: string; url: URL }>(pointer)
+    if (parsed) {
+      const { contractAddress: collection, id: itemId } = parsed
 
-    if (pointer.startsWith('default')) {
```
nchamo, line 23: Removed the `default` access, since there is no need for it

content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -19,47 +18,27 @@ export class AccessCheckerForWearables {
     }
 
     const pointer: Pointer = pointers[0].toLowerCase()
+    const parsed = await parseUrn<{ contractAddress: string; id: string; url: URL }>(pointer)
+    if (parsed) {
```
menduz, line 22: should the type of the parsed URN be taken into account? to avoid future complications with new format additions to `parseUrn`

nchamo: Now that I'm using the new lib version, I'm expecting `parsed` to be  `BlockchainCollectionV2Asset | null`. Is that what you are asking?

menduz: Yep, the static type assertion gives you no guarantees tho, you may also add a `parsed.type == 'blockchain-collection-v2'` condition somewhere. https://github.com/decentraland/urn-resolver/blob/main/src/types.ts#L73

marcosnc (approved):
Great work!

PR #262, opened 2021-02-04
fix: update collections endpoint to support urn

nchamo:
We are now changing our `/collections` endpoint to support urns

We previously had
```
/standard/erc721/:contract/:option/:emission?
/contents/:contract/:option/image
/contents/:contract/:option/thumbnail
```
We now have
```
/standard/erc721/:chainId/:contract/:option/:emission?
/contents/:urn/image
/contents/:urn/thumbnail
```


marcosnc (approved):
Great work!

PR #280, opened 2021-02-09
feat: adding wearables by owner endpoint

nchamo:
We are now adding a new lambda: `/collections/wearables-by-owner/:ethAddress`

Based on the ethAddress, this lambda will return the ids of the owned wearables, with their respective amount. The payload will look like this:
```
type Response = { 
  urn: WearableId,
  amount: number,
  definition?: Wearable
}[]
```
There is an optional property called `definition`. If the query param `includeDefinitions` is set on the request, then the wearable's definition will be included in the response.

**Note**
There was some refactoring done, in order to create `TheGraphClient`. This client will be in charge of handling all interactions with the graph. I noted that we were having the graph logic in lots of places, sometimes repeated, so I grouped everything under one class.


marcosnc (approved):
I love it! Great work!

PR #292, opened 2021-02-10
fix: adapting collections lambdas to use new wearables schemas

nchamo:
Now that we have a wearables schema defined (check [this](https://diagrams.menduz.com/#/notebook/znWjvuMBelUeormB3YBkkcy7MsB3/-MS9X1X3-Zy-V1SWyClm)), we now have to adapt our code to use. We are making 2 big changes here:
1. We are using the new [url-resolver](https://github.com/decentraland/urn-resolver) to translate from old wearable id format into the new one
1. When returning wearables, we are taking the metadata and adding the download url to the response, so clients don't need to understand about hashes


marcosnc (approved):
Great work!

PR #325, opened 2021-02-15
fix: map wearable id into correct one

nchamo:
During the wearables migration into the content server, we realized that a wearable had the wrong id.
We are now fixing all wearable ids that were stored in profiles. This will need to be here until we are sure that there are no more profiles with the wrong id

Also, we are undoing a lib update made by dependabot, because the new version is not working. We are making `husky` go from 5.0.9 to 4.3.0


marcosnc (approved):
Good catch.

PR #337, opened 2021-02-16
feat: adding wearables lambda

nchamo:
We are now adding a new lambda, in `/collections/wearables`

This new lambda will take some filters, and will return the definitions of all wearables that match those filters. The supported filters are:
```
{
  collectionIds?: string[] // Could either be the collection eth address, or 'base-avatars'
  wearableIds?: string[] // A list of wearable urns
  textSearch?: string // This search term currently goes through all english wearable names
}
```

_Note: the text search for on-chain wearables doesn't work right now, because the subgraph needs to be updated to be case-insensitive. We are assuming that all text will be lowercase. But when it happens, no changes are required on the lambda_


lambdas/src/utils/TheGraphClient.ts
```diff
@@ -112,7 +150,7 @@ export class TheGraphClient {
 
 const QUERY_WEARABLES_BY_OWNER: string = `
   query WearablesByOwner($owner: String, $first: Int, $skip: Int) {
-    nfts(where: {owner: $owner}, first: $first, skip: $skip) {
+    nfts(where: {owner: $owner, searchItemType_in: ["wearable_v1", "wearable_v2"]}, first: $first, skip: $skip) {
```
nchamo, line 153: I had forgotten about this when I added this query before

marcosnc (approved):
Excellent Work!

PR #359, opened 2021-02-22
fix: build lighthouse in CI

nchamo:
We are now building the lighthouse server as part of the CI, to catch any problems with the build process


pablitar (approved):
LGTM!

PR #362, opened 2021-02-22
Feat/filters wearables

agusaldasoro:
Add fields for filtering and pagination to the /wearables endpoint.


lambdas/src/apis/collections/controllers/wearables.ts
```diff
@@ -120,7 +123,18 @@ export async function getWearables(
...
 
-  return paginatedOffChainWearables.concat(onChainWearables)
+  let moreData: boolean = false
+  if (offChainWearables.length > (onChainPagination.offset + 1) * onChainPagination.limit) {
+    moreData = true
+  }
+
+  const allPagination = { ...onChainPagination, moreData: moreData }
+
+  return {
```
menduz, line 133: we can't break backwards compatibility with people using this endpoint. I understand it changed the schema of the response. 

We should have another new endpoint to handle this filtered case.

nchamo: This endpoint has not been released, nor has it been deployed to a dev catalyst. We can still make breaking changes

menduz: Thanks Nico

lambdas/src/apis/collections/controllers/wearables.ts
```diff
@@ -79,48 +79,87 @@ export async function getWearablesEndpoint(
...
+  )
+
+  // Check if there is moreData and then slice for the correct limit
+  const allWearablesWithMore = paginatedOffChainWearablesWithMore.concat(onChainWearablesWithMore)
+  const moreData: boolean = allWearablesWithMore.length > pagination.limit
+  const allWearables = allWearablesWithMore.slice(0, pagination.limit)
+
+  return {
+    wearables: allWearables,
+    filters: filters,
```
nchamo, line 141:
```suggestion
    filters,
```

lambdas/src/apis/collections/controllers/wearables.ts
```diff
@@ -79,48 +79,87 @@ export async function getWearablesEndpoint(
...
 export async function getWearables(
   filters: WearablesFilters,
   pagination: WearablesPagination,
   client: SmartContentClient,
   theGraphClient: TheGraphClient,
   offChainManager: OffChainWearablesManager
-): Promise<Wearable[]> {
+): Promise<{
+  wearables: Wearable[]
+  filters: { collectionIds?: string[]; wearableIds?: string[]; textSearch?: string }
```
nchamo, line 110: Why not use `WearablesFilters`?

lambdas/src/apis/collections/controllers/wearables.ts
```diff
@@ -79,48 +79,87 @@ export async function getWearablesEndpoint(
...
-    res.send(result)
+    const response = await getWearables(requestFilters, sanitizedPagination, client, theGraphClient, offChainManager)
+    res.send(response)
   } catch (error) {
     res.status(500).send(error.message)
   }
 }
 
+/**
+ * As Base Wearables are off chain wearables, we need to render both off-chain and on-chain wearables in the same collection.
```
nchamo, line 98:
```suggestion
 * As Base Wearables are off chain, we need to return both off-chain and on-chain wearables in the same query.
```

nchamo (approved):
Great work!

PR #384, opened 2021-02-24
fix: translate body shape to urn

nchamo:
We were translating worn wearables to urns, but forgot about the body shape.

### DIFF
--- a/lambdas/test/apis/profiles/controller/profiles.spec.ts
+++ b/lambdas/test/apis/profiles/controller/profiles.spec.ts
@@ -118,7 +118,7 @@ function profileWith(
         description: 'description',
         hasClaimedName: true,
         avatar: {
-          bodyShape: {},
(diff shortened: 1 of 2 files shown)


lambdas/test/apis/profiles/controller/profiles.spec.ts
```diff
@@ -118,7 +118,7 @@ function profileWith(
         description: 'description',
         hasClaimedName: true,
         avatar: {
-          bodyShape: {},
+          bodyShape: 'bodyShape',
```
marcosnc, line 121: From the change in `profiles.ts` I don't understand why this attribute changes from an empty object to the string 'bodyShape'.

nchamo: Because it was never an object, it was a string. The mock in the test was wrong

agusaldasoro (approved):

PR #404, opened 2021-03-01
Ensure cache update process is called even on errors during the update task.

marcosnc:
We realised that any error while updating the cache value may imply that the cache will never get updated again
This PR ensures that any error during the update process will only affect that refresh but will continue refresh the cache in the future.
As an additional feature, we're exposing the last update time.


nchamo (approved):
Look good!

Could you add a description to the PR, so that everybody can get an overview of what is being done without looking at the code?

agusaldasoro (approved):
Great work!

nchamo (approved):

PR #411, opened 2021-03-04
fix: make correct call to collections subgraph

nchamo:
When we tried to generalize how NFT ownership was checked against the subgraph, we did it incorrectly. We had a cache that looked like this: `Cache<NFTId, Owner | null>`. This worked for names, because names are 1/1 NFTs, but this wouldn't work for wearables. This is because there can be hundreds of instances of the same wearable, each with a different owner.

To fix this, we are changing the cache to `Cache<{ethAddress, NftId}, owned: boolean>`


agusaldasoro (approved):
Great Work!

PR #413, opened 2021-03-06
fix: wearables access check

nachomazzara:
Missing cast when trying to fetch items from thegraph

Screen Shot 2021-03-06 at 12 09 05


nchamo (approved):
Nice catch!

PR #419, opened 2021-03-15
BREAKING CHANGE: disable denylist by default

nchamo:
We are now disabling the denylist by default


agusaldasoro (approved):
Should we change the title to breaking change?

marcosnc:
Please, create a new issue (or check if we already have one) and link it here to re-think the denylist and re-enable it.

marcosnc (approved):

marcosnc:
I've created this issue to re-enable the deny-list: https://github.com/decentraland/catalyst/issues/420

PR #430, opened 2021-03-22
BREAKING CHANGE: remove history and origin information

nchamo:
We are removing the `/history` endpoint from the content server, and we will no longer be updating the `originServerUrl` and the `originTimestamp` properties. From now on, they will be hardcoded.

In the near future, we will just delete them


package.json
```diff
@@ -108,8 +108,8 @@
     "aws-sdk": "^2.854.0",
     "compression": "^1.7.4",
     "cors": "^2.8.5",
-    "dcl-catalyst-client": "^3.2.0",
-    "dcl-catalyst-commons": "^2.1.0",
+    "dcl-catalyst-client": "^4.0.3-20210322145441.commit-8c2b9c7",
```
nchamo, line 111: This is on purpose. Once [this change](https://github.com/decentraland/catalyst-client/pull/75) is tested and merged, we will release a new version of the library and then update this dependency

agusaldasoro (approved):

PR #436, opened 2021-03-29
fix: use client version with sync fix

nchamo:
We are doing the following changes:
* Using the client version with [this fix](https://github.com/decentraland/catalyst-client/pull/77)
* Increasing the db-client pool size to 20
* Logging the amount of pending queries every 1 minute


pablitar (approved):
LGTM!

PR #437, opened 2021-03-30
fix: make profiles work even if subgraph is down

nchamo:
We are now making a small change so that the `/profile` and `/profiles` endpoints continue to work, even if the subgraph is down.

In that scenario, all names and wearables will be considered valid and will remain unfiltered.


pablitar (approved):
LGTM!

PR #438, opened 2021-03-31
feat: add support for L2 wearables in lambdas

nchamo:
We are making some important changes here. First, we are adding support for L2 wearables on the `/wearables` endpoint. But we are also making some other changes that will really improve how the endpoint works:
1. We removed the offset, and returning a `next` property that contains the url for the next page. In this particular endpoint, we will be using a `lastId` for pagination
2. If the request only asks for wearable ids, we are not checking the subgraph anymore. We are querying the content server directly, so this should provide more reliability 
3. If the request only asks for off chain wearable ids, we are not even checking the content server (since they are all in-memory)


lambdas/test/apis/collections/controller/wearables/wearables-by-owner.spec.ts
```diff
@@ -0,0 +1,107 @@
+import { getWearablesByOwner } from '@katalyst/lambdas/apis/collections/controllers/wearables'
```
nchamo, line 1: We split the previous test file into two. `wearables-by-owner` and `wearables`

pablitar (approved):
LGTM!

PR #440, opened 2021-04-05
feat: add new collections endpoint

nchamo:
We are now adding a new `/collections` endpoint that will return all available collections.

The payload will look like this:
```
{
    "collections": [
        {
            "id": "urn:decentraland:off-chain:base-avatars",
            "name": "Base Wearables"
        },
        ...
    ]
}
```
The idea was to make it like this, in order to easily add pagination/filters in the future consistently with other endpoints


agusaldasoro (approved):
Please add a test that executes this new endpoint

PR #441, opened 2021-04-05
chore: make config consistent

nchamo:
We are making two changes here:
1. We are setting the version as 1.0.0, since we will be releasing a new version soon
2. We are making config consistent between lambdas & the content server. 
Now, by setting `COLLECTIONS_L1_SUBGRAPH_URL` or `COLLECTIONS_L2_SUBGRAPH_URL` as environment variables, both services will read the same config. Previously, each service's config had a different name

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM node:12.18.1-slim
 
 RUN apt-get update && \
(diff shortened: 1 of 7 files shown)


agusaldasoro (approved):
I like it!

PR #442, opened 2021-04-05
fix: use X-Forwarded-Host header to create next url

agusaldasoro:
Uses[ 'X-Forwarded-Host' header ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host)to build the url of the next page in the colection, instead of using host. As host is the local host from the nginx, but the forwarded host is the host that the original request was made to.

This change has been tested in peer-testing with the following configuration: https://github.com/decentraland/catalyst-owner/pull/46


nchamo (approved):
Great work!

PR #445, opened 2021-04-05
feat: use correct subgraph when validating wearables

nchamo:
We are now supporting both L1 and L2 to validate wearable ownership. But, we are checking one or the other, depending on the network in the urn

### DIFF
--- a/lambdas/src/utils/TheGraphClient.ts
+++ b/lambdas/src/utils/TheGraphClient.ts
@@ -224,7 +224,7 @@ export class TheGraphClient {
     pagination: { limit: number; lastId: string | undefined }
   ): Promise<WearableId[]> {
     // Order will be L1 > L2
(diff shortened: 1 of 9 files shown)


lambdas/src/utils/TheGraphClient.ts
```diff
@@ -224,7 +224,7 @@ export class TheGraphClient {
     pagination: { limit: number; lastId: string | undefined }
   ): Promise<WearableId[]> {
     // Order will be L1 > L2
-    const L1_NETWORKS = ['ethereum', 'ropsten', 'kovan', 'rinkeby', 'goerli']
+    const L1_NETWORKS = ['mainnet', 'ropsten', 'kovan', 'rinkeby', 'goerli']
```
nchamo, line 227: This was a bug, since the reported network is actually `"mainnet"`

agusaldasoro (approved):
Great work!

PR #448, opened 2021-04-06
feat: expose all lambdas metrics

nchamo:
We are now exposing metrics for all lambda endpoints. The main problem was that lambdas use routers, so we couldn't calculate the `req.route.path` correctly. 

The solution was to create a `Router` proxy (an ES6 Proxy) that would inject the metrics handler when each route is defined. 
How it looks now:
Screen Shot 2021-04-06 at 16 10 29

### DIFF
--- a/commons/servers/metrics.ts
+++ b/commons/servers/metrics.ts
@@ -37,12 +37,12 @@ export class Metrics {
 
   private static requestCounters = function (req, _res, next) {
     numRequests.inc({ method: req.method })
-    pathsTaken.inc({ path: req.route.path })
+    pathsTaken.inc({ path: req.baseUrl + req.route.path })
     next()
   }
 
   private static responseCounters = ResponseTime(function (req, res, time) {
-    responses.labels(req.method, req.route.path, res.statusCode).observe(time)
+    responses.labels(req.method, req.baseUrl + req.route.path, res.statusCode).observe(time)
   })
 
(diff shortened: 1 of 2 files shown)


commons/servers/metrics.ts
```diff
@@ -37,12 +37,12 @@ export class Metrics {
...
   private static requestCounters = function (req, _res, next) {
     numRequests.inc({ method: req.method })
-    pathsTaken.inc({ path: req.route.path })
+    pathsTaken.inc({ path: req.baseUrl + req.route.path })
     next()
   }
 
   private static responseCounters = ResponseTime(function (req, res, time) {
-    responses.labels(req.method, req.route.path, res.statusCode).observe(time)
+    responses.labels(req.method, req.baseUrl + req.route.path, res.statusCode).observe(time)
```
agusaldasoro, line 45: What does this do?

agusaldasoro (approved):
Great work!

commons/servers/metrics.ts
nchamo: In most cases (like comms or content server) it does nothing. But in lambdas, we registered a router like this

```
this.app.use('/collections', router...)
```

And inside the router, we did:
```
router.get('/wearables', (req, res) => ...)
```

So `req.route.path` would only return '/wearables', but adding the baseUrl makes it contain '/collections also

Issue #443, opened 2021-04-05
Add a “next” property to the paginated endpoints so clients don’t use the offset and simply use this property


agusaldasoro:
We've found that using the offset property to paginate results in performance issues, we need to change the pagination of our endpoints to use the property next as an iterator for the next page of the current collection.

This change needs to be done in the endpoints:
- /content/deployments
- /content/pointerChanges

PR #452, opened 2021-04-07
Use next link to paginate

agusaldasoro:
This PR is blocked by https://github.com/decentraland/catalyst-commons/pull/64

Fixes https://github.com/decentraland/catalyst/issues/443

We want to deprecate pagination using `offset`, to do that we need to provide a new way to paginate collections: the `next` link.
That link will contain the relative URL to the next page of the collection.


content/src/service/deployments/DeploymentManager.ts
```diff
@@ -80,11 +91,46 @@ export class DeploymentManager {
...
+      }
+    } else {
+      if (order == SortingOrder.ASCENDING) {
+        nextFilters.fromEntityTimestamp = lastDeployment.entityTimestamp
+      } else {
+        nextFilters.toEntityTimestamp = lastDeployment.entityTimestamp
+      }
+    }
+
+    const nextQueryParams = qs.stringify(
```
menduz, line 121: Every time you need to create or parse a query string, you can use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams), it doesn't require a new package and the API is stable and as standard as `fetch`

```ts
qs = new URLSearchParams({key: values} | "?string=true")
newQueryString = '?' + qs.toString()
```

menduz: This was not a request for change

nchamo: The main problem with `URLSearchParams` are multi params. 

If you have `deploymentId: string[]`, `URLSearchParams` would transform it into `deploymentId=id1,id2,id3`. But, our API uses the following format: `deploymentId=id1&deploymentId=id2&...`. 

That's why we use `{ arrayFormat: 'repeat' }` as an option for `qs.stringify`. Sadly, `URLSearchParams` doesn't have that option available

content/test/integration/service/deployments/pointer-changes-check.spec.ts
```diff
@@ -64,6 +64,15 @@ describe('Integration - Pointer Changes Check', () => {
     assertChangesAre(changes, [P1, { before: E1, after: undefined }], [P2, { before: E2, after: undefined }])
   })
 
+  it('When getting a page of pointer changes then next url is set', async () => {
+    const E1Timestamp = await deployEntitiesCombo(service, E1, E3)
+
+    const { pagination: pagination } = await service.getPointerChanges({ entityTypes: [E3.entity.type] }, undefined, 1)
+
+    expect(pagination.next).toContain(`toLocalTimestamp=${E1Timestamp}`)
```
nchamo, line 72: Could you also make sure that the lastId in `next` is set to the E1's id? If it's not set, you would get the same result

content/src/storage/repositories/DeploymentsRepository.ts
```diff
@@ -71,14 +78,52 @@ export class DeploymentsRepository {
       offset
     }
 
+    if (lastEntityId) {
+      values.lastEntityId = lastEntityId
+    }
+
+    /**  The lastEntityId is a field that we only want to compare with if for pagination.
```
nchamo, line 85: Is `if` a typo?

content/test/integration/service/deployments/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+    expect(nextLink).toBeUndefined()
+  })
+
+  it('When lastEntityId is sent, it stills orders by timestamp', async () => {
+    // Deploy E1, E2 and E3 in that order
+    const [, E2Timestamp] = await deploy(E1, E2, E3)
+
+    const actualDeployments = await service.getDeployments({
+      limit: 2,
+      lastEntityId: E1.entity.id
```
nchamo, line 53: Does it make sense to test `lastEntityId` without a to/from timestamp filter? I mean, is it even used if none of those filters are present?

Also, let's make sure that we use if all 4 filters. from/to entity/local timestamp. And let's make sure that the returned deployments are actually the ones we expect.

Could make sense to make those tests in another test file, but we should do them

content/src/storage/repositories/DeploymentsRepository.ts
```diff
@@ -71,14 +78,52 @@ export class DeploymentsRepository {
...
+     * If the filter specifies a timestamp value that it's repeated among many deployments,
+     * then to know where the page should start we will use the lastEntityId.
+     */
+    const pageBorder: string =
+      (order == SortingOrder.ASCENDING ? 'from' : 'to') +
+      (timestampField == SortingField.ENTITY_TIMESTAMP ? 'EntityTimestamp' : 'localTimestamp')
```
nchamo, line 91: Shouldn't `localTimestamp` be `LocalTimestamp`? If it should, then we don't have enough tests.

Note: I added a comment about it in the test file

nchamo, line 91:
```suggestion
      (order === SortingOrder.ASCENDING ? 'from' : 'to') +
      (timestampField === SortingField.ENTITY_TIMESTAMP ? 'EntityTimestamp' : 'localTimestamp')
```

content/src/storage/repositories/DeploymentsRepository.ts
```diff
@@ -127,6 +172,12 @@ export class DeploymentsRepository {
     }))
   }
 
+  private createOrClause(timestampField: string, compare: string, timestampFilter: string): string {
```
nchamo, line 175: Great work!

nchamo (changes requested):
Great work, but we still have to make some more changes

content/src/controller/Controller.ts
```diff
@@ -354,6 +356,7 @@ export class Controller {
     const fields: string | undefined = req.query.fields
     const sortingField: SortingField | undefined | 'unknown' = this.asEnumValue(SortingField, req.query.sortingField)
     const sortingOrder: SortingOrder | undefined | 'unknown' = this.asEnumValue(SortingOrder, req.query.sortingOrder)
+    const lastEntityId: string | undefined = req.query.lastEntityId?.toLowerCase()
```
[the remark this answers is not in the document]
agusaldasoro: What do you think of having `toTimestamp` and `fromTimestamp` only? And the sorting field is the one that indicates which timestamp to look at.

content/src/controller/ControllerDeploymentFactory.ts
```diff
@@ -21,6 +21,7 @@ export class ControllerDeploymentFactory {
     if (fields.includes(DeploymentField.AUDIT_INFO)) {
       result.auditInfo = auditInfo
     }
+    result.local_timestamp = auditInfo.localTimestamp
```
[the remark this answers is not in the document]
agusaldasoro: Yes, it will be repeated in order to leave the audit info just the same as it is now. The snake case is a typo 😄

content/src/controller/Controller.ts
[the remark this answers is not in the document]
nchamo: Yeah, that sounds like a really smart idea!

But,
1. Let's make sure we don't break compatibility
2. Let's update the docs

content/src/service/deployments/DeploymentManager.ts
```diff
@@ -80,11 +91,46 @@ export class DeploymentManager {
...
         limit: curatedLimit,
-        moreData: moreData
+        moreData: moreData,
+        lastEntityId: options?.lastEntityId,
+        next: nextRelativePath
       }
     }
   }
 
+  private calculateNextRelativePath(options: DeploymentOptions | undefined, lastDeployment: Deployment): string {
```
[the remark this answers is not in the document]
agusaldasoro: I agree, but if we are changing that then we should move all the logic for creating pagination data to the controller, right?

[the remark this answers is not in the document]
nchamo: I'm not sure what you mean with creating pagination data. I mean, at the service layer we could return an object called `nextFilters` if you want. Or we could just generate it on the controller based on the filters sent to the deployment manager.

content/src/storage/repositories/DeploymentsRepository.ts
agusaldasoro: yay

content/test/integration/service/deployments/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+    const a = await buildDeployData([P1], { type, timestamp, metadata: 'metadata1' })
+    const b = await buildDeployData([P1], { type, timestamp, metadata: 'metadata2' })
+    if (a.entity.id.toLowerCase() < b.entity.id.toLowerCase()) {
+      E1 = a
+      E2 = b
+    } else {
+      E1 = b
+      E2 = a
+    }
+    E3 = await buildDeployData([P1], { type, timestamp, metadata: 'metadata3' })
```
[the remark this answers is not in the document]
agusaldasoro: You were right, I've added a new (older) timestamp for deploying E3.

content/test/integration/service/deployments/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+    const result: Timestamp[] = []
+    for (const { deployData } of entities) {
+      const newAuditInfo = { version: EntityVersion.V2, authChain: deployData.authChain, ...overrideAuditInfo }
+      const deploymentResult: DeploymentResult = await service.deployEntity(
+        Array.from(deployData.files.values()),
+        deployData.entityId,
+        newAuditInfo,
+        ''
+      )
+      if (isSuccessfulDeployment(deploymentResult)) {
```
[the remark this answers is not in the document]
agusaldasoro: I'm using the timestamp from the response to test the filters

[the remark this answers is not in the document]
nchamo: But shouldn't we fail if one of the deployments was unsuccessful? I mean, for the test to make sense all deployments need to be successful, right?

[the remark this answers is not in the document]
agusaldasoro: True

commons/servers/QueryParameters.ts
```diff
@@ -0,0 +1,73 @@
+import { DeploymentFilters, EntityType, SortingField, SortingOrder } from 'dcl-catalyst-commons'
+import qs from 'qs'
+
+export function toQueryParamsForGetAllDeployments(
```
nchamo, line 4: Instead of manually remove the 's' manually, maybe we could do it programmatically? We are doing something similar here
https://github.com/decentraland/catalyst-client/blob/master/src/utils/Helper.ts#L176

That should make things a lot simpler here

docs/content-server-api/pointerChanges/get.md
```diff
@@ -0,0 +1,76 @@
...
+
+**Auth required** : NO
+
+**Permissions required** : NO
+
+**Query Parameters** :
+
+
+
+*DEPRECATED: PLEASE USE from*
```
nchamo, line 19: Maybe it makes sense to avoid adding these? I mean, adding documentation about deprecated parameters sounds weird to me. The less people know about these, the easier it will be to remove them

docs/content-server-api/pointerChanges/get.md
```diff
@@ -0,0 +1,76 @@
...
+  - Example: to=1606829553969
+  - Default value: NULL - no upper filter is set
+  - Description: Acts as a filter in the collection of deployments, this value is the maximum value of timestamp (of the field indicated by SortingField: localTimestamp is the default) that any deployment in the collection will have.
+
+- lastId
+  - Format: string
+  - Value: EntityId
+  - Example: lastId=QmNknKv8MuKbfZ73z4QdUEsNbTd1ZAN1fSuwTFGiNGeCt5
+  - Default value: -
+  - Description: It is the last entity id that will be listed in the collection
```
nchamo, line 54: If `lastId` is set, then it won't be listed at all, it should be avoided. Right? Or am I getting something wrong?

docs/content-server-api/deployments/get.md
```diff
@@ -90,6 +107,13 @@ List all deployments stored in the content server
   - Default value: 500
   - Description: The deployments are a paginated collection, this parameter corresponds to the limit for each page.
 
+- lastId
+  - Format: string
+  - Value: EntityId
+  - Example: lastId=QmNknKv8MuKbfZ73z4QdUEsNbTd1ZAN1fSuwTFGiNGeCt5
+  - Default value: -
+  - Description: It is the last entity id that will be listed in the collection
```
nchamo, line 115: If `lastId` is set, then it won't be listed at all, it should be avoided. Right? Or am I getting something wrong?

docs/content-server-api/pointerChanges/get.md
```diff
@@ -0,0 +1,76 @@
+# Get Pointer Changes
+
+List all deployment changes from the same pointer
```
nchamo, line 3: This description is kind of weird. It doesn't list all changes for the same pointer. It lists all changes made to pointers in general

docs/content-server-api/pointerChanges/get.md
```diff
@@ -0,0 +1,76 @@
...
+
+**URL** : `/pointerChanges`
+
+<!-- **URL Parameters** : `pk=[integer]` where `pk` is the ID of the Account on the server. -->
```
nchamo, line 7: What's this?

docs/content-server-api/CONTENT_SERVER_API.md
```diff
@@ -15,7 +15,7 @@
 * Get Entities: `GET /entities/:type`
 * Failed Deployments: `GET /failedDeployments`
 * [History](history/get.md) : `GET /history`
-* Pointer Changes: `GET /pointerChanges'`
+* [Pointer Changes](pointerChanges/get.md) : `GET /pointerChanges'`
```
nchamo, line 18: Thank you so much for adding this!!!

content/test/helpers/service/MockedMetaverseContentService.ts
```diff
@@ -146,16 +146,6 @@ export class MockedMetaverseContentService implements MetaverseContentService {
     return MockedMetaverseContentService.STATUS
   }
 
-  getLegacyHistory(
```
nchamo, line 149: Nice catch!

content/src/controller/Controller.ts
```diff
@@ -304,34 +313,65 @@ export class Controller {
...
+
+    if (controllerPointerChanges.length > 0 && pagination.moreData) {
+      const lastPointerChange = controllerPointerChanges[controllerPointerChanges.length - 1]
+      pagination.next = this.calculateNextRelativePathForPointer(lastPointerChange, pagination.limit, filters)
+    }
+
     res.send({ deltas: controllerPointerChanges, filters, pagination })
   }
 
+  calculateNextRelativePathForPointer(
```
nchamo, line 360:
```suggestion
 private calculateNextRelativePathForPointerChange(
```

content/src/controller/Controller.ts
```diff
@@ -304,34 +313,65 @@ export class Controller {
...
+    lastPointerChange: ControllerPointerChanges,
+    limit: number,
+    filters?: PointerChangesFilters
+  ): string | undefined {
+    const nextFilters = Object.assign({}, filters)
```
nchamo, line 365: Is this copy necessary? You are not modifying `nextFilters`

content/src/service/deployments/DeploymentManager.ts
```diff
@@ -40,11 +40,21 @@ export class DeploymentManager {
...
         : DeploymentManager.MAX_HISTORY_LIMIT
 
+    const curatedFilters = Object.assign({}, options?.filters)
+
+    if (!options?.sortBy?.field || options?.sortBy?.field != SortingField.ENTITY_TIMESTAMP) {
+      curatedFilters.from = curatedFilters.fromLocalTimestamp
+      curatedFilters.to = curatedFilters.toLocalTimestamp
+    }
+    curatedFilters.fromLocalTimestamp = undefined
+    curatedFilters.toLocalTimestamp = undefined
```
nchamo, line 50: Is this necessary? Aren't you doing this at the controller?

content/src/service/deployments/DeploymentManager.ts
```diff
@@ -200,8 +217,12 @@ export type DeploymentOptions = {
   sortBy?: DeploymentSorting
   offset?: number
   limit?: number
+  lastId?: string
 }
 
-export type PointerChangesFilters = Pick<DeploymentFilters, 'fromLocalTimestamp' | 'toLocalTimestamp' | 'entityTypes'>
+export type PointerChangesFilters = Pick<
+  DeploymentFilters,
+  'from' | 'to' | 'fromLocalTimestamp' | 'toLocalTimestamp' | 'entityTypes'
```
nchamo, line 225: Would it make sense to exclude 'fromLocalTimestamp' & 'toLocalTimestamp'?

content/src/storage/repositories/DeploymentsRepository.ts
```diff
@@ -66,19 +73,54 @@ export class DeploymentsRepository {
...
+    /**  The lastId is a field that we only want to compare with when paginating.
+     * If the filter specifies a timestamp value that it's repeated among many deployments,
+     * then to know where the page should start we will use the lastId.
+     */
+    const pageBorder: string =
+      (order === SortingOrder.ASCENDING ? 'from' : 'to') +
+      (timestampField === SortingField.ENTITY_TIMESTAMP ? 'EntityTimestamp' : 'LocalTimestamp')
+
+    if (filters?.from && timestampField == SortingField.LOCAL_TIMESTAMP) {
+      values.fromLocalTimestamp = filters.fromLocalTimestamp ?? filters.from
```
nchamo, line 93: Can't you just use `filters.from` here? I think you already handled this

content/src/storage/repositories/DeploymentsRepository.ts
```diff
@@ -66,19 +73,54 @@ export class DeploymentsRepository {
...
+      if (pageBorder == 'fromLocalTimestamp' && lastId) {
+        whereClause.push(this.createOrClause('local_timestamp', '>', 'fromLocalTimestamp'))
+      } else {
+        whereClause.push(`dep1.local_timestamp >= to_timestamp($(fromLocalTimestamp) / 1000.0)`)
+      }
+    }
+    if (filters?.to && timestampField == SortingField.LOCAL_TIMESTAMP) {
+      values.toLocalTimestamp = filters.toLocalTimestamp ?? filters.to
```
nchamo, line 101: Can't you just use `filters.to` here? I think you already handled this

content/test/integration/controller/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+      server.getAddress() +
+      `/deployments?` +
+      toQueryParamsForGetAllDeployments(
+        options.filters,
+        options.sortBy?.field,
+        options.sortBy?.order,
+        options.lastId,
+        options.limit
+      )
+    const response = await fetcher.fetchJson(url)
```
nchamo, line 303: I think you can use the `fetchJson` expose by the library directly, without having to instantiate a `Fetcher`

content/test/integration/controller/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+  let E1: EntityCombo, E2: EntityCombo, E3: EntityCombo
+
+  const testEnv = loadStandaloneTestEnvironment()
+  let server: TestServer
+  const fetcher = new Fetcher()
+
+  beforeEach(async () => {
+    server = await testEnv
+      .configServer()
+      .withBean(Bean.SYNCHRONIZATION_MANAGER, new MockedSynchronizationManager())
```
nchamo, line 25: I think it will be better if you use
```
.withConfig(EnvironmentConfig.DISABLE_SYNCHRONIZATION, true)
```

content/test/integration/controller/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+    const b = await buildDeployData([P2], { type, timestamp, metadata: 'metadata2' })
+    if (a.entity.id.toLowerCase() < b.entity.id.toLowerCase()) {
+      E1 = a
+      E2 = b
+    } else {
+      E1 = b
+      E2 = a
+    }
+    const laterTimestamp = Date.now()
+    E3 = await buildDeployData([P3], { type, timestamp: laterTimestamp, metadata: 'metadata3' })
```
nchamo, line 47:
```suggestion
    E3 = await buildDeployData([P3], { type, timestamp: timestamp + 1, metadata: 'metadata3' })
```

content/test/integration/controller/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+    const actualDeployments = await fetchDeployments({
+      limit: 3
+    })
+
+    expect(actualDeployments.deployments.length).toBe(3)
+    const nextLink = actualDeployments.pagination.next
+    expect(nextLink).toBeUndefined()
+  })
+
+  it('When local timestamp filter is set, then only toLocalTimestamp is modified in next ', async () => {
```
nchamo, line 63: I think we need to update the query param name now

content/test/integration/controller/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+    // Deploy E1, E2 and E3 in that order
+    const [E1Timestamp, E2Timestamp, E3Timestamp] = await deploy(E1, E2, E3)
+
+    const actualDeployments = await fetchDeployments({
+      limit: 2,
+      filters: { fromLocalTimestamp: E1Timestamp, toLocalTimestamp: E3Timestamp }
```
nchamo, line 69: I would recommend using `to` and `from` with the `sortBy`. That will make deprecation easier

(Also in other tests)

content/test/integration/controller/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+  it('When order is by entity timestamp, then only to is modified in next', async () => {
+    // Deploy E1, E2 and E3 in that order
+    const [E1Timestamp, , E3Timestamp] = await deploy(E1, E2, E3)
+
+    const actualDeployments = await fetchDeployments({
+      limit: 2,
+      sortBy: {
+        field: SortingField.ENTITY_TIMESTAMP
+      },
+      filters: { fromLocalTimestamp: E1Timestamp, toLocalTimestamp: E3Timestamp }
```
nchamo, line 120: Is this valid? Using `fromLocalTimestamp` or `toLocalTimestamp` when sorting with entity timestamp? Shouldn't it be `from` & `to`?

content/test/integration/controller/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+
+    expect(actualDeployments.deployments.length).toBe(2)
+    expect(actualDeployments.deployments[0].entityId).toBe(E3.entity.id)
+    expect(actualDeployments.deployments[1].entityId).toBe(E2.entity.id)
+    const nextLink = actualDeployments.pagination.next
+    expect(nextLink).toContain(`to=${E2.entity.timestamp}`)
+    expect(nextLink).toContain(`lastId=${E2.entity.id}`)
+  })
+
+  it('When getting pointer Changes then the pagination is correctly done', async () => {
```
nchamo, line 250:
```suggestion
  it('When getting pointer changes then the pagination is correctly done', async () => {
```

content/test/integration/controller/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+
+    const pointerChanges = await fetchPointerChanges({ from: E1Timestamp }, 1)
+
+    expect(pointerChanges.deltas.length).toBe(1)
+    expect(pointerChanges.pagination.next).toContain(`to=${E2Timestamp}`)
+    expect(pointerChanges.pagination.next).toContain(`from=${E1Timestamp}`)
+    expect(pointerChanges.pagination.next).toContain(`lastId=${E2.entity.id}`)
+  })
+
+  it('When getting pointer Changes with deprecated filter then the pagination is correctly done', async () => {
```
nchamo, line 262:
```suggestion
  it('When getting pointer changes with deprecated filter then the pagination is correctly done', async () => {
```

content/test/integration/controller/deployment-pagination.spec.ts
```diff
@@ -0,0 +1,323 @@
...
+    expect(pointerChanges.pagination.next).toContain(`to=${E2Timestamp}`)
+    expect(pointerChanges.pagination.next).toContain(`from=${E1Timestamp}`)
+    expect(pointerChanges.pagination.next).toContain(`lastId=${E2.entity.id}`)
+  })
+
+  async function deploy(...entities: EntityCombo[]): Promise<Timestamp[]> {
+    const result: Timestamp[] = []
+    for (const { deployData } of entities) {
+      const deploymentResult = await server.deploy(deployData)
+      if (isSuccessfulDeployment(deploymentResult)) {
```
nchamo, line 283: This isn't necessary at the controller level. The controller will return an error

nchamo (approved):
Great work!

docs/content-server-api/pointerChanges/get.md
agusaldasoro: You mean that the users shouldn't use that param unless they are using our auto-generated pagination?

docs/content-server-api/pointerChanges/get.md
agusaldasoro: It's the template for documenting URL params, I'm deleting it as we are not using it.

content/src/controller/Controller.ts
agusaldasoro: True!

content/src/service/deployments/DeploymentManager.ts
agusaldasoro: I didn't want to delete them as they are currently supported on that endpoint. What do you think?

content/test/integration/controller/deployment-pagination.spec.ts
agusaldasoro: True, I forgot to fail on that case 😄

content/src/service/deployments/DeploymentManager.ts
nchamo: I think that you already handled that at the controller level, and you are defining a type used at the service layer. You will never get those parameters here, right?

content/test/integration/controller/deployment-pagination.spec.ts
agusaldasoro: I wanted to be sure that it still works, I'll add a separate test for that then.

PR #455, opened 2021-04-13
feat: adding concurrent and pending db requests limit

nchamo:
We are now adding a limit to the queue that stores pending requests that will be made to the database. We set the amount of concurrent requests to 20, and the max amount of queued requests to 50.

We've also added a concept of priority. This means that requests with a high priority will be queued regardless of the amount of queued requests. But requests with low priority will be rejected in that case.

### How did we do it?
The `Repository` used to be an instance of the database client directly. We now renamed that instance to `Database`, and we created a wrapper as a `Repository`. This wrapper has a priority queue inside, that will be used to store pending requests and enforce the limit described above

#### How did we decide priorities?
The idea is rather simple. All requests that are necessary for the content server to work internally, like snapshots, garbage collection, sync/deployments, etc are considered high priority.

All external requests are considered low priority

## What's next?
We will explore how to add timeouts both to the database client and the database, so the queue is cleared up more quickly


content/src/storage/RepositoryFactory.ts
```diff
@@ -23,6 +25,7 @@ export class RepositoryFactory {
       }
     }
 
-    return build(connection, contentCredentials, rootCredentials)
+    const database = await build(connection, contentCredentials, rootCredentials)
+    return new Repository(database, new RepositoryQueue())
```
agusaldasoro, line 29: can we add two env variables to configure the queue in case we need to override the defaults?

agusaldasoro (approved):
Great work!

Maybe it's a thing to do separately, anyway it'd be great to have a dashboard with the status of the queue, some metrics that come to my mind are:
- Current elements in the queue by priority
- Current elements running by priority
- Rejected queries. by priority

PR #460, opened 2021-04-16
chore: stop creating db in content server

nchamo:
Before this change, the content server tried to connect to the database during startup. If it failed, it would connect as the root user, and create the database for itself. This was clearly an anti-pattern.

Thanks to https://github.com/decentraland/catalyst-owner/pull/53, we now have the ability to run psql scripts. We added a script that creates the database there, and stop giving that responsibility to the content server. That is done in https://github.com/decentraland/catalyst-owner/pull/55

### DIFF
--- a/content/src/storage/Database.ts
+++ b/content/src/storage/Database.ts
@@ -39,29 +39,8 @@ export type DBCredentials = {
 /**
  * Builds the database client by connecting to the content database. If it isn't present, then it tries to connect with the root user and creates the content database and user.
  */
-export async function build(
-  connection: DBConnection,
-  contentCredentials: DBCredentials,
-  rootCredentials?: DBCredentials
-): Promise<FullDatabase> {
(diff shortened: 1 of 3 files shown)


content/src/storage/Database.ts
```diff
@@ -39,29 +39,8 @@ export type DBCredentials = {
 /**
  * Builds the database client by connecting to the content database. If it isn't present, then it tries to connect with the root user and creates the content database and user.
```
agusaldasoro, line 40: Should we update the comment?

agusaldasoro (approved):
Excelent!!

content/src/storage/Database.ts
nchamo: Yeah, definitely. Good catch!

menduz:
This was SO nice

PR #462, opened 2021-04-20
feat: adding cache for entities

nchamo:
We are now adding a cache to the `/entities` endpoint. The `/deployments` endpoint is too generic, has too many filters and is usually used to get historic deployments, so we decided to had the cache directly to the `/entities` endpoint. 

To do so we added `getEntitiesByIds` and `getEntitiesByPointers` to the service, who will be in charge of caching.

This cache will help a lot in avoiding requests to the database. Since users tend to group together in the world, they will ask for the same scenes & profiles, so the cache makes a lot of sense

### DIFF
--- a/content/src/Environment.ts
+++ b/content/src/Environment.ts
@@ -8,6 +8,7 @@ import { FetcherFactory } from './helpers/FetcherFactory'
 import { MigrationManagerFactory } from './migrations/MigrationManagerFactory'
 import { AccessCheckerImplFactory } from './service/access/AccessCheckerImplFactory'
 import { AuthenticatorFactory } from './service/auth/AuthenticatorFactory'
+import { CacheManagerFactory } from './service/caching/CacheManagerFactory'
 import { DeploymentManagerFactory } from './service/deployments/DeploymentManagerFactory'
 import { FailedDeploymentsManager } from './service/errors/FailedDeploymentsManager'
 import { GarbageCollectionManagerFactory } from './service/garbage-collection/GarbageCollectionManagerFactory'
@@ -120,7 +121,8 @@ export const enum Bean {
   MIGRATION_MANAGER,
   GARBAGE_COLLECTION_MANAGER,
   SYSTEM_PROPERTIES_MANAGER,
-  SNAPSHOT_MANAGER
+  SNAPSHOT_MANAGER,
+  CACHE_MANAGER
 }
 
 export enum EnvironmentConfig {
@@ -159,7 +161,8 @@ export enum EnvironmentConfig {
   DISABLE_DENYLIST,
   CONTENT_SERVER_ADDRESS,
   REPOSITORY_QUEUE_MAX_CONCURRENCY,
-  REPOSITORY_QUEUE_MAX_QUEUED
+  REPOSITORY_QUEUE_MAX_QUEUED,
+  CACHE_SIZES
 }
 
 export class EnvironmentBuilder {
@@ -357,6 +360,16 @@ export class EnvironmentBuilder {
       () => process.env.REPOSITORY_QUEUE_MAX_QUEUED ?? 50
     )
 
+    /*
+     * These are configured as 'CACHE_{CACHE_NAME}_{ENTITY_TYPE}=MAX_SIZE'.
+     * For example: 'CACHE_ENTITIES_BY_POINTERS_SCENE=1000
+     */
+    this.registerConfigIfNotAlreadySet(
+      env,
+      EnvironmentConfig.CACHE_SIZES,
+      () => new Map(Object.entries(process.env).filter(([name]) => name.startsWith('CACHE')))
+    )
+
     // Please put special attention on the bean registration order.
     // Some beans depend on other beans, so the required beans should be registered before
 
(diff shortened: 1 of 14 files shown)


content/src/service/caching/Cache.ts
```diff
@@ -0,0 +1,77 @@
...
+      .filter(([, value]) => value === undefined)
+      .map(([key]) => key)
+
+    // Calculate values for those keys
+    const calculated: Map<Key, Value | undefined> = missing.length > 0 ? await orCalculate(missing) : new Map()
+
+    // Save the calculated values. When a key doesn't have a value, set it as null
+    Array.from(calculated.entries())
+      .map<[Key, Value | null]>(([key, value]) => [key, value ?? null])
+      .forEach(([key, value]) => this.internalCache.set(key, value))
```
menduz, line 33: `for(const [key, value] of calculated)` is streamable (every iteration consumes the iterator) and memory efficient in comparison to this currying approach

content/src/Environment.ts
```diff
@@ -357,6 +360,16 @@ export class EnvironmentBuilder {
       () => process.env.REPOSITORY_QUEUE_MAX_QUEUED ?? 50
     )
 
+    /*
+     * These are configured as 'CACHE_{CACHE_NAME}_{ENTITY_TYPE}=MAX_SIZE'.
```
agusaldasoro, line 364: Should this key be something more specific like `ENTITIES_CACHE`? Just in case we have another cache configuration in the future.

content/src/service/caching/CacheManager.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/** This manger is used to create cache instances, based on some pre-configured defaults */
+export class CacheManager {
+  private readonly cacheSizes: Map<string, number>
+
+  constructor(cacheSizes: Map<string, number> = new Map()) {
+    this.cacheSizes = new Map(Array.from(cacheSizes.entries()).map(([name, size]) => [name.toUpperCase(), size]))
+  }
+
+  buildEntityTypedCache<K, V>(cacheConfig: CacheConfig): CacheByType<K, V> {
```
agusaldasoro, line 12: why do you need `K` and `V` here?

content/src/denylist/DenylistServiceDecorator.ts
```diff
@@ -238,6 +238,23 @@ export class DenylistServiceDecorator implements MetaverseContentService {
     return this.service.listenToDeployments(listener)
   }
 
+  getEntitiesByIds(ids: EntityId[]): Promise<Entity[]> {
+    return this.repository.task(async (task) => {
```
agusaldasoro, line 242: Is it necessary to run something similar to `filterDenylisted()` here?

content/src/denylist/DenylistServiceDecorator.ts
```diff
@@ -282,6 +299,44 @@ export class DenylistServiceDecorator implements MetaverseContentService {
...
+    targetBuild: (element: T) => DenylistTarget
+  ): Promise<T[]> {
+    const elementToTarget: Map<T, DenylistTarget> = new Map(elements.map((element) => [element, targetBuild(element)]))
+    const areDenylisted = await this.denylist.areTargetsDenylisted(denylistRepo, Array.from(elementToTarget.values()))
+    return Array.from(elementToTarget.entries())
+      .filter(([, target]) => !isTargetDenylisted(target, areDenylisted))
+      .map(([element]) => element)
+  }
+
+  /** When an entity is denylisted, we don't want to show its content and metadata  */
```
agusaldasoro, line 315: The entities that are used as param here have already been filtered, so is this method to filter the denylisted nested objects on those?

agusaldasoro (approved):
Great work!

content/src/Environment.ts
```diff
@@ -357,6 +360,16 @@ export class EnvironmentBuilder {
...
+     * For example: 'CACHE_ENTITIES_BY_POINTERS_SCENE=1000
+     */
+    this.registerConfigIfNotAlreadySet(
+      env,
+      EnvironmentConfig.CACHE_SIZES,
+      () => new Map(Object.entries(process.env).filter(([name]) => name.startsWith('CACHE')))
```
agusaldasoro, line 370: By default the caches are not being used, right?

nchamo: Yes, they are being used. This is just to be able to override the max sizes of the caches

content/src/Environment.ts
nchamo: In this line I'm explaining the schema. The `{CACHE_NAME}` for the cache that is added in this PR is `ENTITIES_BY_POINTERS`. It's defined in `CacheManager`

content/src/service/caching/CacheManager.ts
nchamo: It's because you want to receive a `CacheByType<K, V>`. So you need to pass them to `buildEntityTypedCache`

content/src/denylist/DenylistServiceDecorator.ts
nchamo: No, it isn't necessary.

When a pointer is denylisted, then no entities are returned for it. But when an entity is denylisted, it is returned, but the content files are not shown, and the metadata only says `'Denylisted'`. All that is done in `sanitizeEntities`, and that's why it isn't necessary to filter the entity ids

content/src/denylist/DenylistServiceDecorator.ts
nchamo: As I said before, entities are not filtered when denylisted. We just hide some of their properties. That "hiding" is done here

content/src/Environment.ts
agusaldasoro: Okay, my doubt was because it is filtering all environment variables that start with only `CACHE`

PR #466, opened 2021-04-22
chore: add active user metrics

nchamo:
We are now adding a new metric to report the amount of users per layer


pablitar (approved):
LGTM!

PR #479, opened 2021-05-06
Fix: Deployed by filter capitalization

### LINKED ISSUE #291: The deployedBy filter in the deployments endpoint is not filtering the deployments properly
See this case:
https://peer.decentraland.org/content/entities/profile?pointer=0xa7c825bb8c2c4d18288af8efe38c8bf75a1aab51

This call returns many deployments (207 at the time when this issue was created):
https://peer.decentraland.org/content/deployments?entityType=profile&pointer=0xa7c825bb8c2c4d18288af8efe38c8bf75a1aab51

However this other call is returning only one deployment:
https://peer.decentraland.org/content/deployments?entityType=profile&deployedBy=0xa7c825bb8c2c4d18288af8efe38c8bf75a1aab51

@nchamo suggested that the problem may be caused by the deployedBy filter, which may take into account the word casing of the filter address.

jmoguilevsky:
This PR:
- Introduces a new index for deployer_address for the lowercase of the column, since we'll be searching ignoring case, but saving values as they come
- Creates a test suite for DeploymentRepository, achieving ~100% coverage mocking the database
- Fixes a bug in the `createOrClause` method where the timestamp type was hardcoded and didn't change based on the filter type
- Modifies the query for deployer_address so that it looks for the LOWER of the incoming address
- Modifies some queries from multiline to single line to make them more testable

Fixes #291

### DIFF
--- a/content/src/migrations/scripts/1620221157277_add_lowercase_deployer_index.ts
+++ b/content/src/migrations/scripts/1620221157277_add_lowercase_deployer_index.ts
@@ -10,5 +10,6 @@ export async function up(pgm: MigrationBuilder): Promise<void> {
 }
 
 export async function down(pgm: MigrationBuilder): Promise<void> {
+  pgm.addIndex('deployments', 'deployer_address')
   pgm.sql(`DROP INDEX deployer_address_lower_case;`)
 }
--- a/content/src/migrations/scripts/1620221157277_add_lowercase_deployer_index.ts
+++ b/content/src/migrations/scripts/1620221157277_add_lowercase_deployer_index.ts
@@ -1,0 +1,13 @@
+import { MigrationBuilder } from 'node-pg-migrate'
+
+/*
+  In this migration, we are creating a index in deployments table with the 'entity_timestamp' and 'entity_id' columns
+ */
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`CREATE INDEX deployer_address_lower_case on deployments (LOWER(deployer_address) text_pattern_ops);`)
+}
+
+export async function down(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`DROP INDEX deployer_address_lower_case;`)
+}
--- a/content/src/migrations/scripts/1620221157277_add_lowercase_deployer_index.ts
+++ b/content/src/migrations/scripts/1620221157277_add_lowercase_deployer_index.ts
@@ -5,6 +5,7 @@ import { MigrationBuilder } from 'node-pg-migrate'
  */
 
 export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`DROP INDEX deployments_deployer_address_index;`)
   pgm.sql(`CREATE INDEX deployer_address_lower_case on deployments (LOWER(deployer_address) text_pattern_ops);`)
 }
 
--- a/content/src/repository/extensions/DeploymentsRepository.ts
+++ b/content/src/repository/extensions/DeploymentsRepository.ts
@@ -184,15 +184,10 @@ export class DeploymentsRepository {
     entityType: EntityType
   ): Promise<{ entityId: EntityId; pointers: Pointer[]; localTimestamp: Timestamp }[]> {
     return this.db.map(
-      `
-            SELECT
-                entity_id,
-                entity_pointers,
-                date_part('epoch', local_timestamp) * 1000 AS local_timestamp
-            FROM deployments
-            WHERE entity_type = $1 AND deleter_deployment IS NULL
-            ORDER BY local_timestamp DESC, entity_id DESC
-            `,
+      `SELECT entity_id, entity_pointers, date_part('epoch', local_timestamp) * 1000 AS local_timestamp ` +
+        `FROM deployments ` +
+        `WHERE entity_type = $1 AND deleter_deployment IS NULL ` +
+        `ORDER BY local_timestamp DESC, entity_id DESC`,
       [entityType],
       (row) => ({
         entityId: row.entity_id,
@@ -204,10 +199,9 @@ export class DeploymentsRepository {
 
   deploymentsSince(entityType: EntityType, timestamp: Timestamp): Promise<number> {
     return this.db.one(
-      `
-            SELECT COUNT(*) AS count
-            FROM deployments
-            WHERE entity_type = $1 AND local_timestamp > to_timestamp($2 / 1000.0)`,
+      `SELECT COUNT(*) AS count ` +
+        `FROM deployments ` +
+        `WHERE entity_type = $1 AND local_timestamp > to_timestamp($2 / 1000.0)`,
       [entityType, timestamp],
       (row) => row.count
     )
@@ -215,30 +209,10 @@ export class DeploymentsRepository {
 
   saveDeployment(entity: Entity, auditInfo: AuditInfo, overwrittenBy: DeploymentId | null): Promise<DeploymentId> {
     return this.db.one(
-      `
-            INSERT INTO deployments (
-                deployer_address,
-                version,
-                entity_type,
-                entity_id,
-                entity_timestamp,
-                entity_pointers,
-                entity_metadata,
-                local_timestamp,
-                auth_chain,
-                deleter_deployment
-            ) VALUES (
-                $(deployer),
-                $(auditInfo.version),
-                $(entity.type),
-                $(entity.id),
-                to_timestamp($(entity.timestamp) / 1000.0),
-                $(entity.pointers),
-                $(metadata),
-                to_timestamp($(auditInfo.localTimestamp) / 1000.0),
-                $(auditInfo.authChain:json),
-                $(overwrittenBy)
-            ) RETURNING id`,
+      `INSERT INTO deployments (deployer_address, version, entity_type, entity_id, entity_timestamp, entity_pointers, entity_metadata, local_timestamp, auth_chain, deleter_deployment)` +
+        ` VALUES ` +
+        `($(deployer), $(auditInfo.version), $(entity.type), $(entity.id), to_timestamp($(entity.timestamp) / 1000.0), $(entity.pointers), $(metadata), to_timestamp($(auditInfo.localTimestamp) / 1000.0), $(auditInfo.authChain:json), $(overwrittenBy))` +
+        ` RETURNING id`,
       {
         entity,
         auditInfo,
--- a/content/src/repository/extensions/DeploymentsRepository.ts
+++ b/content/src/repository/extensions/DeploymentsRepository.ts
@@ -27,7 +27,6 @@ export class DeploymentsRepository {
     )
 
     const deployedIds = new Set(result)
-    console.log('deployedIds', deployedIds)
     return new Map(entityIds.map((entityId) => [entityId, deployedIds.has(entityId)]))
   }
 
--- a/content/src/repository/extensions/DeploymentsRepository.ts
+++ b/content/src/repository/extensions/DeploymentsRepository.ts
@@ -50,7 +50,7 @@ export class DeploymentsRepository {
     lastId?: string
   ) {
     const sorting = Object.assign({ field: SortingField.LOCAL_TIMESTAMP, order: SortingOrder.DESCENDING }, sortBy)
-    return this.getDeploymentsBy(sorting?.field, sorting?.order, offset, limit, filters, lastId)
+    return this.getDeploymentsBy(sorting.field, sorting.order, offset, limit, filters, lastId)
   }
 
   private getDeploymentsBy(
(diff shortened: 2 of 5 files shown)


content/src/migrations/scripts/1620221157277_add_lowercase_deployer_index.ts
```diff
@@ -0,0 +1,13 @@
+import { MigrationBuilder } from 'node-pg-migrate'
+
+/*
+  In this migration, we are creating a index in deployments table with the 'entity_timestamp' and 'entity_id' columns
+ */
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`CREATE INDEX deployer_address_lower_case on deployments (LOWER(deployer_address) text_pattern_ops);`)
```
agusaldasoro, line 8: Should we delete the original index on `deployer_address` ?

content/src/repository/extensions/DeploymentsRepository.ts
```diff
@@ -175,23 +176,18 @@ export class DeploymentsRepository {
 
   private createOrClause(timestampField: string, compare: string, timestampFilter: string): string {
     const equalWithEntityIdComparison = `(LOWER(dep1.entity_id) ${compare} LOWER($(lastId)) AND dep1.${timestampField} = to_timestamp($(${timestampFilter}) / 1000.0))`
-    const timestampComparison = `(dep1.entity_timestamp ${compare} to_timestamp($(${timestampFilter}) / 1000.0))`
+    const timestampComparison = `(dep1.${timestampField} ${compare} to_timestamp($(${timestampFilter}) / 1000.0))`
```
agusaldasoro, line 179: 👏

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+
+      beforeEach(() => {
+        when(db.map(anything(), anything(), anything())).thenReturn(Promise.resolve(dbResult))
+      })
+
+      it('should return a map of entity id to the deployment status', async () => {
+        const entities = ['1', '3']
+        const result = await repository.areEntitiesDeployed(entities)
+
+        expect(result).toEqual(
```
agusaldasoro, line 33: I'd remove this as the only thing that you are testing is the mock

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+      result = await repository.getAmountOfDeployments()
+    })
+
+    it('should call the database with the expected query', () => {
+      verify(
+        db.map(`SELECT entity_type, COUNT(*) AS count FROM deployments GROUP BY entity_type`, deepEqual([]), anything())
+      ).once()
+    })
+
+    it('should return a Map with the result', () => {
```
agusaldasoro, line 78: Same as before

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+
+    beforeEach(async () => {
+      db = mock(MockedDataBase)
+      repository = new DeploymentsRepository(instance(db) as any)
+
+      when(db.map(anything(), anything(), anything())).thenReturn(Promise.resolve(expectedResult))
+      result = await repository.getAmountOfDeployments()
+    })
+
+    it('should call the database with the expected query', () => {
```
agusaldasoro, line 72: Let's give more detail here, something like `the db makes a count`

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+      ).once()
+    })
+
+    it('should return a Map with the result', () => {
+      expect(result).toEqual(new Map(expectedResult as any))
+    })
+  })
+
+  describe('getHistoricalDeployments', () => {
+    beforeEach(() => {
```
agusaldasoro, line 84: Is this `beforeEach` necessary?

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+    })
+
+    describe("when it doesn't receive a lastId", () => {
+      beforeEach(() => {
+        db = mock(MockedDataBase)
+        repository = new DeploymentsRepository(instance(db) as any)
+      })
+
+      describe('when it receives a field or order to sort by', () => {
+        beforeEach(() => {
```
agusaldasoro, line 96: I'm not sure why you need the nested beforeEach, should this be all one only describe?

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+            {
+              field: SortingField.ENTITY_TIMESTAMP,
+              order: SortingOrder.ASCENDING
+            }
+          )
+
+          const args = capture(db.map).last()
+
+          expect(args[0]).toContain(`ORDER BY dep1.entity_timestamp ASC`)
+          expect(args[0]).toContain(
```
agusaldasoro, line 114: niceeeeee 👏

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+
+        expect(result).toEqual(
+          new Map([
+            ['1', true],
+            ['3', false]
+          ])
+        )
+
+        verify(
+          db.map('SELECT entity_id FROM deployments WHERE entity_id IN ($1:list)', deepEqual([entities]), anything())
```
agusaldasoro, line 41: can we test the content of the params in the query too?

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+          )
+          expect(args[1]).toEqual(jasmine.objectContaining({ fromLocalTimestamp: 1, toLocalTimestamp: 11 }))
+        })
+      })
+    })
+
+    describe('when it receives a lastId', () => {
+      const lastId = '1'
+
+      beforeEach(() => {
```
agusaldasoro, line 143: As we don't really care about the response from the MockedDataBase can we configure this only once for the entire file?

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+    describe('when there is a deployed by filter', () => {
+      beforeEach(() => {
+        db = mock(MockedDataBase)
+        repository = new DeploymentsRepository(instance(db) as any)
+
+        when(db.map(anything(), anything(), anything())).thenReturn(Promise.resolve([]))
+      })
+
+      it('should add the expected where clause to the query with the addresses on lowercase', async () => {
+        const deployedBy = ['jOn', 'aGus']
```
agusaldasoro, line 211: 🤣

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+        await repository.getHistoricalDeployments(0, 10, { deployedBy })
+
+        const args = capture(db.map).last()
+
+        expect(args[0]).toContain(`LOWER(dep1.deployer_address) IN ($(deployedBy:list))`)
+        expect(args[1]).toEqual(jasmine.objectContaining({ deployedBy: deployedBy.map((x) => x.toLowerCase()) }))
```
agusaldasoro, line 217: nit: as this is the expected param, shouldn't it be hardcoded?

agusaldasoro (changes requested):
Great Work! Just a few comments, also do you mind adding an integration test that executes the same behaviour as described in the bug report?

content/test/unit/repository/DeploymentsRepository.spec.ts
jmoguilevsky: This is asserting on the logic of the map creation in the return statement of the function

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+          ])
+        )
+
+        verify(
+          db.map('SELECT entity_id FROM deployments WHERE entity_id IN ($1:list)', deepEqual([entities]), anything())
+        ).once()
+      })
+    })
+
+    describe('when the entities list is empty', () => {
```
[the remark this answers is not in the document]
jmoguilevsky: This is asserting on the first if of the function (https://github.com/decentraland/catalyst/blob/774344fb9ff5fd3d27adf29fb09bab7d136050a4/content/src/repository/extensions/DeploymentsRepository.ts#L20) that's why It also asserts on `never called` of the mock

content/test/unit/repository/DeploymentsRepository.spec.ts
jmoguilevsky: Done

content/test/unit/repository/DeploymentsRepository.spec.ts
jmoguilevsky: As before, this asserts that the response is a map with the result of the DB call as the content

content/test/unit/repository/DeploymentsRepository.spec.ts
jmoguilevsky: Yes, deleted the nested beforeEaches

content/test/unit/repository/DeploymentsRepository.spec.ts
jmoguilevsky: This one is, not the inner ones

content/test/unit/repository/DeploymentsRepository.spec.ts
jmoguilevsky: Updated to have it hardcoded

content/test/unit/repository/DeploymentsRepository.spec.ts
jmoguilevsky: We can, but that makes the tests coupled since the mock is the same among all tests

jmoguilevsky: That's why I initialized it in each test context

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+    })
+
+    describe('when the entities list is not empty', () => {
+      const dbResult = ['1', '2']
+
+      beforeEach(() => {
+        when(db.map(anything(), anything(), anything())).thenReturn(Promise.resolve(dbResult))
+      })
+
+      it('should return a map of entity id to the deployment status', async () => {
```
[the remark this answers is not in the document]
jmoguilevsky: This is the correct verification, splitted the test for two different verifications

content/test/unit/repository/DeploymentsRepository.spec.ts
agusaldasoro: My bad!

content/test/unit/repository/DeploymentsRepository.spec.ts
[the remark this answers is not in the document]
agusaldasoro: My bad!

content/src/migrations/scripts/1620221157277_add_lowercase_deployer_index.ts
```diff
@@ -0,0 +1,14 @@
...
+  In this migration, we are creating a index in deployments table with the 'entity_timestamp' and 'entity_id' columns
+ */
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`DROP INDEX deployments_deployer_address_index;`)
+  pgm.sql(`CREATE INDEX deployer_address_lower_case on deployments (LOWER(deployer_address) text_pattern_ops);`)
+}
+
+export async function down(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`DROP INDEX deployer_address_lower_case;`)
```
agusaldasoro, line 13: If we are maintaining those down functions then we should create the original index here

content/test/unit/repository/DeploymentsRepository.spec.ts
```diff
@@ -0,0 +1,402 @@
...
+  let repository: DeploymentsRepository
+  let db: MockedDataBase
+
+  beforeEach(() => {
+    db = mock(MockedDataBase)
+    repository = new DeploymentsRepository(instance(db) as any)
+  })
+
+  describe('areEntitiesDeployed', () => {
+    beforeEach(() => {
```
agusaldasoro, line 17: Is this necessary as you have the beforeEach above with the same code?

agusaldasoro (approved):
Great work, especially creating the repository tests! 👏

Just check the two comments and then you are free to merge.

PR #484, opened 2021-05-11
Don't access denylist in db when disabled

agusaldasoro:
As part of improving the performance of the Content Server, we've decided to avoid connecting to the DB when it wasn't necessary. 

This PR does changes for the servers with the denylist disabled:
- When adding or removing an entity from the denylist, then an error is thrown.
- When asking for deny-listed items, then none are shown.

Also, avoid checking if entities are deny-listed or not when the list is empty.


content/src/controller/Controller.ts
```diff
@@ -568,6 +568,11 @@ export class Controller {
     // Path: /denylist/{type}/{id}
     // Body: JSON with ethAddress, signature and timestamp
 
+    if (this.denylist.isDisabled()) {
+      res.status(500).send()
```
agusaldasoro, line 572: I didn't want to expose much information, so it would be visible whether or not the denylist is enabled.

content/src/denylist/Denylist.ts
```diff
@@ -1,141 +1,41 @@
...
 import { DenylistTarget, DenylistTargetId, DenylistTargetType } from './DenylistTarget'
 
-export class Denylist {
-  constructor(
-    private readonly repository: Repository,
-    private readonly authenticator: ContentAuthenticator,
-    private readonly cluster: ContentCluster,
-    private readonly network: string
-  ) {}
+export abstract class Denylist {
```
nchamo, line 6: Just curious, any reason why this is an abstract class and not an interface?

content/src/denylist/DummyDenylist.ts
```diff
@@ -0,0 +1,40 @@
+import { DenylistRepository } from '../repository/extensions/DenylistRepository'
+import {
+  Denylist,
+  DenylistMetadata,
+  DenylistSignatureValidationResult,
+  DenylistSignatureValidationStatus
+} from './Denylist'
+import { DenylistTarget, DenylistTargetId, DenylistTargetType } from './DenylistTarget'
+
+export class DummyDenylist extends Denylist {
```
nchamo, line 10: Nit: maybe NoopDenylist would be a little bit clearer?

content/src/denylist/DenylistServiceDecorator.ts
```diff
@@ -60,9 +60,12 @@ export class DenylistServiceDecorator implements MetaverseContentService {
     }
   }
 
-  /** Is content is denylisted, then we will return that it is not available */
+  /** If content is denylisted, then we will return that it is not available */
   async isContentAvailable(fileHashes: ContentFileHash[]): Promise<Map<string, boolean>> {
     const availability: Map<ContentFileHash, boolean> = await this.service.isContentAvailable(fileHashes)
+    if (await denylistEmpty(this.denylist)) {
```
nchamo, line 66: I think that instead of doing this here, the denylist itself should manage this scenario.

This is what I'm thinking: when the denylist is created or the first time it's used, it can detect if it there is something denylisted or not. If there isn't, then it works as a noop/dummy denylist. If there is something, then it works normally. If something is added to the denylist, it will know and it can change from one mode to the other by itself.

By doing that, you can avoid having that responsibility spread all over this decorator. And if we add a new method, we can't forget to add this `if`

nchamo (commented):
A few comments

content/src/denylist/Denylist.ts
agusaldasoro: I used an abstract class as I left the static implementations inside it, does it make sense?

content/src/denylist/ActiveDenylist.ts
```diff
@@ -0,0 +1,149 @@
...
+      async (transaction) => {
+        // Remove denylist
+        await transaction.denylist.removeTarget(target)
+
+        // Add to history
+        await transaction.denylist.addEventToHistory(target, metadata, DenylistAction.REMOVAL)
+      },
+      { priority: DB_REQUEST_PRIORITY.HIGH }
+    )
+    this.size--
```
nchamo, line 75: Writing it here, but have the same concern for `addTarget`. What happens if you remove the same target many times? Does it fail the second time? If not, then this might not be the best way to keep track of whether there is something on the denylist

content/src/denylist/DenylistFactory.ts
```diff
@@ -1,9 +1,15 @@
 import { Bean, Environment, EnvironmentConfig } from '../Environment'
+import { ActiveDenylist } from './ActiveDenylist'
 import { Denylist } from './Denylist'
+import { NoopDenylist } from './NoopDenylist'
 
 export class DenylistFactory {
   static create(env: Environment): Denylist {
-    return new Denylist(
+    if (env.getConfig(EnvironmentConfig.DISABLE_DENYLIST)) {
```
nchamo, line 8: Could you remove the logic that is in the `ControllerFactory` that used to do this? It makes no sense to have both

content/src/denylist/NoopDenylist.ts
```diff
@@ -0,0 +1,40 @@
...
+  Denylist,
+  DenylistMetadata,
+  DenylistSignatureValidationResult,
+  DenylistSignatureValidationStatus
+} from './Denylist'
+import { DenylistTarget, DenylistTargetId, DenylistTargetType } from './DenylistTarget'
+
+export class NoopDenylist extends Denylist {
+  async addTarget(target: DenylistTarget, metadata: DenylistMetadata): Promise<DenylistSignatureValidationResult> {
+    return { status: DenylistSignatureValidationStatus.ERROR }
```
nchamo, line 12: We probably should rename this, since this is not a signature validation anymore. Also, would it make sense to add a message explaining why we are failing?

content/src/denylist/NoopDenylist.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+  async removeTarget(target: DenylistTarget, metadata: DenylistMetadata): Promise<DenylistSignatureValidationResult> {
+    return { status: DenylistSignatureValidationStatus.ERROR }
```
nchamo, line 15: Same as above

content/test/integration/denylist/NoopDenylist.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  const decentralandIdentity = createIdentity()
+  const ownerIdentity = createIdentity()
+  const testEnv = loadStandaloneTestEnvironment()
+  let server: TestServer
+
+  beforeEach(async () => {
+    server = await testEnv
+      .configServer()
+      .withBean(Bean.SYNCHRONIZATION_MANAGER, new MockedSynchronizationManager())
+      .withBean(Bean.CONTENT_CLUSTER, MockedContentCluster.withAddress(ownerIdentity.address))
```
nchamo, line 20: I don't think you need this

content/test/integration/denylist/NoopDenylist.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  const metadata: string = 'Some metadata'
+  const decentralandIdentity = createIdentity()
+  const ownerIdentity = createIdentity()
+  const testEnv = loadStandaloneTestEnvironment()
+  let server: TestServer
+
+  beforeEach(async () => {
+    server = await testEnv
+      .configServer()
+      .withBean(Bean.SYNCHRONIZATION_MANAGER, new MockedSynchronizationManager())
```
nchamo, line 19: I think you can just use the config that disables the sync instead of adding the mock

nchamo (commented):
Adding some comments

content/src/denylist/NoopDenylist.ts
agusaldasoro: Does it make sense to throw an exception with the error message and then the controller will handle the response?

nchamo: I think the current approach makes sense, so we can differentiate between some unwanted error, and a validation error

agusaldasoro: Done!

content/test/integration/Server.spec.ts
```diff
@@ -42,11 +43,13 @@ describe('Integration - Server', function () {
       .registerBean(Bean.REPOSITORY, MockedRepository.build())
       .registerBean(Bean.SERVICE, service)
       .registerBean(Bean.SYNCHRONIZATION_MANAGER, new MockedSynchronizationManager())
+      .registerBean(Bean.DENYLIST, new DeactivatedDenylist())
```
nchamo, line 46: Why are you adding this and the config? I think that the config be enough on it's own, but why was it necessary in the first place?

content/src/denylist/DeactivatedDenylist.ts
```diff
@@ -0,0 +1,49 @@
...
+  DenylistValidationType
+} from './Denylist'
+import { DenylistTarget, DenylistTargetId, DenylistTargetType } from './DenylistTarget'
+
+export class DeactivatedDenylist extends Denylist {
+  async addTarget(target: DenylistTarget, metadata: DenylistMetadata): Promise<DenylistOperationResult> {
+    return {
+      status: DenylistOperationStatus.ERROR,
+      type: DenylistValidationType.CONFIGURATION,
+      message: 'Enable target from denylist is not activated.'
```
nchamo, line 16: Could you make this a little bit clearer? Maybe something like "Denylist is not activated, so you can't add a target"

content/src/denylist/DeactivatedDenylist.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+  async removeTarget(target: DenylistTarget, metadata: DenylistMetadata): Promise<DenylistOperationResult> {
+    return {
+      status: DenylistOperationStatus.ERROR,
+      type: DenylistValidationType.CONFIGURATION,
+      message: 'Remove target from denylist is not activated.'
```
nchamo, line 23: Same here

content/test/integration/denylist/DeactivatedDenylist.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      .andBuild()
+
+    await server.start()
+  })
+
+  it(`When an entity is denylisted, then an error is thrown`, async () => {
+    // Prepare entity to deploy
+    const { deployData, controllerEntity: entityBeingDeployed } = await buildDeployData(['0,0', '0,1'], {
+      metadata,
+      contentPaths: ['content/test/integration/resources/some-binary-file.png']
```
nchamo, line 27: Just a nit. I think you can avoid setting the metadata and the content paths to this entity (and the ones in other tests). It would make the test a little bit more readable

nchamo (commented):
A few minor comments

content/test/integration/Server.spec.ts
agusaldasoro: Because this emulates the behaviour under the `DeploymentFactory`, I needed to add this for the test cases to work.

nchamo: We will use the activated denylist, and remove the config.

Also, we will enable the denylist by default

nchamo (approved):

PR #493, opened 2021-05-26
fix: refresh base wearables daily

nchamo:
Before this change, when we fetched base-wearables in our lambdas, we only did it once, and we cached it forever. This is because we assumed that we wouldn't have to make changes to base-wearables (and that had been the case since the launch).

But, we recently made some changes to the wearables, and the cache was a problem. The only way to avoid this problem was to restart the catalyst.

So now, instead of caching the base wearables forever, we are using a `TimeRefreshedDataHolder`. This holder fetches the new definitions of the base wearables in a daily manner.


Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN yarn install &&\
 WORKDIR /app
 
 ENV COMMIT_HASH=bc34832282cfa746cfb1f27184cf3b53f321a164
-ENV CATALYST_VERSION=1.1.1
+ENV CATALYST_VERSION=1.1.2
```
nchamo, line 30: We are releasing a new version with this fix

lambdas/src/apis/collections/off-chain/OffChainWearablesManager.ts
```diff
@@ -10,17 +10,21 @@ import baseAvatars from './base-avatars'
...
-  private definitions: IFuture<LocalOffChainWearables>
+  private readonly definitions: TimeRefreshedDataHolder<LocalOffChainWearables>
 
   constructor(
-    private readonly client: SmartContentClient,
-    private readonly collections: OffChainCollections = DEFAULT_COLLECTIONS
-  ) {}
+    client: SmartContentClient,
+    collections: OffChainCollections = DEFAULT_COLLECTIONS,
+    refreshTime: string = '1d'
```
agusaldasoro, line 18: Can we make the refresh time configurable?

agusaldasoro (approved):

lambdas/src/apis/collections/off-chain/OffChainWearablesManager.ts
nchamo: Done!

PR #494, opened 2021-05-26
Fix: Change default Comms server url in the lambdas service

jmoguilevsky:
In this PR: 
- Change default internal url to use port 9000
- Make the comms server URL configurable
- Load the URL at startup time


lambdas/src/Environment.ts
```diff
@@ -124,11 +128,15 @@ export class EnvironmentBuilder {
...
-      EnvironmentConfig.COMMS_SERVER_ADDRESS,
-      () => process.env.COMMS_SERVER_ADDRESS
+
+    const realCommsServerAddress = await getCommsServerUrl(
+      EnvironmentBuilder.LOGGER,
+      process.env.INTERNAL_COMMS_SERVER_ADDRESS ?? DEFAULT_INTERNAL_COMMS_SERVER_URL,
+      process.env.COMMS_SERVER_ADDRESS
     )
+
+    this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.COMMS_SERVER_ADDRESS, () => realCommsServerAddress)
```
agusaldasoro, line 138: Please check where was `COMMS_SERVER_ADDRESS` being used as we are sure that we didn't break anything with this change

lambdas/src/apis/status/PeerHealthStatus.ts
```diff
@@ -12,13 +11,12 @@ export default class PeerHealthStatus {
...
   private commsServerStatus: TimeRefreshedDataHolder<HealthStatus>
   private lambdaServerStatus: TimeRefreshedDataHolder<HealthStatus>
-  private commsServerUrl: string
 
   constructor(
     contentClient: SmartContentClient,
     maxSynchronizationTime: string,
     maxDeploymentObtentionTime: string,
-    private readonly externalCommsServerAddress: string
+    private readonly commsServerAddress: string
```
agusaldasoro, line 19: Where is this constructor being called? Do we need to update to receive this new parameter?

jmoguilevsky: This was used  [here](https://github.com/decentraland/catalyst/blob/f2e4aa49229bc9bb6e082a1113f5609b2ccdf51c/lambdas/src/apis/status/routes.ts#L11) and since the variable is still named the same, it works

lambdas/src/Environment.ts
jmoguilevsky: This was only used  [here](https://github.com/decentraland/catalyst/blob/f2e4aa49229bc9bb6e082a1113f5609b2ccdf51c/lambdas/src/apis/status/routes.ts#L11) and since the variable is still named the same, it works

agusaldasoro (approved):

jmoguilevsky:
Tested it in zone and it's working

PR #495, opened 2021-05-27
feat: support different max sizes depending on entity type

nchamo:
Before this change, all entity types would have the same 15MB per entity time limit. Now, it can be different, depending on the entity type.

**Important: we don't want this to be configurable per catalyst, since that could lead to a lot of failures during the sync**


content/src/service/validations/Validations.ts
```diff
@@ -22,15 +22,28 @@ import { happenedBefore } from '../time/TimeSorting'
 import { Validation, ValidationContext } from './ValidationContext'
 
 export class Validations {
+  private static MAX_UPLOAD_SIZE_PER_POINTER_MB: Map<EntityType, number> = new Map([
+    [EntityType.SCENE, 15],
+    [EntityType.PROFILE, 15], // TODO: Investigate and update profiles to a more appropriate number
+    [EntityType.WEARABLE, 2.5]
```
nchamo, line 28: This value might change later today, but the rest is ready for review

menduz:
If it is a change that may impact synchronzation, should we timestamp it? to make it available from XXX date and time? In the same fashion as protocol upgrades in Ethereum start at X block in the future to enable every implementation and deployment to upgrade.

nchamo:
> If it is a change that may impact synchronzation, should we timestamp it? to make it available from XXX date and time? In the same fashion as protocol upgrades in Ethereum start at X block in the future to enable every implementation and deployment to upgrade.

Only new wearables will go through this validation, so that approach is not needed in this case

agusaldasoro (approved):

PR #489, opened 2021-05-20
Add health check

jmoguilevsky:
Add a health check endpoint to the catalyst service to know if the service is available to receive connections.

In this PR:
- Added `/health` to the lambdas service
- Added a new method in the lambdas service
- Added a commons helper (it only includes getCommsUrl)
- Make use of `TimeRefreshedDataHolder` for the health of each service so that the health endpoint can not be DDosed 
- Added doc for the endpoint

PR #497, opened 2021-05-31
feat: remove content server check

nchamo:
We added a new healthcheck endpoint in https://github.com/decentraland/catalyst/pull/489, that will help our clients determine whether they should connect to a specific catalyst or not.

Since this new endpoint exists, we can remove the responsibility from the lighthouse


pablitar (commented):
This will create a hard dependency with having to use the health check on explorer when selecting the realm. Have we done that already?

I'd suggest we only merge this when that is implemented there.

nchamo:
> This will create a hard dependency with having to use the health check on explorer when selecting the realm. Have we done that already?
> 
> I'd suggest we only merge this when that is implemented there.

@jmoguilevsky is close to finishing it. I agree, this won't be merged until that change is merged on the explorer

pablitar (approved):
Nice work!

PR #498, opened 2021-05-31
fix: use correct block when validating wearables

nchamo:
Approval for wearables can change over time. So if a wearable is approved when first deployed, it might not be approved a few months later. 

The problem with the current validation approach is that it only checks the current approval values on the blockchain. So it could happen that a new catalyst that is trying to sync will find some deployments invalid now, while they where valid when deployed. 

So the idea is to check the approval values of the ethereum block when the wearable was deployed.

Now, there is one extra situation to consider. It could happen that a user loses access but manages to deploy an update before the subgraph is indexed. That means that the deployment will be considered valid for the current catalyst, but invalid for other catalysts that might sync later. To fix this, we will consider a deployment valid if the user had access at the time of the deployment, or five minutes before.


content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -59,33 +71,61 @@ export class AccessCheckerForWearables {
...
+        timestamp
+      )
+      // It could happen that the subgraph hasn't synced yet, so someone who just lost access still managed to make a deployment. The problem would be that when other catalysts perform
+      // the same check, the subgraph might have synced and the deployment is no longer valid. So, in order to prevent inconsistencies between catalysts, we will allow all deployments that
+      // have access now, or had access 5 minutes ago.
       return (
-        (permissions.collectionCreator && permissions.collectionCreator === ethAddressLowercase) ||
-        (permissions.collectionManagers && permissions.collectionManagers.includes(ethAddressLowercase)) ||
-        (permissions.itemManagers && permissions.itemManagers.includes(ethAddressLowercase))
+        (await this.hasPermission(ethAddress, collectionsSubgraphUrl, collection, itemId, blockNumberNow)) ||
```
agusaldasoro, line 90: is this `blockNumberNow` at current time or at deployment time?

agusaldasoro (approved):
Great!

content/src/service/access/AccessCheckerForWearables.ts
nchamo: Good point! It's at the deployment's time, but it definitely isn't clear by its name

Issue #500, opened 2021-06-01
Add archipelago library & bind heartbeat to setting position in archipelago


pablitar:

nchamo:
Done in https://github.com/decentraland/catalyst/pull/523

PR #523, opened 2021-06-02
feat: add archipelago library to lighthouse

nchamo:
We are simply adding the archipelago library to the lighthouse. We are also updating the position with each heartbeat

Closes https://github.com/decentraland/catalyst/issues/500


pablitar (approved):
LGTM! Great work!

PR #525, opened 2021-06-04
chore: Extract authentication logic to other file

### LINKED ISSUE #507: Move authentication logic to its own file
Authentication logic for peers is currently in server.ts. We should move it to its own file / service / abstraction

pablitar:
Closes #507

### DIFF
--- a/comms/peer/peer.package.json
+++ b/comms/peer/peer.package.json
@@ -1,6 +1,6 @@
 {
-  "name": "decentraland-katalyst-peer",
-  "description": "Katalyst peer library",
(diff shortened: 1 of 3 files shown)


comms/peer/peer.package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "decentraland-katalyst-peer",
-  "description": "Katalyst peer library",
+  "name": "@dcl/catalyst-peer",
```
nchamo, line 2: Yeees. Much better!!! ❤️

nchamo (approved):
Great work!

PR #526, opened 2021-06-05
feat: add support for global config

nchamo:
**tl;dr;**
* We are adding a new source of configuration. This new source is an endpoint that is polled every 15 min
* We are adding a new endpoint `/config` to read server's the current configuration
* We are adding listeners to the `configService`, so other services can react to the change in config. By doing this, we are avoiding the need for whole server restarts

### What are we doing here?
We are now giving a lot more love to the `configService`. There will be 4 different sources for configuration, with the following precedence:
```
Environment config >> storage config >> global config >> defaults
```

The main change is that now, there is a global configuration. This configuration can be controlled by the foundation, but it can be easily overridden by catalyst owners. This approach will make it easier to test different configuration 
without having to connect to the server and restart it.

We are also created the concept of `LighthouseConfig`, so we avoid strings that can lead to problems with typos

_Note: this change was tested in .zone, by changing the global config and making sure that the server updates its config correctly._

### DIFF
--- a/comms/lighthouse/BUILD.bazel
+++ b/comms/lighthouse/BUILD.bazel
@@ -69,6 +69,7 @@ ts_library(
         "@npm//@types",
         "@npm//cors",
         "@npm//dcl-crypto",
+        "@npm//dcl-catalyst-commons",
         "@npm//express",
         "@npm//fast-deep-equal",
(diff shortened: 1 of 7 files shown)


pablitar (approved):
LGTM! Great work!

PR #528, opened 2021-06-07
chore: Refactor to add websocket abstraction (Part 1)

pablitar:
Still not added, but wanted to merge this to master before keep going


nchamo (approved):
Best PR ever ❤️

comms/lighthouse/src/peers/initPeerJsServer.ts
```diff
@@ -0,0 +1,73 @@
...
+  peerServer.on('disconnect', (client: any) => {
+    console.log('User disconnected from server socket. Removing from all rooms & layers: ' + client.id)
+    layersServiceGetter().removePeer(client.id)
+    archipelagoServiceGetter().clearPeer(client.id)
+  })
+
+  peerServer.on('error', console.log)
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  //@ts-ignore
```
menduz, line 48: What is the purpose of that `//@ts-ignore`?

pablitar: LET US WORK @menduz.

This was here before and will be removed IF WE CAN

pablitar: It is because PeerJSServer types don't correspond to the actual event that you receive.

This was actually here before. We just moved it around. Eventually, we'll try to get rid of the ts-ignore, at least trying to replace them for any or something not so drastic.

PR #530, opened 2021-06-08
feat: send config updates to archipelago controller

nchamo:
We are doing two things in this PR:
1. We updated the `@dcl/archipelago` library
2. We now listen to events in the configService and send those updates to the archipelago controller


pablitar (approved):
LGTM! :D

PR #534, opened 2021-06-08
chore: Added some abstraction & typing to peer messages.

### LINKED ISSUE #506: Encapsulate websocket communication with peers in a service
Web socket communications with peer are currently handled in server.ts file. There is no abstraction.

We should add an abstraction to encapsulate web socket communications with peers, likely a service.

pablitar:
Closes #506 
- Added a messages handler
- Added messages type structure
- Added method in peer service to send arbitrary messages to peers
- Also fixed the test app so we can start using it also for testing archipelago :D


nchamo (approved):
Great work!

PR #537, opened 2021-06-09
feat: Forwarding island updates to peers

pablitar:
Currently not much processing is done. Each change is forwarded to those that are interested in it, in order


comms/lighthouse/src/peers/peersService.ts
```diff
@@ -143,4 +144,38 @@ export class PeersService implements IPeersService {
         .slice(0, 100)
     )
   }
+
+  sendChangeToIsland(
```
nchamo, line 148: `sendIslandChange` and `sendChangeToIsland` are very similar in name. Could we make them a little bit more different. Maybe `informIslandAboutUpdate` and `sendNewIslandToPeer`?

comms/lighthouse/src/peers/messageTypes.ts
```diff
@@ -14,10 +47,6 @@ export type HeartbeatMessage = {
     connectedPeerIds: string[]
     parcel?: [number, number]
     position?: Position
-    // These fields will be removed soon
-    optimizeNetwork?: boolean
```
nchamo, line 18: Good catch!

comms/lighthouse/src/peers/archipelagoService.ts
```diff
@@ -27,7 +35,49 @@ export class ArchipelagoService {
...
+      }
+    }
+
+    for (const id in updates) {
+      const update = updates[id]
+
+      const island = await getIsland(updates[id].islandId)
+      // This could be undefined for a short lived island, in the round trip between the worker & this service.
+      if (island) {
+        switch (update.action) {
```
nchamo, line 66: I'm just thinking out loud here, and it can definitely be left for another PR,  but would it make sense to group updates by island?

Because I'm seeing that we might be sending a lot of messages to the same peer to let them know about the updates on their island. And we could just send a message that is something like
```ts
export type ChangesInIsland = {
  type: PeerOutgoingMessageType.CHANGES_IN_ISLAND
  payload: {
    left: string[]
    joined: string[]
  }
}
```

nchamo (commented):
Great work!

Added a comment and I'd like to see what you think about it

nchamo (approved):

comms/lighthouse/src/peers/archipelagoService.ts
pablitar: Yeah, I thought about this too... But in the end I decided to go for the simple implementation because I'm not really sure it'll be better.

Intuitively, it seems better (less messages over the network, right?). But I'm not sure we'll have much overhead, it being a websocket, and I'm also not sure the added latency will be worth the gains that we will have. It may also be better to forward the message directly to the socket, since it is native code and it should have better performance.

I think we should go with this simple implementation and once we have it working we can add performance tests and optimize.

PR #540, opened 2021-06-10
fix: expose errors as json

nchamo:
Before this change, we would expose deployments errors as a plain text. The problem is that by using `res.send`, express would set the content type as `text/html`. Then some clients would fail to render the error message correctly, because they failed to interpret the error message as a valid html.

So in an effort to be consistent, we will return the error message as a json (as we do with successful deployments)


Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN yarn install &&\
 WORKDIR /app
 
 ENV COMMIT_HASH=bc34832282cfa746cfb1f27184cf3b53f321a164
-ENV CATALYST_VERSION=1.1.4
+ENV CATALYST_VERSION=1.1.5
```
nchamo, line 30: We will release a new version with this fix

menduz (approved):

PR #542, opened 2021-06-14
feat: avoid the need to name the entity file

nchamo:
Before this change, the entity file had to be named `entity.json`. We required this naming so that we could find it easily. The thing is, that we were already hashing the files, so we could just find it by its hash. 

By doing this change, we can now forget about names when deploying and just send the files without any other considerations.

### DIFF
--- a/content/src/service/Service.ts
+++ b/content/src/service/Service.ts
@@ -27,20 +27,20 @@ import { FailedDeployment, FailureReason } from './errors/FailedDeploymentsManag
 export interface MetaverseContentService {
   start(): Promise<void>
   deployEntity(
-    files: ContentFile[],
+    files: DeploymentFiles,
     entityId: EntityId,
     auditInfo: LocalDeploymentAuditInfo,
     origin: string,
     task?: Database
   ): Promise<DeploymentResult>
   deployLocalLegacy(
-    files: ContentFile[],
+    files: DeploymentFiles,
     entityId: EntityId,
     auditInfo: LocalDeploymentAuditInfo,
     task?: Database
   ): Promise<DeploymentResult>
   deployToFix(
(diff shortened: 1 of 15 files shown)


content/src/service/Service.ts
```diff
@@ -27,20 +27,20 @@ import { FailedDeployment, FailureReason } from './errors/FailedDeploymentsManag
 export interface MetaverseContentService {
   start(): Promise<void>
   deployEntity(
-    files: ContentFile[],
+    files: DeploymentFiles,
```
nchamo, line 30: We are now allowing to send the already hashed files, so that we avoid hashing them both at the denylist and service layer

content/src/service/ServiceImpl.ts
```diff
@@ -384,16 +384,15 @@ export class ServiceImpl implements MetaverseContentService, ClusterDeploymentsS
...
 
-  static findEntityFile(files: ContentFile[]): ContentFile {
-    const filesWithName = files.filter((file) => file.name === ENTITY_FILE_NAME)
-    if (filesWithName.length === 0) {
-      throw new Error(`Failed to find the entity file. Please make sure that it is named '${ENTITY_FILE_NAME}'.`)
-    } else if (filesWithName.length > 1) {
-      throw new Error(
-        `Found more than one file called '${ENTITY_FILE_NAME}'. Please make sure you upload only one with that name.`
+  static async hashFiles(files: DeploymentFiles): Promise<Map<ContentFileHash, ContentFile>> {
+    if (files instanceof Map) {
```
agusaldasoro, line 388: Can you add a comment here explaining why is this case? thanks

agusaldasoro (approved):
Great work!

PR #549, opened 2021-06-17
feat: support content hash for wearable validations

nchamo:
We are now changing the way we validate wearables. If the entity's id matches the `contentHash` property in the subgraph, we will allow the deployment, regardless of who made it. 

Also, we split the unit tests for the AccessChecker, and added a lot of them


content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -115,12 +136,17 @@ export class AccessCheckerForWearables {
...
     const ethAddressLowercase = ethAddress.toLowerCase()
-    return (
+    const addressHasAccess =
       (permissions.collectionCreator && permissions.collectionCreator === ethAddressLowercase) ||
       (permissions.collectionManagers && permissions.collectionManagers.includes(ethAddressLowercase)) ||
       (permissions.itemManagers && permissions.itemManagers.includes(ethAddressLowercase))
-    )
+
+    // Deployments to the content server are made after the collection is completed, so that the committee can then approve it
+    const isCollectionValid = !permissions.isApproved && permissions.isCompleted
```
agusaldasoro, line 147: is it okay to negate `!permissions.isApproved`? If so, can you make it more clear in the comment, please?

agusaldasoro (approved):

content/src/service/access/AccessCheckerForWearables.ts
nchamo: Just made the comment clearer

PR #555, opened 2021-06-22
feat: Add transmission filter for islands

pablitar:
Also, made it so peerjs-server tests run (they were not running)


comms/lighthouse/src/peerjs-server/instance.ts
```diff
@@ -40,24 +40,24 @@ export const createInstance = ({
     config
   })
 
-  wss.on('connection', (client: IClient) => {
+  wss.on('connection', async (client: IClient) => {
```
menduz, line 43: careful, wss will ignore the Promise returned by this function. Failing silently or killing the program, or leading to leaks.

comms/lighthouse/src/peerjs-server/instance.ts
```diff
@@ -40,24 +40,24 @@ export const createInstance = ({
...
+        await messageHandler.handle(client, message)
       }
       realm.clearMessageQueue(client.getId())
     }
 
     app.emit('connection', client)
   })
 
-  wss.on('message', (client: IClient, message: IMessage) => {
+  wss.on('message', async (client: IClient, message: IMessage) => {
```
menduz, line 58: Same here, wss won't handle any error of the async function, nor wait for it in any scenario

comms/lighthouse/src/peerjs-server/messageHandler/handlersRegistry.ts
```diff
@@ -17,7 +17,7 @@ export class HandlersRegistry implements IHandlersRegistry {
     this.handlers.set(messageType, handler)
   }
 
-  public handle(client: IClient | undefined, message: IMessage): boolean {
+  public handle(client: IClient | undefined, message: IMessage): boolean | Promise<boolean> {
```
menduz, line 20: Is there a reason to keep `bool | Promise<bool>` instead of simply `Promise<bool>`? in the later we can take advantage of linters to warn us about

```ts
const a: bool | Promise<bool> = fn()

// not handled by linter, may always return "true" because a is a Promise
if(a) {...}
```

comms/lighthouse/src/peerjs-server/services/messagesExpire/index.ts
```diff
@@ -37,8 +37,12 @@ export class MessagesExpire implements IMessagesExpire {
     }
 
     // Clean up outstanding messages
-    this.timeoutId = setTimeout(() => {
-      this.pruneOutstanding()
+    this.timeoutId = setTimeout(async () => {
```
menduz, line 40: Same as first comment, setTimeout should have a function returning void, it won't handle the catch from the Promise

I'm thinking that a high order function to wrap "async callbacks" and returns void functions may be a good idea.

comms/lighthouse/src/peerjs-server/instance.ts
pablitar: Nice catch. I'll see what's the best way to handle this. I'm thinking since WSS already has an `error callback, maybe I should try-catch and emit `error`?

comms/lighthouse/src/peerjs-server/messageHandler/handlersRegistry.ts
pablitar: Hmm, the only reason I made it so it could return boolean | Promise was because it seemed like it would limit the ripple effect of the change. But it is true that it won't be so hard to go full promises, so I'll do that.

comms/lighthouse/src/peerjs-server/services/messagesExpire/index.ts
pablitar: This one is try-catched, so it shouldn't be a problem. But maybe I should propagate the error somehow? I think it is not necessary, though.

menduz: Ok sounds good, I'd do something to start getting used to it, this is something that would break the program in a future. But not this specific case, due to the "catch"

```
Received has value: undefined (node:32535) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:32535) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

comms/lighthouse/src/peerjs-server/instance.ts
menduz: I have many ideas, I don't know what are the implications of the 'error' event. For instance, in web sockets, whenever you receive an 'error' event, you can assume the socket is closed. 

The one I like the most is creating a HighOrderFunction that forces us to consider corner cases, like this:

```ts
function asyncCallback(fn: (args) => Promise<any>): (args) => void {
  return (...args) => 
    fn.apply(null, args).catch(err => {
      console.log('error in async callback, the program will finish')
      console.error(err)
      process.exit(1)
    })
}

  wss.on('connection', asyncCallback(async (client: IClient) => {
    try {
    ..
    } catch(e) {
      // explicitly handle every error for each case
    }
  })
```

The idea is to force us to program for corner cases, because the program will exit if we don't

pablitar: Hmmm, but I'm not really sure we want to be so forceful. This code (peerjs-server) has been running for a while with no issues, and I wouldn't want to make it crash on every error that we don't explicitly handle.

It is true that we are making modifications of peerjs server, and that could introduce new issues that maybe we should handle. But I'd rather wait until modifications are more significant before introducing a change in behaviour so significant. I prefer to limit the changes if possible.

menduz: Very well, please be mindful to properly handle the "catch" at least

comms/lighthouse/src/peers/initPeerJsServer.ts
```diff
@@ -25,7 +25,8 @@ export function initPeerJsServer({
   const options: Partial<IConfig> = {
     path: '/',
     idGenerator: () => idService.nextId(),
-    authHandler: peerAuthHandler({ noAuth, peersServiceGetter: peersService, ethNetwork })
+    authHandler: peerAuthHandler({ noAuth, peersServiceGetter: peersService, ethNetwork }),
+    transmissionFilter: async (src, dst) => await archipelagoService().areInSameIsland(src, dst)
```
nchamo, line 29: Super super nit:
```suggestion
    transmissionFilter: (src, dst) => archipelagoService().areInSameIsland(src, dst)
```

nchamo (approved):
Amazing work Pablo!!!! Truly amazing work 👏

package.json
```diff
@@ -49,7 +49,7 @@
     "karma-requirejs": "^1.1.0",
     "karma-sourcemap-loader": "^0.3.8",
     "karma-spec-reporter": "^0.0.32",
-    "mock-socket": "^9.0.3",
+    "mock-socket": "8.0.5",
```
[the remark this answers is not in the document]
pablitar: Yep. Seems the usage that the existing peerjs tests were doing of mock sock was not compatible with the new version (peerjs used 8.0.5)

comms/lighthouse/test/peerjs-server/services/webSocketServer/index.spec.ts
```diff
@@ -171,10 +173,10 @@ describe('WebSocketServer', () => {
       return errorSent
     }
 
-    expect(await getError(fakeURL)).toBeTrue()
-    expect(await getError(`${fakeURL}?key=${config.key}`)).toBeTrue()
-    expect(await getError(`${fakeURL}?key=${config.key}&id=1`)).toBeTrue()
-    expect(await getError(`${fakeURL}?key=notValidKey&id=userId&token=userToken`, Errors.INVALID_KEY)).toBeTrue()
+    expect(await getError(fakeURL)).toBe(true)
```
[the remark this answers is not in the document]
pablitar: Not really. But for some reason toBeTrue would explode, and I didn't want to deal with that ATM: https://github.com/JamieMason/Jasmine-Matchers/issues/104

PR #557, opened 2021-06-23
fix: add fields to next link

nchamo:
We realized that the the `next` field in the `/deployments` endpoint did not include the query param `fields`, even when it was included in the original request. We are now fixing that

### DIFF
--- a/commons/servers/QueryParameters.ts
+++ b/commons/servers/QueryParameters.ts
@@ -12,10 +12,11 @@ function convertFiltersToQueryParams(filters?: Record<string, any>): Map<string,
   const entries = Object.entries(filters)
     .filter(([_, value]) => !!value)
     .map<[string, string[]]>(([name, value]) => {
-      const newName = name.endsWith('s') ? name.slice(0, -1) : name
+      let newName = name
       let newValues: string[]
       // Force coercion of number, boolean, or string into string
       if (Array.isArray(value)) {
+        newName = name.endsWith('s') ? name.slice(0, -1) : newName
         newValues = [...value].filter(isValidQueryParamValue).map((_) => `${_}`)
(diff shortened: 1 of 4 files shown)


commons/servers/QueryParameters.ts
```diff
@@ -12,10 +12,11 @@ function convertFiltersToQueryParams(filters?: Record<string, any>): Map<string,
   const entries = Object.entries(filters)
     .filter(([_, value]) => !!value)
     .map<[string, string[]]>(([name, value]) => {
-      const newName = name.endsWith('s') ? name.slice(0, -1) : name
+      let newName = name
```
nchamo, line 15: Before this change, the standard was that all variables that ended with `s` had the `s` removed. This was because we wanted to convert `entityTypes: [ type1, type2 ]` to `&entityType=type1&entityType=type2`.

But `fields` is as string, so are now only making that "s modification" when the value is an array

content/src/controller/Controller.ts
```diff
@@ -525,8 +527,12 @@ export class Controller {
         nextFilters.to = lastDeployment.entityTimestamp
       }
     }
+
+    const fields = !options.fields || options.fields === DEFAULT_FIELDS_ON_DEPLOYMENTS ? '' : options.fields.join(',')
+
     const nextQueryParams = toQueryParams({
       ...nextFilters,
+      fields,
```
agusaldasoro, line 535: Is there a way to be sure that in the future we don't forget to add anything?

agusaldasoro (approved):
Great work!

content/src/controller/Controller.ts
nchamo: Not that I can think of right now

PR #560, opened 2021-06-24
feat: Exposing current clients count

pablitar:
Also, added an async handler for async routes (otherwise, errors would be ignored)


nchamo (approved):
Great catch!

PR #562, opened 2021-06-28
fix: make endpoint casing consistent

nchamo:
For some reason, `/failedDeployments` and `/pointerChanges` were not consistent in casing with the other endpoints. We are taking a first step towards changing that


agusaldasoro (approved):
👏

PR #565, opened 2021-06-28
chore: update catalyst client

nchamo:
We are updating catalyst client so that the changes in https://github.com/decentraland/catalyst-client/pull/93 are used during sync


lambdas/src/utils/SmartContentClient.ts
```diff
@@ -99,6 +102,18 @@ export class SmartContentClient implements ContentAPI {
     throw new Error('New deployments are currently not supported')
   }
 
+  buildEntity({ type, pointers, files, metadata }: BuildEntityOptions): Promise<DeploymentPreparationData> {
```
agusaldasoro, line 105: I'm not sure what those no implemented methods are

agusaldasoro (approved):

lambdas/src/utils/SmartContentClient.ts
nchamo: It's this 
https://github.com/decentraland/catalyst-client/pull/90

PR #569, opened 2021-07-06
feat: Archipelago Support

pablitar:
Long running branch to add archipelago support.

This can be merged now. But once done, it may be hard going back if we have to :scream:


package.json
```diff
@@ -93,8 +93,8 @@
     "migrate": "node-pg-migrate -m content/src/migrations/scripts -j ts"
   },
   "dependencies": {
-    "@dcl/archipelago": "^1.0.0-20210609183931.commit-668cf2c",
-    "@dcl/catalyst-peer": "^1.0.0-20210706172727.commit-a9877de",
+    "@dcl/archipelago": "^1.0.0-20210713173135.commit-dbc0d1c",
+    "@dcl/catalyst-peer": "^1.0.0-20210713201511.commit-523f1ac",
```
menduz, line 97: When to use the snapshot vs when to use "releases" is a question for the near future.

pablitar: True. I think both of these can be released at the moment. So I'll replace them with a release in the sake of cleanliness :)

comms/README.md
```diff
@@ -1,47 +1,25 @@
...
-
-The Lighthouse is an server that tracks peers in different "layers" and "rooms" in order to enable them to make the necessary connections. It also handles authentication.
 
-The following docs show how to run a local instance of the Lighthouse, and how to run a simple react app that serves to test the local lighthouse.
-
-## Set up
+Comms is comprised of two distinct modules: One for Catalyst server called "Lighthouse" and other for clients & P2P network called "Peer Library".
 
-* Install libs
+The Lighthouse is an server that tracks peers in different positions in order to enable them to make the necessary connections. It also handles authentication.
```
agusaldasoro, line 5:
```suggestion
The Lighthouse is a server that tracks peers in different positions in order to enable them to make the necessary connections. It also handles authentication.
```

tsconfig.json
```diff
@@ -31,8 +31,7 @@
       "@katalyst/test-helpers/*": ["content/test/helpers/*"],
       "decentraland-katalyst-contracts/*": ["contracts/*"],
       "decentraland-katalyst-commons/*": ["commons/servers/*"],
-      "decentraland-katalyst-utils/*": ["commons/utils/*"],
-      "katalyst/comms/peer/proto/*": ["dist/bin/comms/peer/proto/*"]
+      "decentraland-catalyst-utils/*": ["commons/utils/*"]
```
agusaldasoro, line 34: If we are changing from katalyst to catalyst, shouldn't we do it for all?

agusaldasoro (approved):

tsconfig.json
pablitar: Yeah... I remember starting that change and reverting it for some reason. Though I cannot remember now why hahaha. I assume it was more difficult that it seemed.

PR #570, opened 2021-07-06
Use pow in catalyst-client

agusaldasoro:
This PR includes the necessary changes to implement Proof of Work when fetching another catalysts by using latest version of Catalyst Client and Catalyst Commons which supports Cookie handling for JWT.


content/src/Environment.ts
```diff
@@ -296,7 +297,11 @@ export class EnvironmentBuilder {
           ? DEFAULT_BLOCKS_SUBGRAPH_MATIC_MAINNET
           : DEFAULT_BLOCKS_SUBGRAPH_MATIC_MUMBAI)
     )
-
+    this.registerConfigIfNotAlreadySet(
+      env,
+      EnvironmentConfig.PROOF_OF_WORK,
+      () => process.env.PROOF_OF_WORK == 'enabled'
```
nchamo, line 303: We usually use `true` or `false` with boolean variables
```suggestion
      () => process.env.PROOF_OF_WORK === 'true'
```

content/test/integration/TestServer.ts
```diff
@@ -40,7 +40,12 @@ export class TestServer extends Server {
     super(env)
     this.serverPort = env.getConfig(EnvironmentConfig.SERVER_PORT)
     this.storageFolder = env.getConfig(EnvironmentConfig.STORAGE_ROOT_FOLDER)
-    this.client = new ContentClient(this.getAddress(), '', env.getBean(Bean.FETCHER))
+    this.client = new ContentClient({
+      contentUrl: this.getAddress(),
+      origin: '',
+      proofOfWorkEnabled: true,
```
nchamo, line 46: Let's set this to false, since we don't need it in tests

lambdas/src/Environment.ts
```diff
@@ -91,6 +91,7 @@ export const enum EnvironmentConfig {
   PROFILE_NAMES_CACHE_TIMEOUT,
   PROFILE_WEARABLES_CACHE_MAX,
   PROFILE_WEARABLES_CACHE_TIMEOUT,
+  PROOF_OF_WORK,
```
nchamo, line 94: I don't think this is being configured at all

lambdas/src/utils/SmartContentClient.ts
```diff
@@ -142,7 +141,13 @@ export class SmartContentClient implements ContentAPI {
       } catch {
         SmartContentClient.LOGGER.info('Defaulting to external content server url: ', contentClientUrl)
       }
-      this.contentClient.resolve(new ContentClient(contentClientUrl, 'lambdas'))
+      this.contentClient.resolve(
+        new ContentClient({
+          contentUrl: contentClientUrl,
+          origin: 'lambdas',
+          proofOfWorkEnabled: this.proofOfWorkEnabled
```
nchamo, line 148: If the content url is the internal one, then we shouldn't enable PoW. Since we are not going through nginx at all
```suggestion
          proofOfWorkEnabled: contentClientUrl === this.externalContentServerUrl && this.proofOfWorkEnabled
```

content/src/helpers/FetcherFactory.ts
```diff
@@ -13,4 +13,8 @@ export class FetcherFactory {
       }
     })
   }
+
+  static copy(originalFetcher: Fetcher): Fetcher {
+    return new Fetcher(originalFetcher['customDefaults'])
```
nchamo, line 18: Does this work? hahah

Wouldn't it be better if `Fetcher` has a `.clone()` or exposed the default config?

nchamo (commented):
Some comments

content/src/helpers/FetcherFactory.ts
agusaldasoro: It works, any way you are right 😂

content/src/Environment.ts
```diff
@@ -296,7 +297,7 @@ export class EnvironmentBuilder {
           ? DEFAULT_BLOCKS_SUBGRAPH_MATIC_MAINNET
           : DEFAULT_BLOCKS_SUBGRAPH_MATIC_MUMBAI)
     )
-
+    this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.PROOF_OF_WORK, () => process.env.PROOF_OF_WORK == 'true')
```
nchamo, line 300:
```suggestion
    this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.PROOF_OF_WORK, () => process.env.PROOF_OF_WORK === 'true')
```

lambdas/src/Environment.ts
```diff
@@ -200,6 +201,11 @@ export class EnvironmentBuilder {
     this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.PROFILE_WEARABLES_CACHE_TIMEOUT, () =>
       ms(process.env.PROFILE_WEARABLES_CACHE_TIMEOUT ?? '30m')
     )
+    this.registerConfigIfNotAlreadySet(
+      env,
+      EnvironmentConfig.PROFILE_WEARABLES_CACHE_TIMEOUT,
```
nchamo, line 206: What? I think this is a copy-paste error. You are not configuring PROOF_OF_WORK

nchamo (approved):
Great work!

PR #579, opened 2021-07-15
refactor: remove unused code

nchamo:
We are now deleting:
* The SQS reporter
* The Segment.io reporter
* Manual tests (these didn't even compile)

None of this code was being used, so we will just delete it


agusaldasoro (approved):
👏 👏 👏 👏

menduz (approved):

PR #580, opened 2021-07-15
migration: remove invalid deployments

nchamo:
There are some deployments that were made to the catalysts that were invalid. thegraph wasn't working correctly, and that's why they were considerer valid at the time, but they shouldn't have. Since then, these deployments have been overwritten with valid deployments, so there should be no impact in deleting them.

It is important to mention that catalysts that start from scratch will consider these deployments invalid and avoid deploying them locally. Since we want all catalysts to be in sync, we will remove them from all current catalysts.

This is the list of all 7 invalid deployments:
```
urn:decentraland:matic:collections-v2:0x9cdd5e329b4aa5c1cec3a0d99fe56f9f8f7892b5:0
    QmX97nxQMXfPyU2NWdJQ5n14wcvaEFRVrcbVPgLua4YqFj (oldest)
    QmPR9vrkjSAXDoVAaR5n7WHjQqQZssvCfgwtJoc19T8bPi
    QmXwahFWd6kcKWzuZU9ZCgY7W2xCDwygM9AbatPKMBDjp4
    QmQfNCQYgNxg4GURQUB1LUKEDcztYTUa9btRazy5ggBRMf (newest)
urn:decentraland:matic:collections-v2:0xf6f601efee04e74cecac02c8c5bdc8cc0fc1c721:0
    QmeiCGFNg3xbqEUpUnxmcaYkgk6mmW9qkq3dRwwHeVaER2
urn:decentraland:matic:collections-v2:0x321cd6d0901ede3b10e042febcc96e4e84202c89:0
    QmUgjH2fmZ43nuQAUcJrtG7uLEThjH8Spe8YCw7Ff5rcHs
urn:decentraland:matic:collections-v2:0xf60ad515b3dd3a4ad232011697b25dfc1535ecbe:0
    QmU9JciPcj8mo3p86r5i8iEQzEkj3h7BMcpE1JvfCqS9Um
```

Since it is extremely hard to re-write history, the idea will be to:
* Move all deployments on the affected pointers (including valid ones) to failed-deployments
* Delete the deployments from other the database tables
* Re-deploy them

By doing this, we will end up using the usual mechanism to write pointer history, and invalid deployments will now fail to deploy.

_Note: this migration + manual fix has been tested successfully on peer-testing_


content/src/repository/extensions/LastDeployedPointersRepository.ts
```diff
@@ -5,8 +5,8 @@ import { DeploymentId } from './DeploymentsRepository'
 export class LastDeployedPointersRepository {
   constructor(private readonly db: Database) {}
 
-  /** Returns the last deployments (could be active or not) on the given pointers */
-  getLastDeploymentsOnPointers(
+  /** Returns the last deployments that were active on the given pointers (could be active or not right now) */
+  getLastActiveDeploymentsOnPointers(
```
nchamo, line 9: Some renaming so that it is clearer for everyone

content/src/service/pointers/PointerManager.ts
```diff
@@ -17,8 +17,11 @@ export class PointerManager {
     deploymentId: DeploymentId,
     entity: Entity
   ): Promise<DeploymentResult> {
-    // Fetch last deployments on pointers
-    const lastDeployments = await lastDeployedPointersRepo.getLastDeploymentsOnPointers(entity.type, entity.pointers)
+    // Fetch active last deployments on pointers
```
nchamo, line 20: Some renaming so that it is clearer for everyone

agusaldasoro (approved):

PR #584, opened 2021-07-20
refactor: remove origin

nchamo:
In https://github.com/decentraland/catalyst/pull/579 we removed the Segment.io reporter since it wasn't being used anymore. With that change, we can now get read of the origin that we've been passing around the code everywhere.

This "origin" property was only used in that reporter, so it doesn't make much sense to keep it around


agusaldasoro (approved):
👏

PR #594, opened 2021-07-23
feat: add support for ipfs hashing

nchamo:
We are now adding support for v4 entities. This new version of entities will have IPFS v2 hashes instead of the current ones.

As part of this change we are also:
* Making snapshots use IPFS hashes
* Supporting wearable reverts with IPFS hashes

### DIFF
--- a/content/src/service/access/AccessCheckerForWearables.ts
+++ b/content/src/service/access/AccessCheckerForWearables.ts
@@ -118,7 +118,7 @@ export class AccessCheckerForWearables {
 
       if (!!permissions.contentHash) {
         const deployedByCommittee = permissions.committee.includes(ethAddressLowercase)
-        const calculateHash = () => {
+        const calculateHashes = () => {
           // Compare both by key and hash
           const compare = (a: { key: string; hash: string }, b: { key: string; hash: string }) => {
             if (a.hash > b.hash) return 1
@@ -128,9 +128,9 @@ export class AccessCheckerForWearables {
           const entries = Array.from(content?.entries() ?? [])
           const contentAsJson = entries.map(([key, hash]) => ({ key, hash })).sort(compare)
           const buffer = Buffer.from(JSON.stringify({ content: contentAsJson, metadata }))
-          return Hashing.calculateBufferHash(buffer)
+          return Promise.all([Hashing.calculateBufferHash(buffer), Hashing.calculateIPFSHash(buffer)])
         }
-        return deployedByCommittee && (await calculateHash()) === permissions.contentHash
+        return deployedByCommittee && (await calculateHashes()).includes(permissions.contentHash)
       } else {
         const addressHasAccess =
           (permissions.collectionCreator && permissions.collectionCreator === ethAddressLowercase) ||
--- a/content/src/service/snapshots/SnapshotManager.ts
+++ b/content/src/service/snapshots/SnapshotManager.ts
@@ -78,7 +78,7 @@ export class SnapshotManager {
           const buffer = Buffer.from(JSON.stringify(inArrayFormat))
 
           // Calculate the snapshot's hash
-          const hash = await Hashing.calculateBufferHash(buffer)
+          const hash = await Hashing.calculateIPFSHash(buffer)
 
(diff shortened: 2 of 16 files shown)


content/test/integration/E2ETestUtils.ts
```diff
@@ -118,7 +118,7 @@ export async function deployEntitiesCombo(
   for (const { deployData } of entitiesCombo) {
     deploymentResult = await service.deployEntity(Array.from(deployData.files.values()), deployData.entityId, {
       authChain: deployData.authChain,
-      version: EntityVersion.V2
+      version: EntityVersion.V3
```
nchamo, line 121: Should have been V3 all along

content/test/integration/service/deployments/deployment-filters.spec.ts
```diff
@@ -122,7 +122,7 @@ describe('Integration - Deployment Filters', () => {
   async function deployWithAuditInfo(entities: EntityCombo[], overrideAuditInfo?: Partial<AuditInfo>) {
     const result: Timestamp[] = []
     for (const { deployData } of entities) {
-      const newAuditInfo = { version: EntityVersion.V2, authChain: deployData.authChain, ...overrideAuditInfo }
+      const newAuditInfo = { version: EntityVersion.V3, authChain: deployData.authChain, ...overrideAuditInfo }
```
nchamo, line 125: Should have been V3 all along

menduz (approved):
Awesome, thanks for adding the TODO comments to make next steps clearer

content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -128,9 +128,9 @@ export class AccessCheckerForWearables {
           const entries = Array.from(content?.entries() ?? [])
           const contentAsJson = entries.map(([key, hash]) => ({ key, hash })).sort(compare)
           const buffer = Buffer.from(JSON.stringify({ content: contentAsJson, metadata }))
-          return Hashing.calculateBufferHash(buffer)
+          return Promise.all([Hashing.calculateBufferHash(buffer), Hashing.calculateIPFSHash(buffer)])
```
agusaldasoro, line 131: Don't you have the entity version here so you can only execute the necessary algorithm?

content/src/service/snapshots/SnapshotManager.ts
```diff
@@ -78,7 +78,7 @@ export class SnapshotManager {
           const buffer = Buffer.from(JSON.stringify(inArrayFormat))
 
           // Calculate the snapshot's hash
-          const hash = await Hashing.calculateBufferHash(buffer)
+          const hash = await Hashing.calculateIPFSHash(buffer)
```
agusaldasoro, line 81: Why are you sure that all the entities are v4?

content/test/unit/service/validations/Validations.spec.ts
```diff
@@ -348,16 +348,56 @@ describe('Validations', function () {
       await assertNoErrors(result)
     })
   })
+
+  describe('IFPS hashing', () => {
+    it(`when an entity's id is not an ipfs hash, then it fails`, async () => {
```
agusaldasoro, line 353: Should we test also the older hashing algorithm so it fails?

content/src/service/access/AccessCheckerForWearables.ts
nchamo: But this hash is stored on the blockchain by the committee. I'm not sure which version they'll use, and it won't depend on the entity version. That's why we are checking for both

content/src/service/snapshots/SnapshotManager.ts
nchamo: This has nothing to do with the entities themselves. The snapshot file will be exposed with the new hash, that's it. It doesn't matter if the entities are v3 or v4

content/test/unit/service/validations/Validations.spec.ts
nchamo: Done!

agusaldasoro (approved):

PR #596, opened 2021-07-27
Configure cors on servers

agusaldasoro:
1. Configure cors to send the origin when received in the request.
2. Expose OPTIONS method on every path to support cors with credentials.


comms/lighthouse/src/server.ts
```diff
@@ -72,6 +91,8 @@ const CURRENT_ETH_NETWORK = process.env.ETH_NETWORK ?? DEFAULT_ETH_NETWORK
 
   const idService = new IdService({ alphabet: idAlphabet, idLength })
 
+  app.options('*', cors(corsOptions))
```
menduz, line 94:
```suggestion
  app.use(cors(corsOptions))
```

Cors works at many levels, for all HTTP methods it should inject headers if necessary

agusaldasoro: What it's doing here it's handling the requests with `OPTIONS` method, I configured it for all paths. Anyway, I can change it to only respond on known routes if you consider it better.

menduz: Cors headers must be present in responses too, not only in the OPTIONS method. Please use `app.use` instead of `app.options`

lambdas/src/Server.ts
```diff
@@ -41,7 +41,26 @@ export class Server {
...
     }
 
-    this.app.use(cors())
+    const corsOptions: cors.CorsOptions = {
+      origin: true,
+      methods: 'GET,HEAD,POST,PUT,DELETE,CONNECT,OPTIONS,TRACE,PATCH',
+      allowedHeaders: [
+        'DNT',
+        'X-Mx-ReqToken',
+        'Keep-Alive',
```
menduz, line 50: This applies to all configs: why do we need those headers with cors?

lambdas/src/Server.ts
```diff
@@ -58,6 +77,8 @@ export class Server {
     const theGraphClient: TheGraphClient = env.getBean(Bean.THE_GRAPH_CLIENT)
     const offChainManager: OffChainWearablesManager = env.getBean(Bean.OFF_CHAIN_MANAGER)
 
+    this.app.options('*', cors(corsOptions))
```
menduz, line 80: Is this correct? Line 63 already registers the middleware for all methods

content/src/Server.ts
```diff
@@ -67,6 +87,7 @@ export class Server {
       Metrics.initialize()
     }
 
+    this.app.options('*', cors(corsOptions))
```
menduz, line 90: Line 80 already registers cors for all methods, this may be unnecesary

menduz (changes requested):
Since it may break, I'd suggest to add some tests for cors headers

lambdas/src/Server.ts
agusaldasoro: I wasn't sure which headers do the client send, and if they are sending a non-allowed Header, it will fail. So, I took a list from common headers and added `X-Peer-Token`

content/src/Server.ts
agusaldasoro: Line 80 registers cors for all methods and paths that are configured later, but the OPTIONS method is never enabled. I could modify the registerRoute method, to also register OPTIONS method for each path, so we can avoid answering any path if that seems better.

menduz: https://github.com/expressjs/cors/blob/master/lib/index.js#L163-L171 

The `cors` package manages the `OPTIONS` method internally

content/src/Server.ts
```diff
@@ -44,6 +44,26 @@ export class Server {
     this.port = env.getConfig(EnvironmentConfig.SERVER_PORT)
 
     this.app = express()
+
+    const corsOptions: cors.CorsOptions = {
+      origin: true,
+      methods: 'GET,HEAD,POST,PUT,DELETE,CONNECT,OPTIONS,TRACE,PATCH',
```
menduz, line 50: OPTIONS should not be part of the allowed methods
```suggestion
      methods: 'GET,HEAD,POST,PUT,DELETE,CONNECT,TRACE,PATCH',
```

lambdas/src/Server.ts
menduz: I'd say we focus on the headers we are interested-in, ignoring what the client actually sends. Otherwise when anyone come to this code in the future would have to research what does `X-Mx-ReqToken` is and why it is there. And it is another thing we will have to maintain indefinitely

menduz (approved):

PR #635, opened 2021-09-02
Parse cache sizes as numbers

agusaldasoro:


content/src/Environment.ts
```diff
@@ -94,7 +94,20 @@ export class Environment {
   logConfigValues() {
     Environment.LOGGER.info('These are the configuration values:')
     for (const [config, value] of this.configs.entries()) {
-      Environment.LOGGER.info(`${EnvironmentConfig[config]}: ${JSON.stringify(value)}`)
+      Environment.LOGGER.info(`${EnvironmentConfig[config]}: ${this.printObject(value)}`)
```
jmoguilevsky, line 97: why not just `JSON.stringify(value, null, 4)`

jmoguilevsky (approved):

PR #645, opened 2021-09-08
chore: remove bazel and setup yarn workspaces and tsc

guidota:
# Description 

This PR removes `bazel` and introduces yarn workspaces with Typescript compiler as the main and only tool generating distribution files.

In order to organize the workspace there are a few changes:
- All packages are defined in the main `package.json` under `workspaces` field.
- All packages have their own `package.json` where runtime dependencies should be declared.
- All packages have their `tsconfig.json` extension.
- Tests depends on build tasks, running Jasmine (framework) against compiled tests.
- Some imports were modified in order to remove `path mappings` which complicate transpiled code references.
- `commons` and `contracts` packages now export everything through an `index.ts` file.

## Things to review
- [x] Circle CI configuration is updated, but we must check everything is OK.
- [x] Dockerfile is updated in order to build all services in the same image but only runtime dependencies are included.
  - This means, docker image size is smaller now  (~400mb) but we should check nothing is missing.
- [x] Catalyst owner repository should not suffer any consequences by this changes since the new image preserves the entrypoints.

## TO DO 
- [x] Update readme/documentation files


menduz:
You may want to delete the .yarn cache folder

guidota:
> You may want to delete the .yarn cache folder

@menduz 
It was intentionally to be there as an offline mirror, check out https://classic.yarnpkg.com/blog/2016/11/24/offline-mirror/ 
Do you think is not a good idea?

menduz:
No, I don't think it is a good idea. If we move to github actions (which we should eventually) the caché is managed by the CI, the package-lock.json or yarnlock files should be more than enough to rebuild the whole dependency tree without bloating the git repository. Having a huge repository plays against a good dev-exp

https://github.com/actions/setup-node
```yaml
- uses: actions/setup-node@v2
  with:
    node-version: '14'
    cache: 'yarn'
```

jmoguilevsky (approved):

.eslintrc.json
```diff
@@ -1,14 +1,11 @@
 {
-  "parser": "@typescript-eslint/parser", // Specifies the ESLint parser
   "parserOptions": {
     "ecmaVersion": 2020, // Allows for the parsing of modern ECMAScript features
     "sourceType": "module" // Allows for the use of imports
   },
   "extends": [
     "plugin:@typescript-eslint/recommended", // Uses the recommended rules from the @typescript-eslint/eslint-plugin
-    "prettier/@typescript-eslint", // Uses eslint-config-prettier to disable ESLint rules from @typescript-eslint/eslint-plugin that would conflict with prettier
```
agusaldasoro, line 9: why is this change?

.eslintrc.json
```diff
@@ -1,14 +1,11 @@
 {
-  "parser": "@typescript-eslint/parser", // Specifies the ESLint parser
```
agusaldasoro, line 2: why is this change? Aren't we using eslint as pre-commit hook?

.gitignore
```diff
@@ -10,3 +9,5 @@ local/certbot/conf/**
 local/nginx/conf.d/00-katalyst.conf
 tmpbin
 linked-peer-package
+.husky
```
agusaldasoro, line 12: I'm not sure that this file should be ignored, that's where the configuration for the pre commit hook is stored

Dockerfile
```diff
@@ -1,30 +1,54 @@
-FROM node:14.16.1-slim
-
-RUN apt-get update && \
-  apt-get upgrade -yq && \
-  apt-get install -yq yarn git zlib1g zlib1g-dev
-
+FROM node:14-alpine as base
```
agusaldasoro, line 1: what's the difference between slim and alpine?

Dockerfile
```diff
@@ -1,30 +1,54 @@
...
-  rm -rf /root/.cache/bazel
-
-WORKDIR /app
+# build sources
+FROM base as catalyst-builder
+COPY . .
+RUN yarn install
+
+RUN yarn workspace @catalyst/contracts build
+RUN yarn workspace @catalyst/commons build
```
agusaldasoro, line 16: What are those two doing? Was that already done? I'm not sure if we are using the contracts folder

.dockerignore
```diff
@@ -1,9 +1,8 @@
...
-bazel-out
 .git
 .gitignore
 .vscode
 node_modules
 *.log
 tmpbin
-comms/performance-test
\ No newline at end of file
+comms/performance-test
```
agusaldasoro, line 8: Wasn't this folder deleted?Shouldn't we remove it from here too?

agusaldasoro (changes requested):

pablitar (approved):
Awesome work! This is great!

.eslintrc.json
guidota: `eslint-config-prettier` was updated in this PR and this version includes it:
`
"prettier/@typescript-eslint" has been merged into "prettier" in eslint-config-prettier 8.0.0
`

Dockerfile
guidota: `commons` package depends on `contracts`, but not sure if we need to build them explicitly, will check if we can avoid doing this.

Dockerfile
guidota: `alpine` image uses a different linux distribution (Alpine) and size is smaller.

do you think that could be a concern for us?

agusaldasoro (approved):
Great Work!!!

Dockerfile
agusaldasoro: no, it shouldn't be

Issue #450, opened 2021-04-07
Add catalyst version to the docker image


nchamo:
Since https://github.com/decentraland/catalyst/pull/363, we now have the ability to create versions of our catalyst docker images. What we need to do now is take the tag's semver version and send it to the catalyst script that generates the image, so that it can be exposed through the API.

Where to do it: https://github.com/decentraland/catalyst/blob/master/prepare_for_docker_image.sh#L8

### Questions?
What about images created by PRs? And what about the ones created on master? What version would those take? How will that work with the minimum version in the explorer?

PR #648, opened 2021-09-08
chore: automatic docker versioning

menduz:
Fixes https://github.com/decentraland/catalyst/issues/450


menduz:
Are we merging this?

PR #649, opened 2021-09-10
Explicit declare db priorities and add timeout to queue

agusaldasoro:
## What changes does this PR include?
- Remove default priority for database queries, you need to declare if the query will be queued as HIGH or LOW priority.
- Add a timeout for any entry in the database queue, default 1min.
- Changed default for max concurrency to 50 and for max queued to 300.

## How are queries classified now between priorities?
The queries are classified according to which endpoint they are used from:

### Endpoints that don't access the database
- `/status`
- `/challenge`
- `/snapshot/:type`

### High priority endpoints
- `/entities/:type`
- `/entities`
- `/contents/:hashId`
- `/available-content`
- `/audit/:type/:entityId`
- `/deployments`
- `PUT /denylist/:type/:id`
- `DELETE /denylist/:type/:id`

### Low priority endpoints
- `/contents/:hashId/active-entities` (endpoint for debugging purposes)
- `/denylist` (get all denylist targets, not used for filtering)
- `HEAD /denylist/:type/:id`
- `/failed-deployments` (endpoint for debugging purposes)
- `/pointer-changes` (not currently used for sync)

## Note
Most of the content endpoints will be now queued as a high priority, we should consider removing the priorities at all, as the low endpoints are not frequently used.


content/src/denylist/ActiveDenylist.ts
```diff
@@ -73,22 +73,31 @@ export class ActiveDenylist extends Denylist {
...
     return { status: DenylistOperationStatus.OK, type: DenylistValidationType.SIGNATURE_VALIDATION }
   }
 
   async getAllDenylistedTargets(): Promise<{ target: DenylistTarget; metadata: DenylistMetadata }[]> {
     if (this.empty) {
       return []
     }
-    return this.repository.run((db) => db.denylist.getAllDenylistedTargets())
+    return this.repository.run((db) => db.denylist.getAllDenylistedTargets(), {
+      priority: DB_REQUEST_PRIORITY.LOW
```
pablitar, line 90: Why is this low? Isn't this used when fetching content or something like that?

pablitar (approved):
LGTM!

content/src/denylist/ActiveDenylist.ts
agusaldasoro: Not really, the method `getAllDenylistedTargets` is only used when fetching `/denylist` that lists every element in the denylist. To check if content should be retrieved we are using: https://github.com/decentraland/catalyst/pull/649/files#diff-dd8d71d82d98295be817d09a1d0edc30cc893b9611b53511e48b7b6e5d4508c7R57

PR #656, opened 2021-09-15
Add stream timeout

jmoguilevsky:
In this PR:
- Add a stream timeout to prevent that the stream stays open
- Add an env var to handle timeout time


agusaldasoro (approved):

pablitar (approved):
LGTM! Let's test this in dev!

content/src/service/synchronization/SynchronizationManager.ts
```diff
@@ -23,15 +24,40 @@ export class ClusterSynchronizationManager implements SynchronizationManager {
...
     private readonly systemProperties: SystemPropertiesManager,
     private readonly deployer: EventDeployer,
     private readonly timeBetweenSyncs: number,
-    private readonly disableSynchronization: boolean
+    private readonly disableSynchronization: boolean,
+    private readonly syncStreamTimeout: string
   ) {}
 
+  timeoutStreams(streams: Transform[]): void {
+    this.timeoutStreamId = setTimeout(() => {
```
menduz, line 39: Since this function is called from an async function, I'd suggest to return the timer instead of using the `this` object to prevent the logic from breaking due to asynchronous conditions.

Also, I'd create a free function, avoiding methods when possible

content/src/service/synchronization/SynchronizationManager.ts
```diff
@@ -93,6 +119,8 @@ export class ClusterSynchronizationManager implements SynchronizationManager {
         return deploymentStream.pipe(sourceData)
       })
 
+      this.timeoutStreams(streams)
+
       // Process them together
       await this.deployer.processAllDeployments(streams)
```
menduz, line 125: I'd move the timeout logic inside `this.deployer.processAllDeployments`. That should handle the timeout of the received streams. This file may be too high level

menduz (changes requested):

content/src/service/synchronization/streaming/StreamHelper.ts
```diff
@@ -28,7 +28,7 @@ export function streamMap<T1, T2>(mapping: (value: T1) => T2) {
   })
 }
 
-export function mergeStreams(readables: Readable[]): NodeJS.ReadableStream {
+export function mergeStreams(readables: Readable[]): Readable {
```
jmoguilevsky, line 31: I updated this, becase ReadableStream doesn't have the method destroy

And checked here `https://github.com/grncdr/merge-stream/blob/master/index.js` that the library returns a Passthrough wich is also a Readable

menduz (approved):

PR #659, opened 2021-09-16
Exit if not sync

agusaldasoro:
Exiting with failure if the sync hangs for more than 20minutes (this is configurable though `CHECK_SYNC_RANGE` env var).


pablitar (approved):
LGTM!

PR #661, opened 2021-09-16
feat: add openapi express middleware to validate requests

### LINKED ISSUE #638: Validate Catalyst server with the API spec
We need to create a API spec module and include it on the Catalyst server so that we can validate that we are not breaking server API when introducing changes 

https://github.com/decentraland/catalyst-api-specs

guidota:
Resolves #638

### DIFF
--- a/comms/lighthouse/src/server.ts
+++ b/comms/lighthouse/src/server.ts
@@ -1,8 +1,9 @@
-import { DAOContractClient, DECENTRALAND_ADDRESS } from '@catalyst/commons'
-import { initializeMetricsServer } from '@catalyst/commons'
+import { DAOContractClient, DECENTRALAND_ADDRESS, initializeMetricsServer } from '@catalyst/commons'
 import { DAOContract } from '@catalyst/contracts'
+import { COMMS_API } from '@dcl/catalyst-api-specs'
 import cors from 'cors'
 import express from 'express'
+import * as OpenApiValidator from 'express-openapi-validator'
 import morgan from 'morgan'
 import * as path from 'path'
 import { ConfigService } from './config/configService'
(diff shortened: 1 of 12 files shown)


comms/lighthouse/src/server.ts
```diff
@@ -60,6 +62,13 @@ async function main() {
   if (accessLogs) {
     app.use(morgan('combined'))
   }
+  if (validateAPI) {
+    OpenApiValidator.middleware({
+      apiSpec: COMMS_API,
+      // validateResponses: true, // false by default
```
agusaldasoro, line 68: I'd never validate responses in code, only in tests

content/src/Environment.ts
```diff
@@ -406,6 +407,8 @@ export class EnvironmentBuilder {
       () => new Map(Object.entries(process.env).filter(([name]) => name.startsWith('CACHE')))
     )
 
+    this.registerConfigIfNotAlreadySet(env, EnvironmentConfig.VALIDATE_API, () => process.env.VALIDATE_API !== 'false')
```
agusaldasoro, line 410: Why do we need this to be configurable? Is it only for releasing it?

content/src/controller/Controller.ts
```diff
@@ -412,7 +412,7 @@ export class Controller {
     // Path: /deployments
     // Query String: ?from={timestamp}&toLocalTimestamp={timestamp}&entityType={entityType}&entityId={entityId}&onlyCurrentlyPointed={boolean}&deployedBy={ethAddress}
 
-    const stringEntityTypes = this.asArray<string>(req.query.entityType as string)
+    const stringEntityTypes = this.asArray<string>(req.query.entityType as string | string[])
```
agusaldasoro, line 415: can this turn into an array of arrays?

content/test/integration/E2ETestEnvironment.ts
```diff
@@ -65,7 +65,7 @@ export class E2ETestEnvironment {
   }
 
   async clearDatabases(): Promise<void> {
-    await this.repository.run((db) => db.query(`DROP SCHEMA ${E2ETestEnvironment.TEST_SCHEMA} CASCADE`), {
+    await this.repository.run((db) => db.query(`DROP SCHEMA IF EXISTS ${E2ETestEnvironment.TEST_SCHEMA} CASCADE`), {
```
agusaldasoro, line 68: 👍

agusaldasoro (approved):

content/src/controller/Controller.ts
guidota: As I can see in the method `asArray`, it shouldn't:

```ts
export function asArray<T>(elements: T[] | T): T[] {
  if (!elements) {
    return []
  }
  if (elements instanceof Array) {
    return elements
  }
  return [elements]
}
```

comms/lighthouse/src/server.ts
menduz:
```suggestion
      validateResponses: process.env.CI == 'true' // or isRunningTests()
```

PR #664, opened 2021-09-21
feat: add metadata schema validation for v4 entities deployment

### LINKED ISSUE #527: Add type/metadata validation to content server
Right now, an entity's metadata can be anything. There is no validation at all.

Since we are interested in open our infrastructure for other projects, we need to be more careful in what we allow.

The idea would be to:
* Validate the metadata's schema
* Maybe validate the files being added. For example, in the case of profiles, only snapshot files should be allowed

This breaking change should be done together with https://github.com/decentraland/catalyst/issues/198 and https://github.com/decentraland/catalyst/issues/200

guidota:
## Description

Add metadata schema validation for v4 entities deployment.
Resolves #527.

### DIFF
--- a/content/package.json
+++ b/content/package.json
@@ -16,6 +16,7 @@
     "@catalyst/commons": "0.1.0",
     "@catalyst/contracts": "0.1.0",
     "@dcl/catalyst-api-specs": "1.0.1",
+    "@dcl/schemas": "1.7.1-20210921181546.commit-b869b84",
     "@dcl/urn-resolver": "1.2.1",
     "aws-sdk": "2.854.0",
     "compression": "1.7.4",
--- a/content/package.json
+++ b/content/package.json
@@ -16,7 +16,7 @@
     "@catalyst/commons": "0.1.0",
     "@catalyst/contracts": "0.1.0",
     "@dcl/catalyst-api-specs": "1.0.1",
-    "@dcl/schemas": "1.7.1-20210921181546.commit-b869b84",
+    "@dcl/schemas": "1.8.0",
     "@dcl/urn-resolver": "1.2.1",
     "aws-sdk": "2.854.0",
     "compression": "1.7.4",
--- a/content/src/service/ServiceImpl.ts
+++ b/content/src/service/ServiceImpl.ts
@@ -1,3 +1,4 @@
+import { IPFSv2 } from '@dcl/schemas'
 import {
   AuditInfo,
   ContentFileHash,
@@ -9,8 +10,8 @@ import {
   ServerStatus
 } from 'dcl-catalyst-commons'
 import log4js from 'log4js'
-import { metricsComponent } from '../metrics'
 import { CURRENT_CONTENT_VERSION } from '../Environment'
+import { metricsComponent } from '../metrics'
 import { Database } from '../repository/Database'
 import { Repository } from '../repository/Repository'
 import { DB_REQUEST_PRIORITY } from '../repository/RepositoryQueue'
@@ -324,7 +325,7 @@ export class ServiceImpl implements MetaverseContentService, ClusterDeploymentsS
   }
 
   static isIPFSHash(hash: string) {
-    return hash.startsWith('bafy') && hash.length === 59
+    return IPFSv2.validate(hash)
   }
 
   getContent(fileHash: ContentFileHash): Promise<ContentItem | undefined> {
--- a/content/src/service/validations/Validations.ts
+++ b/content/src/service/validations/Validations.ts
@@ -148,15 +148,13 @@ export class Validations {
       if (entity.type === EntityType.PROFILE) {
         // Validate all content files correspond to at least one avatar snapshot
         if (Profile.validate(entity.metadata)) {
-          Array.from(entity.content.keys())
-            .filter((fileName) => {
-              const fileNameWithoutExtension = fileName.replace(/.[^/.]+$/, '')
-              const testIsSnapshot = (avatar: Avatar) => avatar.avatar.snapshots[fileNameWithoutExtension] !== undefined
-              return entity.metadata.avatars.find(testIsSnapshot) === undefined
-            })
-            .forEach((fileName) => {
-              errors.push(`This file is not expected: ${fileName}. Please, include only snapshot files.`)
-            })
+          Array.from(entity.content.entries())
+            .filter(([fileName, hash]) => !Validations.correspondsToASnapshot(fileName, hash, entity.metadata))
+            .forEach(([fileName, hash]) =>
+              errors.push(
+                `This file is not expected: '${fileName}' or its hash is invalid: '${hash}'. Please, include only valid snapshot files.`
+              )
+            )
         }
       }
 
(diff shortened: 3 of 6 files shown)


content/package.json
```diff
@@ -16,6 +16,7 @@
     "@catalyst/commons": "0.1.0",
     "@catalyst/contracts": "0.1.0",
     "@dcl/catalyst-api-specs": "1.0.1",
+    "@dcl/schemas": "1.7.1-20210921181546.commit-b869b84",
```
agusaldasoro, line 19: Should we use a release version?

content/src/service/validations/Validations.ts
```diff
@@ -144,6 +145,21 @@ export class Validations {
...
         )
 
+      if (entity.type === EntityType.PROFILE) {
+        // Validate all content files correspond to at least one avatar snapshot
+        if (Profile.validate(entity.metadata)) {
+          Array.from(entity.content.keys())
+            .filter((fileName) => {
+              const fileNameWithoutExtension = fileName.replace(/.[^/.]+$/, '')
+              const testIsSnapshot = (avatar: Avatar) => avatar.avatar.snapshots[fileNameWithoutExtension] !== undefined
+              return entity.metadata.avatars.find(testIsSnapshot) === undefined
```
agusaldasoro, line 155: can we extract these two lines to a method so it can be more legible?

agusaldasoro: Also, please validate hashes from the files and not only the name

content/test/unit/service/validations/Validations.spec.ts
```diff
@@ -13,6 +15,44 @@ import {
 } from '../../../../src/service/validations/Validator'
 import { MockedAccessChecker } from '../../../helpers/service/access/MockedAccessChecker'
 
+const avatarInfo = {
+  bodyShape: 'urn:decentraland:off-chain:base-avatars:BaseMale',
+  snapshots: {
+    face: 'https://peer.decentraland.org/content/contents/QmZdwrWnF2kLghFJ9kSj2brFEmywfAiqssr2LCqFj9HVWi',
```
agusaldasoro, line 21: nit: those are hashes not links in the content server

agusaldasoro: Should we validate that on the schema also? 🤔

content/test/unit/service/validations/Validations.spec.ts
```diff
@@ -401,7 +571,7 @@ function buildEntity(options?: {
   )
   return {
     ...opts,
-    type: EntityType.SCENE
+    type: options?.type ?? EntityType.SCENE
```
agusaldasoro, line 574: why do we need this to be added after? Can't it be on the first object with the assignment?

agusaldasoro (changes requested):
Great work! Just a few comments

content/test/unit/service/validations/Validations.spec.ts
guidota: ok, that's good to know, we should!

content/test/unit/service/validations/Validations.spec.ts
guidota: good point 👍

content/test/unit/service/validations/Validations.spec.ts
guidota: it seems too restrictive to use a regex as schema validation for these `hashes`, since it can be ipfs v1 or v2 (different formats) or a different algorithm in the future.

do you think we should add another validation method for that?

agusaldasoro: That's true, but in v4 it must use IPFS. So, maybe we could implement the validation to be similar to what we already have on the content server:

```
  static isIPFSHash(hash: string) {
    return hash.startsWith('bafy') && hash.length === 59
  }
```

content/src/service/validations/Validations.ts
```diff
@@ -145,6 +145,21 @@ export class Validations {
...
+            })
+            .forEach((fileName) => {
+              errors.push(`This file is not expected: ${fileName}. Please, include only snapshot files.`)
+            })
```
menduz: What do you think of using plain fors instead of chained callbacks? In big chunks, performance is best if we are not calling functions and also readability is better

```suggestion
          for (const fileName of entity.content.keys()) {
              const fileNameWithoutExtension = fileName.replace(/.[^/.]+$/, '')
              const testIsSnapshot = (avatar: Avatar) => avatar.avatar.snapshots[fileNameWithoutExtension] !== undefined
              if (entity.metadata.avatars.find(testIsSnapshot) === undefined) {
                errors.push(`This file is not expected: ${fileName}. Please, include only snapshot files.`)
              }
          }
```

content/test/unit/service/validations/Validations.spec.ts
guidota: once [this PR](https://github.com/decentraland/common-schemas/pull/20) is merged, will update this function to use the schema validation to avoid duplicated logic

agusaldasoro: nicee

content/src/service/validations/Validations.ts
guidota: I agree readability is better and performance may improve too.

In this case I tried to do the same way as it was above (same method, multiple validations being done using these `filter` and `forEach` calls), otherwise it would end up with mixed approaches. 

If you think it worth to make these validations in a more clear way, I'm happy to open another PR for that.

content/src/service/validations/Validations.ts
```diff
@@ -144,6 +145,19 @@ export class Validations {
...
+      if (entity.type === EntityType.PROFILE) {
+        // Validate all content files correspond to at least one avatar snapshot
+        if (Profile.validate(entity.metadata)) {
+          Array.from(entity.content.entries())
+            .filter(([fileName, hash]) => !Validations.correspondsToASnapshot(fileName, hash, entity.metadata))
+            .forEach(([fileName, hash]) =>
+              errors.push(
+                `This file is not expected: '${fileName}' or its hash is invalid: '${hash}'. Please, include only valid snapshot files.`
+              )
+            )
```
menduz, line 157:
```suggestion
          for (const [fileName, hash] of entity.content.entries()) {
            if (!Validations.correspondsToASnapshot(fileName, hash, entity.metadata)) {
              errors.push(
                `This file is not expected: '${fileName}' or its hash is invalid: '${hash}'. Please, include only valid snapshot files.`)
            }
          }
```

content/src/service/ServiceImpl.ts
```diff
@@ -324,7 +325,7 @@ export class ServiceImpl implements MetaverseContentService, ClusterDeploymentsS
   }
 
   static isIPFSHash(hash: string) {
-    return hash.startsWith('bafy') && hash.length === 59
+    return IPFSv2.validate(hash)
```
agusaldasoro, line 328: 👏

agusaldasoro (approved):
Great work!!

PR #666, opened 2021-09-27
Add a rate limit for deploying profiles

agusaldasoro:
Add a new validation for V3 and V4 entities which rate limits the number of deployments. It only allows making a deploy per pointer per TTL configured and a max amount of deployments too.

### DIFF
--- a/content/src/service/validations/Validations.ts
+++ b/content/src/service/validations/Validations.ts
@@ -192,6 +192,13 @@ export class Validations {
       return [`The metadata for this entity type (${deployment.entity.type}) is not valid.`]
   }
 
+  /** Validate the deployment is not rate limited */
+  static readonly RATE_LIMIT: Validation = async ({ deployment, externalCalls }) => {
+    if (!(await externalCalls.isEntityRateLimited(deployment.entity))) {
+      return [`The entity with id (${deployment.entity.id}) has been rate limited.`]
+    }
+  }
+
   private static correspondsToASnapshot(fileName: string, hash: string, metadata: Profile) {
     const fileNameWithoutExtension = fileName.replace(/.[^/.]+$/, '')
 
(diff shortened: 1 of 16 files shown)


content/src/service/validations/Validations.ts
```diff
@@ -121,12 +122,38 @@ export class Validations {
   }
 
   /** Validate that uploaded and reported hashes are corrects */
-  static readonly CONTENT: Validation = async ({ deployment, externalCalls }) => {
+  static readonly CONTENT_V3: Validation = async ({ deployment, externalCalls }) => {
```
agusaldasoro, line 125: We had to remove the check for profile files from v3 as we are not sure that all existing v3 deployments fulfills that condition

content/src/service/validations/validations/ValidationsV4.ts
```diff
@@ -4,43 +4,53 @@ import { ValidationsForContext } from '../Validator'
 
 export const VALIDATIONS_V4: ValidationsForContext = {
```
agusaldasoro, line 5: To support deploying V4 entities we need to be sure that all validations are done, if not we may have invalid v4 entities as not all validations were implemented

content/src/service/validations/Validations.ts
```diff
@@ -150,9 +177,12 @@ export class Validations {
           errors.push(`This hash was uploaded but is not referenced in the entity: ${hash}`)
         }
       }
-
-      return errors.length > 0 ? errors : undefined
+    } else if (files) {
+      for (const [hash] of files) {
+        errors.push(`This hash was uploaded but is not referenced in the entity: ${hash}`)
+      }
```
guidota, line 183: I think we can move this logic to avoid duplicated code, what do you say?

guidota (approved):
LGTM 👍

PR #671, opened 2021-09-28
feat: Profiles lambdas cache header

### LINKED ISSUE #667: Add cache for lambdas/profiles endpoint
In Lambdas server, profiles is currently one of the most expensive endpoints:
image

This is also one of the endpoints that we can't cache in Cloudflare or similar.

We can, though, implement an in-memory cache or similar that fetches the entity from content server and sees if the timestamp has been updated since last time. If not, it can return the cached entity and avoid the expensive calculations that follow. If changed, it can invalidate and recalculate.

pablitar:
Closes #667

### DIFF
--- a/lambdas/src/apis/profiles/controllers/profiles.ts
+++ b/lambdas/src/apis/profiles/controllers/profiles.ts
@@ -20,19 +20,24 @@ function setCacheHeaders(res: Response, profiles: ProfileMetadata[]) {
   }
 
   if (maxTimestamp) {
+    // These headers are HTTP standard. See the links below for more information about how they are used
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
     res.header('Last-Modified', new Date(maxTimestamp).toUTCString())
     res.header('Cache-Control', 'public, max-age=0,s-maxage=0')
   }
 }
 
 function getIfModifiedSinceTimestamp(req: Request): number | undefined {
+  // This is a standard HTTP header. See the link below for more information
+  // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since
   const headerValue = req.header('If-Modified-Since')
   if (!headerValue) return
   try {
     const timestamp = Date.parse(headerValue)
     return isNaN(timestamp) ? undefined : timestamp
   } catch (e) {
-    LOGGER.warn('Received an invalid header for IfModifiedSince ', headerValue)
+    LOGGER.warn('Received an invalid header for If-Modified-Since ', headerValue)
   }
 }
 
@@ -46,6 +51,9 @@ function sendProfilesResponse(res: Response, profiles: ProfileMetadata[] | undef
       res.send(profiles)
     }
   } else {
+    // The only case in which we receive undefined profiles is when no profile was updated after de If-Modified-Since specified moment.
+    // In this case, as per spec, we return 304 (not modified) and empty body
+    // See here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since
     res.status(304).send()
   }
 }
--- a/lambdas/src/apis/profiles/controllers/profiles.ts
+++ b/lambdas/src/apis/profiles/controllers/profiles.ts
@@ -11,6 +11,45 @@ import { WearablesOwnership } from '../WearablesOwnership'
 
 const LOGGER = log4js.getLogger('profiles')
 
+function setCacheHeaders(res: Response, profiles: ProfileMetadata[]) {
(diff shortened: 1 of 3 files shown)


lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -11,6 +11,45 @@ import { WearablesOwnership } from '../WearablesOwnership'
...
+function setCacheHeaders(res: Response, profiles: ProfileMetadata[]) {
+  if (profiles.length === 0) return
+
+  let maxTimestamp = profiles[0]?.timestamp
+  for (let i = 1; i < profiles.length; i++) {
+    maxTimestamp = Math.max(maxTimestamp, profiles[i].timestamp)
+  }
+
+  if (maxTimestamp) {
+    res.header('Last-Modified', new Date(maxTimestamp).toUTCString())
```
agusaldasoro, line 23: Please add a comment on how we will be using this headers

lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -11,6 +11,45 @@ import { WearablesOwnership } from '../WearablesOwnership'
 
 const LOGGER = log4js.getLogger('profiles')
 
+function setCacheHeaders(res: Response, profiles: ProfileMetadata[]) {
+  if (profiles.length === 0) return
+
+  let maxTimestamp = profiles[0]?.timestamp
+  for (let i = 1; i < profiles.length; i++) {
+    maxTimestamp = Math.max(maxTimestamp, profiles[i].timestamp)
```
agusaldasoro, line 19: You are using this as you can ensure that it was valid when was deployed?

lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -43,21 +88,31 @@ export async function getProfilesById(
...
-    const profiles = await fetchProfiles(profileIds, client, ensOwnership, wearables)
-    res.send(profiles)
+    const profiles = await fetchProfiles(profileIds, client, ensOwnership, wearables, getIfModifiedSinceTimestamp(req))
+    sendProfilesResponse(res, profiles)
   }
 }
 
+// Dates received from If-Modified-Since headers have precisions of seconds, so we need to round
+function roundToSeconds(timestamp: number) {
+  return Math.floor(timestamp / 1000) * 1000
```
agusaldasoro, line 98: are you diving and multiplying for the same number here?

menduz:
How are you responding to the requests with cache hit?

pablitar:
> How are you responding to the requests with cache hit?

@menduz What do you mean?

Requests that have not been modified since the timestamp sent in "If-Modified-Since" respond 304 with empty content, as per specification: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since. We know they haven't been modified because we have the entity timestamp.

lambdas/src/apis/profiles/controllers/profiles.ts
pablitar: These headers are actually HTTP standard: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified

I can add a link to the documentation though. That'll probably be useful

lambdas/src/apis/profiles/controllers/profiles.ts
pablitar: Yep. We can assume that if we validated once and it was not updated since, then it is valid now. We don't need to perform the expensive calculations below.

lambdas/src/apis/profiles/controllers/profiles.ts
pablitar: We are rounding the timestamp to seconds. So for instance, 1632847801743 will become 1632847801000.

This is because the "If-Modified-Since" header comes with seconds precision: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since. We don't have milliseconds precision there.

lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -11,6 +11,45 @@ import { WearablesOwnership } from '../WearablesOwnership'
...
+  if (profiles) {
+    setCacheHeaders(res, profiles)
+    if (singleProfile) {
+      const returnProfile: ProfileMetadata = profiles[0] ?? { avatars: [], timestamp: 0 }
+      res.send(returnProfile)
+    } else {
+      res.send(profiles)
+    }
+  } else {
+    res.status(304).send()
```
agusaldasoro, line 49: Please add a comment here explaining this case

agusaldasoro (approved):

lambdas/src/apis/profiles/controllers/profiles.ts
menduz: Oh here it was. this is what I was asking in the previous comment

PR #673, opened 2021-09-29
feat: validate v4 entities size including previous deployments

### LINKED ISSUE #198: fix: enforce 15MB limit of pointers
Right now, the content server has a limit of 15MB per pointer being deployed. In scene terms, it means that a scene can upload up to 15MB per occupied parcel. The problem here, is that there is a way to exploit this limitation. The limit is being calculated on files being uploaded, not on all files being used by the scene. So a scene could upload some of the files on one deployment, and then upload the rest on another deployment, and therefore circumvent the limit.

This "hack" has been abused by the community and the dcl foundation, so the first step would be to investigate the current state of the world, and figure out which and how many scenes are over said limit.

Then, we will need to make a decision on how to reinforce this limit, by making sure that the already deployed scenes remain valid, while new ones fail. This has to be properly discussed and communicated to everyone involved, because updates to already deployed scenes could fail. 

A possible fix will probably require the content server to store the file size of each uploaded file in the database, to make future validations quicker.

guidota:
## Description

Introduces entity v4 deployment size validation.
Takes into account all content files and not just the uploaded files in current deployment.

Resolves #198

### DIFF
--- a/content/.gitignore
+++ b/content/.gitignore
@@ -1,1 +1,1 @@
-storage*
+storage/*
--- a/content/src/service/ServiceImpl.ts
+++ b/content/src/service/ServiceImpl.ts
@@ -181,7 +181,7 @@ export class ServiceImpl implements MetaverseContentService, ClusterDeploymentsS
             isEntityDeployedAlready: (entityIdToCheck: EntityId) =>
               Promise.resolve(isEntityAlreadyDeployed && entityId === entityIdToCheck),
             isEntityRateLimited: (entity) => Promise.resolve(this.isEntityRateLimited(entity)),
-            fetchContentFileSize: async (hash) => (await this.getContent(hash))?.getLength() ?? 0
+            fetchContentFileSize: async (hash) => (await this.getSize(hash)) ?? 0
           })
 
           if (!validationResult.ok) {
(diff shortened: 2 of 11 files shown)


content/.gitignore
```diff
@@ -1 +1 @@
-storage*
+storage/*
```
menduz, line 1: I'm curious, why you always change this??

guidota: haha, good question... 
it was wrong and `content/src/storage` (code) was being ignored by linters but inner modifications were being tracked by git since it's already tracked

by default, content service is using `./storage` folder to store all content files, so every time I start it in dev mode it is creating this folder inside `content/`

this `.gitignore` was introduced when we removed `bazel` which run services in other context and `storage` folder lives in other place

content/src/service/ServiceImpl.ts
```diff
@@ -180,7 +180,8 @@ export class ServiceImpl implements MetaverseContentService, ClusterDeploymentsS
             isContentStoredAlready: () => Promise.resolve(alreadyStoredContent),
             isEntityDeployedAlready: (entityIdToCheck: EntityId) =>
               Promise.resolve(isEntityAlreadyDeployed && entityId === entityIdToCheck),
-            isEntityRateLimited: (entity) => Promise.resolve(this.isEntityRateLimited(entity))
+            isEntityRateLimited: (entity) => Promise.resolve(this.isEntityRateLimited(entity)),
+            fetchContentFileSize: async (hash) => (await this.getSize(hash)) ?? 0
```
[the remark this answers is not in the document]
guidota: you're right! fixing it

agusaldasoro (approved):
Great work!

PR #678, opened 2021-10-01
Stream exception handling

jmoguilevsky:
Add handling to unexpected exception, and use the stream promises api


content/src/Server.ts
```diff
@@ -49,7 +49,9 @@ export class Server {
 
     this.app = express()
 
-    this.metricsServer = initializeMetricsServer(this.app, metricsComponent)
+    if (env.getConfig(EnvironmentConfig.METRICS)) {
```
menduz, line 52: What is the case in which you _don't_ want metrics that is good enough to add an if in the code?

jmoguilevsky: This comes from @guidota to fix the running of the service, maybe it's not needed anymore

guidota: for testing purposes we don't want to initialize multiple metrics servers (it fails locally), but did some changes to avoid using this environment config and just skip it for test servers

see #694

PR #694, opened 2021-10-04
feat: better jasmine and vscode configuration

guidota:
## Description

- Improves test debugging experience.
  - VSCode recommended extension added for tests discovery and UI:

vscode-debug

- It also adds one check to not initialize metrics server when running tests, since it was failing when running them locally.
- CircleCi now uses a postgres instance to run content integration tests, instead of using DiD (docker in docker) which was creating and deleting a container for each test.


content/package.json
```diff
@@ -9,8 +9,8 @@
     "start:db": "node ./dist/src/entrypoints/run-local-database.js",
     "copy-test-resources": "yarn cpy test/integration/resources/* dist --parents",
     "test": "run-p 'test:*'",
-    "test:unit": "yarn build && jasmine --config=jasmine.unit.json",
-    "test:integration": "yarn copy-test-resources && jasmine --config=jasmine.integration.json"
+    "test:unit": "yarn build && jasmine-ts --config=jasmine.unit.json",
```
agusaldasoro, line 12: Do we still need to run the build here? If that's the case, then why is not needed during integration?

content/src/Server.ts
```diff
@@ -49,7 +49,9 @@ export class Server {
 
     this.app = express()
 
-    this.metricsServer = initializeMetricsServer(this.app, metricsComponent)
+    if (env.getConfig(EnvironmentConfig.METRICS)) {
```
agusaldasoro, line 52: I think that we are deprecating that config, as metrics will be always enabled. Can we make something only for running tests?

package.json
```diff
@@ -19,7 +19,11 @@
     "start:lambdas": "yarn workspace @catalyst/lambdas-server start",
     "start:content": "yarn workspace @catalyst/content-server start",
     "start:comms": "yarn workspace @catalyst/lighthouse-server start",
-    "test": "yarn workspaces run test",
+    "test": "run-s 'test:*'",
+    "test:lambdas": "yarn workspace @catalyst/lambdas-server test",
+    "test:content": "yarn workspace @catalyst/content-server test",
+    "test:comms": "yarn workspace @catalyst/lighthouse-server test",
+    "test:commons": "yarn workspace @catalyst/commons test",
```
agusaldasoro, line 26: Love this!!

agusaldasoro (changes requested):
Please don't use METRICS environment variable for configuration, let's use something special just for tests instead.

agusaldasoro (approved):
Great work!

PR #726, opened 2021-10-15
feat: compose wearable thumbnail with rarity background

### LINKED ISSUE #753: Generate image with rarity background on Lambdas
- Autogenerate the image with the background based on the item rarity.
- Check on deployments if the rarity is the same for the item by using the subgraph
- Check if we need to change the erc721 lambda where we return the image files.
- Add cache (maybe with e-tags)

guidota:
## Description

`/collections/contents/:urn/image` endpoint refactor to create image with rarity background at runtime.

### Tasks
- [x] Refactor
- [x] Rarity background images
- [x] API Changes documentation ([PR](https://github.com/decentraland/catalyst-api-specs/pull/12))
- [x] Add size parameter to `/thumbnail` endpoint
- [x] ERC721 standard metadata endpoint should point to correct thumbnail/image endpoints

Closes #753 

### Converted to draft in order to review missing points
- [ ] Destroy streams on response finished
- [ ] Move cache logic to another file
- [ ] Review env var LAMBDAS_STORAGE_LOCATION


lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -126,26 +133,30 @@ export async function contentsImage(client: SmartContentClient, req: Request, re
...
   }
 }
 
-function getRarityImagePath(wearableMetadata: WearableMetadata) {
-  return `lambdas/resources/${wearableMetadata.rarity}.png`
+function sendImageBuffer(res: Response, image: Buffer, urn: string) {
+  res.send(image)
+
+  res.writeHead(200, {
+    'Content-Type': 'arraybuffer',
```
menduz, line 151: content-type: image/png

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -94,16 +94,23 @@ export async function contentsThumbnail(client: SmartContentClient, req: Request
...
-
   try {
     const entity = await fetchEntity(client, urn)
 
     const wearableMetadata: WearableMetadata = entity.metadata
     const hash = getFileHash(entity, wearableMetadata.thumbnail)
 
-    const headers: Map<string, string> = await client.pipeContent(hash, res as any as ReadableStream<Uint8Array>)
-    headers.forEach((value, key) => res.setHeader(key, value))
+    if (req.query.size) {
```
menduz, line 103: I suggest to use a `/:size/` url param instead of query param to enable edge caching, QueryParameters are hardly cacheable and preferably avoided

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -126,26 +133,30 @@ export async function contentsImage(client: SmartContentClient, req: Request, re
...
 }
 
-function getRarityImagePath(wearableMetadata: WearableMetadata) {
-  return `lambdas/resources/${wearableMetadata.rarity}.png`
+function sendImageBuffer(res: Response, image: Buffer, urn: string) {
+  res.send(image)
+
+  res.writeHead(200, {
+    'Content-Type': 'arraybuffer',
+    ETag: urn,
```
menduz, line 152: `ETag` spec requires the string to be double quoted and it must include the variant size

```diff
-ETag: urn
+ETag: JSON.stringify(urn + "-" + size)
```

guidota:
We need to make sure results can be cached, it can take up to **1s** to compose image with background. Otherwise we need to store them in content storage. ⚠️

lambdas/src/apis/collections/routes.ts
```diff
@@ -12,11 +12,14 @@ export function initializeCollectionsRoutes(
...
   client: SmartContentClient,
   theGraphClient: TheGraphClient,
-  offChainManager: OffChainWearablesManager
+  offChainManager: OffChainWearablesManager,
+  rootStorageLocation: string
 ): Router {
   router.get('/standard/erc721/:chainId/:contract/:option/:emission?', createHandler(client, getStandardErc721))
-  router.get('/contents/:urn/image', createHandler(client, contentsImage))
-  router.get('/contents/:urn/thumbnail', createHandler(client, contentsThumbnail))
+  router.get('/contents/:urn/image', createHandler(client, contentsImage, rootStorageLocation))
```
agusaldasoro, line 19: Do we need to update the Swagger documentation of the lambdas API?

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -150,46 +200,96 @@ function buildUrn(protocol: string, contract: string, option: string): string {
...
+}
+
+async function saveImage(client: SmartContentClient, rootStorageLocation: string, imageRequest: ImageRequest) {
+  const image = await client.downloadContent(imageRequest.hash)
+  const imagePath = getImagePath(rootStorageLocation, imageRequest)
+  let finalImage: sharp.Sharp
+  if (imageRequest.rarity) {
+    const background = await getRarityBackground(imageRequest.rarity)
+    const shouldResize = sizes[imageRequest.size] !== DEFAULT_IMAGE_SIZE
+    const resizedBackground = shouldResize
```
agusaldasoro, line 270: Should we store the resized backgrounds too or is it not necessary?

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -150,46 +200,96 @@ function buildUrn(protocol: string, contract: string, option: string): string {
...
+
+  // Check if the image is already in the cache, otherwise build and store it
+  fs.existsSync(path) || (await buildImage(client, rootStorageLocation, imageRequest))
+
+  return await sharp(path).toBuffer()
+}
+
+async function buildImage(client: SmartContentClient, rootStorageLocation: string, imageRequest: ImageRequest) {
+  const imagePath = getImagePath(rootStorageLocation, imageRequest)
+  fs.existsSync(imagePath) || (await saveImage(client, rootStorageLocation, imageRequest))
```
agusaldasoro, line 260: Why are you checking if the file exists here again?

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -150,46 +200,96 @@ function buildUrn(protocol: string, contract: string, option: string): string {
...
+function isValidSize(size: string): size is ValidSize {
+  return sizes[size] !== undefined
+}
+
+function getImagePath(root: string, imageRequest: ImageRequest): string {
+  return root + `/` + getImageRequestId(imageRequest) + '.png'
+}
+
+function getImageRequestId({ urn, hash, size, rarity }: ImageRequest): string {
+  return `${urn}-${hash}-${size}` + (rarity ? `-${rarity}` : '')
```
agusaldasoro, line 242: Should we check to store only the latest hash per urn?

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -150,46 +200,96 @@ function buildUrn(protocol: string, contract: string, option: string): string {
...
+
+async function getImage(
+  client: SmartContentClient,
+  rootStorageLocation: string,
+  imageRequest: ImageRequest
+): Promise<Buffer> {
+  const path = getImagePath(rootStorageLocation, imageRequest)
+
+  // Check if the image is already in the cache, otherwise build and store it
+  fs.existsSync(path) || (await buildImage(client, rootStorageLocation, imageRequest))
```
agusaldasoro, line 253: Would it make sense to have a metric that checks the number of images built? (+ The space on disk that they are using)

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -11,92 +14,135 @@ import {
...
+  try {
+    validateSize(size)
 
-  await internalContents(client, res, urn, (wearableMetadata) => wearableMetadata.thumbnail)
+    const entity = await fetchEntity(client, urn)
+    const imageRequest = {
+      urn,
+      hash: getFileHash(entity, entity.metadata.thumbnail),
+      size,
+      rarity: isThumbnail ? undefined : entity.metadata?.rarity
```
agusaldasoro, line 128: nit: when it's a thumbnail it's not that the rarity is undefined but you are not using it, maybe we should add another field for that?

agusaldasoro (changes requested):
Nice work! Only two comments:
- Let's update the Swagger docs if necessary.
- Implement a garbage collection system or delete old when storing news hashes per urn: there's no need to store the files for more than one version as we should always be retrieving the latest change.

lambdas/src/apis/collections/controllers/collections.ts
guidota: yep, it can be a good idea

lambdas/src/apis/collections/controllers/collections.ts
guidota: yes, i should probably rename this property to something like `background`

lambdas/src/apis/collections/Utils.ts
```diff
@@ -78,7 +78,25 @@ export function createExternalContentUrl(
   return undefined
 }
 
-export function findHashForFile(entity: Entity, fileName: string | undefined) {
+export function createExternalLambdasUrl(urn: string): string {
+  const catalystUrl = baseLambdasServerUrl(process.env.CATALYST_URL) ?? 'http://localhost'
+  return `${catalystUrl}/lambdas/collections/contents/${urn}`
```
menduz, line 83:
```suggestion
  return `${catalystUrl}/lambdas/collections/contents/${urn}`
```

lambdas/src/apis/collections/Utils.ts
```diff
@@ -78,7 +78,25 @@ export function createExternalContentUrl(
...
+}
```
menduz, line 84: Is it really necessary to have absolute URLs here? This is usually a bad practice. Servers should not be aware of their domain, doing so prevents flexible configurations using reverse proxies.

lambdas/src/apis/collections/Utils.ts
```diff
@@ -78,7 +78,25 @@ export function createExternalContentUrl(
...
+
+function baseLambdasServerUrl(url?: string): string | undefined {
+  if (!url) return
+
+  let configAddress = url.toLowerCase()
+  if (!configAddress.startsWith('http')) {
+    configAddress = 'http://' + configAddress
+  }
```
menduz, line 92: why does this `if` exist? I think we would  benefit from break in case of  misconfiguration instead of trying to fix it

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -11,92 +14,139 @@ import {
...
-          ...bodyShapeTraits
-        ]
-      }
-      res.send(standardErc721)
-    } else {
-      res.status(404).send()
+        return bodyShapes
+      },
+      []
+    )
```
menduz, line 51: This `reduce` does not spark joy 
```ts
const bodyShapeTraits = 
  getBodyShapes(wearableMetadata.data.representations).reduce(
      (bodyShapes: ERC721StandardTrait[], bodyShape) => {
        bodyShapes.push({
          trait_type: 'Body Shape',
          value: bodyShape
        })

        return bodyShapes
      },
      []
    )
```

Try a map instead?
```ts
const bodyShapeTraits: ERC721StandardTrait[] = 
  getBodyShapes(wearableMetadata.data.representations).map(bodyShape => {
    trait_type: 'Body Shape',
    value: bodyShape
  })
```

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -11,92 +14,139 @@ import {
...
+
+    const tagTraits = wearableMetadata.data.tags.reduce((tags: ERC721StandardTrait[], tag) => {
+      tags.push({
+        trait_type: 'Tag',
+        value: tag
+      })
+
+      return tags
+    }, [])
```
menduz, line 60: same reducer

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -11,92 +14,139 @@ import {
...
+          value: wearableMetadata.data.category
+        },
+        ...tagTraits,
+        ...bodyShapeTraits
+      ]
     }
+    res.send(standardErc721)
   } catch (e) {
-    res.status(500).send(e.message)
+    res.status(e.statusCode ?? 500).send(e.message)
```
menduz, line 84:
```suggestion
    res.status(e.statusCode ?? 500)
    logger.error(e)
```

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -11,92 +14,139 @@ import {
...
+  }
+}
+
+function sendImage(res: Response, image: Buffer, imageRequest: ImageRequest) {
+  res.send(image)
+  res.writeHead(200, {
+    'Content-Type': 'image/png',
+    ETag: JSON.stringify(getImageRequestId(imageRequest)),
+    'Access-Control-Expose-Headers': '*',
+    'Cache-Control': 'public, max-age=31536000, immutable'
```
menduz, line 148: Add `s-maxage` to enable edge & cdn caching
```suggestion
    'Cache-Control': 'public, max-age=31536000, s-maxage=31536000, immutable'
```

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -150,46 +204,117 @@ function buildUrn(protocol: string, contract: string, option: string): string {
...
+}
+
+// Using this folder structure allow us to find and remove older versions of the same urn (entity)
+function getImageRequestId({ urn, hash, size, rarityBackground }: ImageRequest): string {
+  return `${urn}/${hash}/${rarityBackground ?? 'thumbnail'}-${size}`
+}
+
+// Delete all images that are not the latest version (same hash)
+async function pruneObsoleteImages(root: string, urn: string, hash: string) {
+  const existsFolder = fs.existsSync(root + `/images/` + urn + '/' + hash)
```
menduz, line 252: - are urn valid folder names?
- prevent using sync operations in servers, use promisified versions at all costs when possible to prevent blocking the single thread

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -150,46 +204,117 @@ function buildUrn(protocol: string, contract: string, option: string): string {
...
+
+async function getImage(
+  client: SmartContentClient,
+  rootStorageLocation: string,
+  imageRequest: ImageRequest
+): Promise<Buffer> {
+  const path = getImagePath(rootStorageLocation, imageRequest)
+
+  // Check if the image is already in the cache, otherwise build and store it
+  fs.existsSync(path) || (await saveImage(client, rootStorageLocation, imageRequest))
```
menduz, line 268: please do IF instead

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -150,46 +204,117 @@ function buildUrn(protocol: string, contract: string, option: string): string {
...
+async function storeImage(rootStoragePath: string, imageRequest: ImageRequest, finalImage: sharp.Sharp) {
+  const imagesFolder = rootStoragePath + `/images`
+  const urnFolder = imagesFolder + `/${imageRequest.urn}`
+  const hashFolder = urnFolder + `/${imageRequest.hash}`
+  const imagePath = getImagePath(rootStoragePath, imageRequest)
+
+  // ensure folder structure exists before write
+  fs.existsSync(imagesFolder) || (await fs.promises.mkdir(imagesFolder))
+  fs.existsSync(urnFolder) || (await fs.promises.mkdir(urnFolder))
+  fs.existsSync(hashFolder) || (await fs.promises.mkdir(hashFolder))
```
menduz, line 302: please do IF instead

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -150,46 +204,117 @@ function buildUrn(protocol: string, contract: string, option: string): string {
...
+async function saveImage(client: SmartContentClient, rootStorageLocation: string, imageRequest: ImageRequest) {
+  const image = await client.downloadContent(imageRequest.hash)
+  const shouldResize = sizes[imageRequest.size] !== DEFAULT_IMAGE_SIZE
+  let finalImage: sharp.Sharp
+  if (imageRequest.rarityBackground) {
+    const background = await getRarityBackground(imageRequest.rarityBackground, imageRequest.size)
+    const resizedImage = shouldResize
+      ? await sharp(image).resize(sizes[imageRequest.size], sizes[imageRequest.size]).toBuffer()
+      : image
+    finalImage = sharp(background).composite([{ input: resizedImage }])
```
menduz, line 286:
```suggestion
      ? await sharp(image).resize(sizes[imageRequest.size], sizes[imageRequest.size])
      : image
    finalImage = sharp(background).composite([{ input: resizedImage }])
```

Are those `.toBuffer()` avoidable? (here and inside getImage)

Asking because it may be faster to use stream based APIs if available.

menduz (changes requested):

lambdas/src/apis/collections/Utils.ts
guidota: ERC 721 standard metadata is dynamic and use this external URL to reference the image and thumbnail, I don't see any other option here (before this change we were using the external URL too, but for a content endpoint)

lambdas/src/apis/collections/Utils.ts
guidota: Just wanted to be consistent with how we handle the cataylst url, same behaviour as the `SmartContentClient`. 
We can remove the logic tho.

lambdas/src/apis/collections/Utils.ts
menduz: could this be used instead?

```ts
  return `https://${req.headers.get('Host')}/lambdas/collections/contents/${urn}`
```

lambdas/src/apis/collections/controllers/collections.ts
guidota: > are urn valid folder names?
 
 not sure, do you know a breaking case?
 
> prevent using sync operations in servers, use promisified versions at all costs when possible to prevent blocking the single thread

there is no promisifed version for this method (exists) 😞

lambdas/src/apis/collections/controllers/collections.ts
guidota: will check another options!

guidota: Sharp API doesn't provide streams, I guess we can stream from the file once it's stored and avoid reading it again.

https://sharp.pixelplumbing.com/api-output

guidota: in this case `input` receives a `string` (path) or `Buffer`, will check if we can avoid buffer conversions.

lambdas/src/apis/collections/Utils.ts
guidota: yes, excellent 👍

lambdas/src/apis/collections/Utils.ts
agusaldasoro: If the `configAddress` is read from the `.env` file, then I think that it's okay to support both cases. It's not a case of misconfiguration of the leading `/`, this is more like a sanitization.

lambdas/src/apis/collections/controllers/collections.ts
menduz: Dang. We are not still there yet.. 

I've found this on stackoverflow but never used it
```ts
function checkFileExists(file) {
  return fs.promises.access(file, fs.constants.F_OK)
           .then(() => true)
           .catch(() => false)
}
```

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -81,7 +78,10 @@ export async function getStandardErc721(client: SmartContentClient, req: Request
     }
     res.send(standardErc721)
   } catch (e) {
-    res.status(e.statusCode ?? 500).send(e.message)
+    if (e instanceof ServiceError) {
+      res.status(e.statusCode).send(e.message)
+    }
+    res.status(500)
```
menduz, line 84: Prevent double .status call
```suggestion
    else {
      res.status(500)
    }
```

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -81,7 +78,10 @@ export async function getStandardErc721(client: SmartContentClient, req: Request
     }
     res.send(standardErc721)
   } catch (e) {
-    res.status(e.statusCode ?? 500).send(e.message)
+    if (e instanceof ServiceError) {
+      res.status(e.statusCode).send(e.message)
+    }
```
menduz, line 83:
```suggestion
    if (e instanceof ServiceError) {
      res.status(e.statusCode).send(e.message)
      return
    }
```

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -261,20 +267,34 @@ async function getImage(
...
   rootStorageLocation: string,
   imageRequest: ImageRequest
-): Promise<Buffer> {
+): Promise<[NodeJS.ReadableStream, number]> {
   const path = getImagePath(rootStorageLocation, imageRequest)
 
   // Check if the image is already in the cache, otherwise build and store it
-  fs.existsSync(path) || (await saveImage(client, rootStorageLocation, imageRequest))
+  if (fs.existsSync(path)) {
+    metricsComponent.increment('cache_hit')
```
menduz, line 275: ❤️

menduz: I think you are missing the `image_size` label as defined in the `metrics.ts`

menduz (approved):

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -150,46 +211,132 @@ function buildUrn(protocol: string, contract: string, option: string): string {
...
+): Promise<[NodeJS.ReadableStream, number]> {
+  const path = getImagePath(rootStorageLocation, imageRequest)
+
+  // Check if the image is already in the cache, otherwise build and store it
+  if (await checkFileExists(path)) {
+    metricsComponent.increment('cache_hit', { image_size: imageRequest.size })
+  } else {
+    metricsComponent.increment('cache_miss', { image_size: imageRequest.size })
+    await saveImage(client, rootStorageLocation, imageRequest)
+
```
agusaldasoro, line 280: what do you think of adding the prune folder here? So we don't have to check twice if the folder exists

lambdas/src/apis/collections/routes.ts
```diff
@@ -12,11 +12,14 @@ export function initializeCollectionsRoutes(
...
+  router.get('/contents/:urn/image/:size', createHandler(client, contentsImage, rootStorageLocation))
+  router.get('/contents/:urn/thumbnail', createHandler(client, contentsThumbnail, rootStorageLocation))
```
agusaldasoro, line 21: Are routes `'/contents/:urn/image'` and `'/contents/:urn/thumbnail'` implemented?

lambdas/src/apis/collections/controllers/collections.ts
```diff
@@ -11,92 +17,143 @@ import {
...
+  res: Response,
+  storage: string
+): Promise<void> {
   // Method: GET
-  // Path: /contents/:urn/thumbnail
+  // Path: /contents/:urn/image/:size
+  await handleImageRequest(client, req, res, storage, false)
+}
+
+async function handleImageRequest(
```
agusaldasoro, line 111: Should we move the logic for generating and storing images to another file?

agusaldasoro (approved):
Awesome work!!

guidota:
⚠️ Destroy streams on response finished

lambdas/src/apis/collections/controllers/collections.ts
guidota: I think is not the case here, as we are checking for image file and not hash folder. It should be possible that no prune should be done at this point.

lambdas/src/apis/collections/routes.ts
guidota: yes, same function is used both for `'/contents/:urn/image'` and `'/contents/:urn/image/:size'` 

(same for thumbnail)

lambdas/src/apis/collections/controllers/collections.ts
guidota: yes, makes sense to move image cache logic to other file 👍

Issue #620, opened 2021-08-20
Support v4 entities deployment


agusaldasoro:
- Add support to deploy (and validate) v4 entities.
- Migrate all Content Server clients to use v4 entities format
- Configure in code a date as a hard deadline to stop supporting deployments of v3 entities

Issue #618, opened 2021-08-20
Decide new size limit for v4 Deployments


agusaldasoro:
Currently, we have validation for the max size per deployment for each different type of entity. We need to:
- Understand which is the average size
- Decide a limit per entity type 
- Configure that limit for v4 entities.

PR #738, opened 2021-10-25
feat: add wearable thumbnail validations

guidota:
## Description

Adds thumbnail image validations for v4 wearable entities:
- Checks that file exists
- Checks size (width and height) is 1024x1024

Also:
- Increase total size limit for wearables to 3 MB.
- Set model size limit for wearables to 2MB (without taking into account the thumbnail image).

Related to #618 and #620


content/src/service/validations/Validations.ts
```diff
@@ -261,4 +263,28 @@ export class Validations {
...
+    if (!hash) return [`Couldn't find hash for thumbnail file with name: ${metadata.thumbnail}`]
+
+    // check size
+    const thumbnailBuffer = deployment.files.get(hash)
+    if (!thumbnailBuffer) return [`Couldn't find thumbnail file with hash: ${hash}`]
+    try {
+      const { width, height, format } = await sharp(thumbnailBuffer).metadata()
+      if (!width || !height) return [`Couldn't validate thumbnail size for file ${metadata.thumbnail}`]
+      if (!format || format !== 'png') return [`Invalid or unknown image format. Only 'PNG' format is accepted.`]
+      if (width !== DEFAULT_THUMBNAIL_SIZE || height !== DEFAULT_THUMBNAIL_SIZE)
```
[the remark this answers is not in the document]
guidota: Double checked ✅

content/src/service/validations/Validations.ts
```diff
@@ -35,6 +38,7 @@ export class Validations {
         }. You can upload up to ${entity.pointers.length * maxSizeInBytes} bytes but you tried to upload ${totalSize}.`
       ]
     }
+    return this.WEARABLE_SIZE(args)
```
agusaldasoro, line 41: mm, we are using return undefined as when no error is present, changing this will make it incompatible with existing validations

content/src/service/validations/Validations.ts
```diff
@@ -261,4 +262,84 @@ export class Validations {
...
+        const contentSize = await externalCalls.fetchContentFileSize(hash)
+        if (!contentSize) throw new Error(`Couldn't fetch content file with hash: ${hash}`)
+        totalSize += contentSize
+      }
+    }
+    return totalSize
+  }
+
+  /** Validate that given wearable deployment includes a thumbnail with valid format and size */
+  static readonly WEARABLE_CUSTOM: Validation = async (args) => {
```
agusaldasoro, line 285: Let's leave this validation as the only one, and make `WEARABLE_THUMBNAIL` and `WEARABLE_SIZE` private methods. Also, let's change the name `WEARABLE_CUSTOM` to something more explicit, for example, `WEARABLE_FILES`

content/src/service/validations/Validations.ts
```diff
@@ -261,4 +262,84 @@ export class Validations {
...
+    if (args.deployment.entity.type !== EntityType.WEARABLE) return
+
+    let errors: string[] = []
+    errors = [...((await this.WEARABLE_THUMBNAIL(args)) ?? []), ...((await this.WEARABLE_SIZE(args)) ?? [])]
+    return errors.length > 0 ? errors : undefined
+  }
+
+  static readonly WEARABLE_THUMBNAIL: Validation = async ({ deployment }) => {
```
agusaldasoro, line 293: Lets put here a comment indicating what this method does

content/src/service/validations/Validations.ts
```diff
@@ -261,4 +262,84 @@ export class Validations {
...
+  }
+
+  /** Validate wearable files size, excluding thumbnail, is less than expected */
+  static readonly WEARABLE_SIZE: Validation = async ({ deployment, env, externalCalls }) => {
+    const entity = deployment.entity
+    const maxSizeInMB = env.maxUploadSizePerTypeInMB.get(EntityType.WEARABLE)
+    if (!maxSizeInMB) return
+
+    // hardcoded value to be used just in this validation
+    const modelSizeInMB = 2
```
agusaldasoro, line 325: Shouldn't this be an environment global config as the sizes per entity?

content/src/service/validations/ValidatorFactory.ts
```diff
@@ -11,7 +11,7 @@ export class ValidatorFactory {
       maxUploadSizePerTypeInMB: new Map([
         [EntityType.SCENE, 15],
         [EntityType.PROFILE, 15],
-        [EntityType.WEARABLE, 2]
+        [EntityType.WEARABLE, 3]
```
agusaldasoro, line 14: Shouldn't we modify some configuration on the environment file too?

agusaldasoro (changes requested):

content/src/service/validations/Validations.ts
guidota: `WEARABLE_SIZE` also returns `undefined` as when no error is present, I guess there's no problem with that 🤔

content/src/service/validations/Validations.ts
guidota: will change the name!

other methods are not private since I added some tests for them and need to be visible 😞

content/src/service/validations/ValidatorFactory.ts
guidota: This configuration is just used by the `ValidatorFactory` and there is no file for that

content/src/service/validations/Validations.ts
agusaldasoro: I think that this method should return undefined too

agusaldasoro (approved):
Great work!

PR #751, opened 2021-11-08
fix: throw error when the graph queries fail

### LINKED ISSUE #749: The Graph failing queries are being logged but ignored
## Description

When a query to The Graph fails, it logs the error and returns a default value. This is causing valid responses (code 200) while some errors are not exposed, for example: 

Screen Shot 2021-11-05 at 15 45 54

We should find a way to measure this failing queries or expose them correctly.

guidota:
Fixes #749

### DIFF
--- a/lambdas/src/apis/collections/controllers/wearables.ts
+++ b/lambdas/src/apis/collections/controllers/wearables.ts
@@ -113,7 +113,7 @@ export async function getWearablesEndpoint(
 
     res.send({ wearables, filters: requestFilters, pagination: { limit: sanitizedLimit, lastId, next } })
   } catch (error) {
-    res.status(error.code ?? 500).send(error.message)
+    res.status(500).send(error.message)
   }
 }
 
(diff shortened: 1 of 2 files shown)


lambdas/src/apis/collections/controllers/wearables.ts
```diff
@@ -113,7 +113,7 @@ export async function getWearablesEndpoint(
 
     res.send({ wearables, filters: requestFilters, pagination: { limit: sanitizedLimit, lastId, next } })
   } catch (error) {
-    res.status(500).send(error.message)
+    res.status(error.code ?? 500).send(error.message)
```
menduz, line 116: As a rule of thumb, in any service, we should _never_ forward the errors to the clients. Only logging them to the logger. https://owasp.org/www-community/Improper_Error_Handling

menduz (changes requested):

agusaldasoro (approved):

lambdas/src/apis/collections/controllers/wearables.ts
menduz: Edit: I was suggesting to remove the `.send(error.message)`

menduz (approved):

Issue #743, opened 2021-10-29
AccessChecker for third party wearables


agusaldasoro:
To implement https://github.com/decentraland/adr/blob/main/docs/ADR-42-third-party-collections-registry.md#catalyst-acceptance-criteria we need to modify the access checker for wearables.

We need to decide if this change will affect all entities, or only v4 and above.

According to the ADR, the check fulfills:

When deploying a third-party wearable

1. Exists a record with the urn: `urn:decentraland:matic:ext-thirdparty:0x1234:tokenId1`.

2. The content hash of the item with id `0x1234:tokenId1` is the same as the content hash of the item that is being uploaded

Check 2 is already done, we need to add the check for 1.

For that, we need to query the Graph with the following query: https://github.com/decentraland/tpr-graph/tree/feat/add-contenthash#get-an-item-by-urn--contenthash

PR #761, opened 2021-11-16
feat: check items content hash against tpr subgraph

guidota:
## Description

Upgrades `urn-resolver` version and checks if a Third Party wearable exists in the Third Party Registry by querying to the subgraph.

Related to #743 

To do:
- [x] Add tests


content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -15,62 +15,114 @@ export class AccessCheckerForWearables {
...
+
+    if (isThirdparty) {
+      const existsItem = await this.checkItemExistence(
+        pointer,
+        this.thirdPartySubgraphUrl,
+        this.blocksL2SubgraphUrl,
+        accessParams
+      )
+      if (!existsItem) return [`The third-party item ${pointer} does not exist.`]
+      // @todo check if the third-party item is owne by the user
```
agusaldasoro, line 48: Please validate with dApps team, but I believe that the below validation needs to be done whether or not it's a third-party wearable, the only difference is the subgraph URL.

content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -15,62 +15,114 @@ export class AccessCheckerForWearables {
...
+  private async existsItem(
+    subgraphUrl: string,
+    urn: string,
+    block: number,
+    content?: Map<string, string>,
+    metadata?: any
+  ): Promise<boolean> {
+    const query = `
+    query ($urn: String!, $block: Int!) {
+      items(where:{ urn: $urn}, block: { number: $block }) {
```
agusaldasoro, line 111: should the `contentHash` be in the where clause?

agusaldasoro: Also, are we considering a 5 minutes window for checking permissions as we are doing on other wearables collections?

content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -100,6 +152,19 @@ export class AccessCheckerForWearables {
...
+    // Compare both by key and hash
+    const compare = (a: { key: string; hash: string }, b: { key: string; hash: string }) => {
+      if (a.hash > b.hash) return 1
+      else if (a.hash < b.hash) return -1
+      else return a.key > b.key ? 1 : -1
+    }
+    const entries = Array.from(content?.entries() ?? [])
+    const contentAsJson = entries.map(([key, hash]) => ({ key, hash })).sort(compare)
+    const buffer = Buffer.from(JSON.stringify({ content: contentAsJson, metadata }))
+    return Promise.all([Hashing.calculateBufferHash(buffer), Hashing.calculateIPFSHash(buffer)])
```
agusaldasoro, line 165: @menduz if we are changing the `calculateIPFSHash` here, how will it impact the contract?

content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -15,62 +15,114 @@ export class AccessCheckerForWearables {
     private readonly collectionsL2SubgraphUrl: string,
     private readonly blocksL1SubgraphUrl: string,
     private readonly blocksL2SubgraphUrl: string,
+    private readonly thirdPartySubgraphUrl: string,
     private readonly LOGGER: log4js.Logger
   ) {}
 
   public async checkAccess({ pointers, ...accessParams }: WearablesAccessParams): Promise<string[]> {
-    const errors: string[] = []
+    if (pointers.length !== 1) return [`Only one pointer is allowed when you create a Wearable. Received: ${pointers}`]
```
menduz, line 23: `Received: ${pointers}` would render `Received: [Array object]`

content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -100,6 +152,19 @@ export class AccessCheckerForWearables {
...
+      if (a.hash > b.hash) return 1
+      else if (a.hash < b.hash) return -1
+      else return a.key > b.key ? 1 : -1
+    }
+    const entries = Array.from(content?.entries() ?? [])
+    const contentAsJson = entries.map(([key, hash]) => ({ key, hash })).sort(compare)
+    const buffer = Buffer.from(JSON.stringify({ content: contentAsJson, metadata }))
+    return Promise.all([Hashing.calculateBufferHash(buffer), Hashing.calculateIPFSHash(buffer)])
+  }
+
```
menduz, line 167: I believe this function should be shared in `catalyst-common`, otherwise the implementation of the hash may vary from what is actually being calculated from dApps. We could remove the code from here in another PR, I will send the PR to commons now: https://github.com/decentraland/catalyst-commons/pull/127

content/src/service/access/AccessCheckerForWearables.ts
menduz: I think we should only use the new hashes. Here is a pertinent PR https://github.com/decentraland/catalyst-commons/pull/127

menduz (changes requested):

agusaldasoro (approved):

content/src/service/access/AccessCheckerForWearables.ts
nachomazzara, line 48: Not sure I understood this comment. TPI can't be owned by anyone. They are just representations

guidota: yes, you're right... is not part of this flow

content/src/service/access/AccessCheckerForWearables.ts
```diff
@@ -15,62 +15,114 @@ export class AccessCheckerForWearables {
...
+  ): Promise<boolean> {
+    const query = `
+    query ($urn: String!, $block: Int!) {
+      items(where:{ urn: $urn}, block: { number: $block }) {
+        id,
+        contentHash
+      }
+    }
+    `
+    const hashes = await this.calculateHashes(content, metadata)
```
nachomazzara, line 117: are we moving the `calculateHashes` to the Catalyst client or Catalyst common?

menduz: probably makes sense to do it in a following PR, Agustina is working on the type adaptations for the new version of catalyst-commons

content/src/service/access/AccessCheckerForWearables.ts
menduz: OMG I didn't know

content/src/service/access/AccessCheckerForWearables.ts
guidota: yes, probably doing in another PR since we need to catch up with some breaking changes in `catalyst-commons`

PR #771, opened 2021-11-19
add new snapshots endpoint

agusaldasoro:
This PR implements the necessary changes to support the new endpoint `/snapshot` which will replace the currently existing `/snapshot/{entityType}`. Everything is documented at https://github.com/decentraland/catalyst-api-specs/pull/15

These snapshots will be automatically stored in memory every 15min (this frequency can be changed by each Catalyst owner). This new endpoint will be used among `/pointer-changes` for sync and bootstrapping purposes.

## Pending tasks

- [ ] Fix integration tests
- [x] Add tests for the new snapshots


content/src/repository/extensions/DeploymentsRepository.ts
```diff
@@ -176,7 +185,24 @@ export class DeploymentsRepository {
     return `(${equalWithEntityIdComparison} OR ${timestampComparison})`
   }
 
-  getSnapshot(
+  getFullSnapshot(): Promise<FullSnapshot[]> {
+    return this.db.map(
+      `SELECT entity_id, entity_type, entity_pointers, auth_chain, date_part('epoch', local_timestamp) * 1000 AS local_timestamp ` +
+        `FROM deployments ` +
+        `WHERE deleter_deployment IS NULL ` +
+        `ORDER BY local_timestamp DESC, LOWER(entity_id) DESC`,
```
menduz, line 193:
```suggestion
        `ORDER BY local_timestamp ASC, LOWER(entity_id) DESC`,
```

content/src/service/snapshots/SnapshotManager.ts
```diff
@@ -103,6 +134,51 @@ export class SnapshotManager {
...
+      (db) =>
+        db.txIf(async (transaction) => {
+          // Get all the active entities
+          const snapshot: FullSnapshot[] = await transaction.deployments.getFullSnapshot()
+
+          // Calculate the local deployment timestamp of the newest entity in the snapshot
+          const snapshotTimestamp = snapshot[0]?.localTimestamp ?? 0
+
+          // Format the snapshot in a buffer
+          const buffer = Buffer.from(snapshot.join('\n'))
```
menduz, line 152:
```suggestion
          const tmpFile = resolve(contentFolder, 'tmp-snapshot' + Math.random())
          const writeStream = fs.createWriteStream(tmpFile)
          try {
            for (const snapshotElem of snapshot) {
              writeStream.write(JSON.stringify(snapshotElem) + '\n')
            }
          } finally {
            writeStream.close()
          }
          
          // TODO: use require('@dcl/snapshots-fetcher/dist/utils').hashStreamV1 after 
          //       https://github.com/decentraland/snapshots-fetcher/pull/4/files is merged
          const hash = await Hashing.calculateIPFSHash(
            await fs.promises.readFile(tmpFile)
          )
          
          // set the correct name
          await fs.promises.rename(tmpFile, resolve(content, hash))
```

menduz (changes requested):
Would you add some tests? The tests for the entityType-based snapshots look reusable for this case

content/src/repository/extensions/DeploymentsRepository.ts
agusaldasoro: Okay, but shouldn't we start with latest deployments so we can ignore the old ones assuming that we already processed them?

content/package.json
```diff
@@ -16,6 +18,7 @@
     "@catalyst/contracts": "0.1.0",
     "@dcl/catalyst-api-specs": "1.0.1",
     "@dcl/schemas": "1.8.0",
+    "@dcl/snapshots-fetcher": "^1.0.0-20211124191324.commit-0ce02d3",
```
agusaldasoro, line 21: TODO: Use a releasable version

content/src/service/snapshots/SnapshotManager.ts
```diff
@@ -35,18 +48,28 @@ export class SnapshotManager {
...
       },
       { priority: DB_REQUEST_PRIORITY.HIGH }
     )
   }
 
-  getSnapshotMetadata(entityType: EntityType): SnapshotMetadata | undefined {
+  async startCalculateFullSnapshots(): Promise<void> {
+    // TODO: Add metrics regarding snapshots
+    await this.generateSnapshot()
+    setTimeout(async () => await this.startCalculateFullSnapshots(), this.snapshotFrequencyInMilliSeconds)
```
menduz, line 62: Use setInterval instead of setTimeout. Mind that `setInterval` does not accept async functions. You must catch the exception like this:

```diff
-setTimeout(async () => await this.startCalculateFullSnapshots(), this.snapshotFrequencyInMilliSeconds)
+setInterval(() => this.startCalculateFullSnapshots().catch(SnapshotManager.LOGGER.error), this.snapshotFrequencyInMilliSeconds)
```

content/src/service/snapshots/SnapshotManager.ts
```diff
@@ -103,6 +126,90 @@ export class SnapshotManager {
...
+        },
+        { priority: DB_REQUEST_PRIORITY.HIGH }
+      )
+
+      // Delete the previous full snapshot (if it exists)
+      if (!!previousHash && this.shouldPrunePreviousSnapshot(previousHash)) {
+        await this.service.deleteContent([previousHash])
+      }
+    } catch {
+      SnapshotManager.LOGGER.debug('There was an error generating snapshot')
```
menduz, line 198:
```suggestion
    } catch (err: any) {
      SnapshotManager.LOGGER.debug('There was an error generating snapshot')
      SnapshotManager.LOGGER.error(err)
```

We will love to have those logs if something goes wrong

content/src/controller/Controller.ts
```diff
@@ -597,7 +624,10 @@ export class Controller {
     })
   }
 
-  getSnapshot(req: express.Request, res: express.Response) {
+  /**
+   * @deprecated
```
menduz, line 628: is it really deprecated? for how long?

menduz (changes requested):

content/src/controller/Controller.ts
agusaldasoro: Until we migrate all consumers to the new endpoint, I want to remove that behaviour as it adds a transaction per operation in the database.

menduz (approved):

content/src/metrics.ts
```diff
@@ -39,6 +39,12 @@ export const metrics = validateMetricsDeclaration({
     labelNames: []
   },
 
+  dcl_content_snapshot_generation_time: {
+    help: 'Histogram of time spent generating full snapshots',
+    type: 'histogram',
+    labelNames: ['failed']
```
agusaldasoro, line 45: should the label name be `failed` or something more generic?

content/src/service/snapshots/SnapshotManager.ts
```diff
@@ -140,76 +159,90 @@ export class SnapshotManager {
...
   /** This methods queries the database and builds the snapshots, stores it on the content storage, and saves the metadata */
-  private async generateSnapshot(task?: Database): Promise<void> {
+  private async generateSnapshot(): Promise<void> {
+    // Format the snapshot in a tmp file
+    const tmpFile = path.resolve(this.contentStorageFolder, 'tmp-snapshot-file')
+
+    const { end: stopTimer } = metricsComponent.startTimer('dcl_content_snapshot_generation_time')
+
     try {
+      if (await checkFileExists(tmpFile)) {
```
agusaldasoro, line 169: why do we need this? Can we add a comment here?

content/src/service/snapshots/SnapshotManager.ts
```diff
@@ -140,76 +159,90 @@ export class SnapshotManager {
   }
 
   /** This methods queries the database and builds the snapshots, stores it on the content storage, and saves the metadata */
-  private async generateSnapshot(task?: Database): Promise<void> {
+  private async generateSnapshot(): Promise<void> {
+    // Format the snapshot in a tmp file
+    const tmpFile = path.resolve(this.contentStorageFolder, 'tmp-snapshot-file')
```
agusaldasoro, line 164: can you ensure that you will only be processing one file each time?

content/src/metrics.ts
menduz: labels = { failed: "true" | "false" }

If you think a better name please go for it

content/src/service/snapshots/SnapshotManager.ts
menduz:
```suggestion
      // if the process failed while creating the snapshot last time the file may still exists
      // deleting the staging tmpFile just in case
      if (await checkFileExists(tmpFile)) {
```

content/src/service/snapshots/SnapshotManager.ts
menduz: Yes, it can be ensured because `generateSnapshot` is called from only one place inside an awaitable loop

content/src/service/snapshots/SnapshotManager.ts
```diff
@@ -103,6 +184,135 @@ export class SnapshotManager {
...
+      )
+      await compressContentFile(destinationFilename)
+    } else {
+      SnapshotManager.LOGGER.debug(
+        `Snapshot didn't change. hash=${options.hash} lastIncludedDeploymentTimestamp=${options.snapshotTimestamp} elements=${options.elements} timeElapsed=${options.timeElapsed}`
+      )
+    }
+  }
+
+  private async writeToFile(tmpFile: string, entityType: EntityType | ALL_ENTITIES) {
```
guidota, line 269: I see `writeToFile` a bit confusing and it seems to be an important method, what do you think about renaming it to something like `writeDeploymentsToFile`?

content/src/storage/FileSystemContentStorage.ts
```diff
@@ -51,7 +70,7 @@ export class FileSystemContentStorage implements ContentStorage {
 
   async exist(ids: string[]): Promise<Map<string, boolean>> {
     const checks = await Promise.all(
-      ids.map<Promise<[string, boolean]>>(async (id) => [id, await existPath(this.getFilePath(id))])
+      ids.map<Promise<[string, boolean]>>(async (id) => [id, !!(await this.retrieve(id))])
```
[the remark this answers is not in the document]
menduz: It is necessary because the exists also checks about `.gzip` postfixes

jmoguilevsky (approved):

PR #780, opened 2021-11-24
feat: fetch third party integrations from subgraph endpoint

guidota:
## Description

Adds `/third-party-integrations` endpoint.

It will query The Graph third party registry subgraph and return the list of integrations.

### DIFF
--- a/lambdas/src/Server.ts
+++ b/lambdas/src/Server.ts
@@ -1,5 +1,5 @@
-import { initializeMetricsServer } from '@dcl/catalyst-node-commons'
 import { LAMBDAS_API } from '@dcl/catalyst-api-specs'
+import { initializeMetricsServer } from '@dcl/catalyst-node-commons'
 import compression from 'compression'
 import cors from 'cors'
 import express from 'express'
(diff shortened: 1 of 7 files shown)


lambdas/src/Server.ts
```diff
@@ -120,6 +121,8 @@ export class Server {
 
     // Functionality for Explore use case
     this.app.use('/explore', initializeExploreRoutes(express.Router(), env.getBean(Bean.DAO), contentClient))
+
+    this.app.use('/third-party-integrations', initializeThirdPartyIntegrationsRoutes(theGraphClient, express.Router()))
```
agusaldasoro, line 125: Should we update the docs?

lambdas/src/apis/third-party/controllers/third-party.ts
```diff
@@ -0,0 +1,22 @@
+import { Response } from 'express'
+import { TheGraphClient } from '../../../utils/TheGraphClient'
+import { TimeRefreshedDataHolder } from '../../../utils/TimeRefreshedDataHolder'
+import { ThirdPartyIntegration } from '../../collections/types'
+
+let thirdPartyIntegrationsCache: TimeRefreshedDataHolder<ThirdPartyIntegration[]>
+
+export function initCache(theGraphClient: TheGraphClient): void {
+  thirdPartyIntegrationsCache = new TimeRefreshedDataHolder(() => fetchThirdPartyIntegrations(theGraphClient), '1m')
```
agusaldasoro, line 9: Will we querying the graph every one minute? Is this really necessary? Also, can this be configurable?

guidota: I think it's aligned with block times and graph indexing (more less). It is not actively updating this data every 1m, just holding it as caching strategy.

We can add timer configuration for sure.

agusaldasoro: Okay, can we increase that default interval? Or do you think that it's a risk?

lambdas/src/Server.ts
menduz: Can we have a `@beta` channel in the docs? I'm afraid that people consider unstable APIs as part of the protocol and then those should be working forever after that

agusaldasoro (approved):

PR #781, opened 2021-11-24
Bootstrap and sync with snapshots and pointer changes

menduz:
TODO:

- [x] Show the last sync with each other catalyst in /status endpoint
- [x] Show last timestamp that synced in /status endpoint
- [x] What happens with failed deployments now? Both during sync of snapshots and pointer changes
- [ ] Define denylist behavior: When an entity is deny-listed across all nodes, then no entity is deployed?
- [x] Old failed deployments will have to be fixed manually?
- [x] Show #activeEntities instead of #history
- [x] Use bootstrapping/sync/syncing status to avoid serve traffic while bootstrapping
- [x] Use different metrics/tags to show organic deployments vs sync deployments
- [ ] Create migration to delete all failed_deployments `WHERE auth_chain IS NULL` and make column nonNull right after
- [ ] delete dcl_sync_state_summary metric


jmoguilevsky (approved):
LGTM

PR #784, opened 2021-11-26
feat: fetch wearables from third party resolver API

guidota:
## Description

Adding TPW support to `GET /lambdas/collections/wearables-by-owner/:address?collectionId=THIRD_PARTY_URN` endpoint. 

Added behaviour: 
- When a `collectionId` is present.
- Query The Graph for a TPR (third party resolver).
- Fetch wearables from that TPR API (using same `collectionId` and given user `address`)

### DIFF
--- a/lambdas/src/apis/collections/controllers/wearables.ts
+++ b/lambdas/src/apis/collections/controllers/wearables.ts
@@ -2,9 +2,11 @@ import { toQueryParams } from '@dcl/catalyst-node-commons'
 import { EntityType } from 'dcl-catalyst-commons'
 import { EthAddress } from 'dcl-crypto'
 import { Request, Response } from 'express'
+import log4js from 'log4js'
 import { asArray, asInt } from '../../../utils/ControllerUtils'
 import { SmartContentClient } from '../../../utils/SmartContentClient'
 import { TheGraphClient } from '../../../utils/TheGraphClient'
+import { createThirdPartyFetcher, createThirdPartyResolver } from '../../../utils/third-party'
 import { BASE_AVATARS_COLLECTION_ID, OffChainWearablesManager } from '../off-chain/OffChainWearablesManager'
 import { Wearable, WearableId, WearablesFilters, WearablesPagination } from '../types'
 import { isBaseAvatar, translateEntityIntoWearable } from '../Utils'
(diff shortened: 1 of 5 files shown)


lambdas/src/apis/collections/controllers/wearables.ts
```diff
@@ -22,29 +23,41 @@ export async function getWearablesByOwnerEndpoint(
   theGraphClient: TheGraphClient,
   req: Request,
   res: Response
-) {
+): Promise<void> {
   // Method: GET
-  // Path: /wearables-by-owner/:owner
+  // Path: /wearables-by-owner/:owner?collectionId={string}
```
agusaldasoro, line 28: Should we update the docs? Also, can you add a comment here explaining the behaviour when no param is sent

lambdas/src/apis/collections/controllers/wearables.ts
```diff
@@ -72,7 +85,7 @@ export async function getWearablesEndpoint(
   offChainManager: OffChainWearablesManager,
   req: Request,
   res: Response
-) {
+): Promise<unknown> {
```
agusaldasoro, line 88: Shouldn't this be something like: `{wearables: Wearable[], filters: Filters, pagination: Pagination}`?

lambdas/src/utils/TheGraphClient.ts
```diff
@@ -176,7 +176,17 @@ export class TheGraphClient {
       mapper: (response) =>
         response.thirdParties.map((tp) => ({ urn: tp.id, description: tp.metadata.thirdParty.description }))
     }
-    return this.runQuery(query, { thirdPartyType: 'third_party_v1' })
+    return this.paginatableQuery(query, { thirdPartyType: 'third_party_v1' })
+  }
+
+  public async findThirdPartyResolver(subgraph: keyof URLs, id: string): Promise<string | undefined> {
```
agusaldasoro, line 182: can we add a comment here explaining what this method does and why we need it?

lambdas/src/apis/collections/controllers/wearables.ts
guidota: In this case that's what the response will send, but not what the method returns.

agusaldasoro: what about `any` then?

guidota: it doesn't work when result is void (no return)

guidota:
Consider fetching strategies since Third Party resolver endpoint is paginated and `/wearables-by-owner` is not.
We probably need to define if we want to fetch all pages and return complete result or add pagination as well for `/wearables-by-owner` (it may have impact in other services).

agusaldasoro (approved):

PR #801, opened 2021-12-17
New status endpoint

jmoguilevsky:
New status removes history and has a lastupdated time for the snapshot and the amount of entities of each type in the snapshot


content/test/unit/service/Service.spec.ts
```diff
@@ -249,7 +252,13 @@ describe('Service', function () {
       .registerBean(Bean.DEPLOYMENT_MANAGER, NoOpDeploymentManager.build())
       .registerBean(Bean.REPOSITORY, MockedRepository.build(new Map([[EntityType.SCENE, initialAmountOfDeployments]])))
       .registerBean(Bean.CACHE_MANAGER, new CacheManager())
-    return ServiceFactory.create(env)
+
+    const components: AppComponents = {
+      status: createStatusComponent(),
+      logs: createLogComponent()
+    } as any
```
menduz, line 259:
```suggestion
    const components: AppComponents = {
      status: createStatusComponent(),
      logs: createLogComponent(),
      metrics: metricsComponent,
    }
```

content/src/service/snapshots/SnapshotManager.ts
```diff
@@ -166,11 +183,20 @@ export class SnapshotManager {
 
         // add the entoty to the inMemoryArray to be used by the legacy formatter
         appendToInMemoryArray(snapshotElem.entityType as EntityType, [snapshotElem.entityId, snapshotElem.pointers])
+
+        if (newActiveEntitiesCount[snapshotElem.entityType] != null) {
+          newActiveEntitiesCount[snapshotElem.entityType]++
+        } else {
+          newActiveEntitiesCount[snapshotElem.entityType] = 1
+        }
```
menduz, line 191: This is duplicated logic from the line before it

```ts
appendToInMemoryArray(snapshotElem.entityType as EntityType, [snapshotElem.entityId, snapshotElem.pointers])
```

After this for-each it can be inferred from the in-memory-arrays

```ts
   const inMemoryArrays: Partial<Record<EntityType, Array<[string, string[]]>>> = {}
```

@jmoguilevsky

PR #802, opened 2021-12-17
chore: use correct version based on latest release

menduz:


jmoguilevsky (approved):

.circleci/config.yml
```diff
@@ -146,7 +146,8 @@ jobs:
             if echo "${CIRCLE_TAG}" | grep "^[0-9]\+\.[0-9]\+\.[0-9]\+$"; then
               DOCKER_BUILDKIT=1 docker build -t ${DOCKER_IMAGE_NAME}:${CIRCLE_SHA1} --build-arg COMMIT_HASH=${CIRCLE_SHA1} --build-arg CATALYST_VERSION=${CIRCLE_TAG} .
             else
-              DOCKER_BUILDKIT=1 docker build -t ${DOCKER_IMAGE_NAME}:${CIRCLE_SHA1} --build-arg COMMIT_HASH=${CIRCLE_SHA1} .
+              CATALYST_VERSION="$(curl https://api.github.com/repos/decentraland/catalyst/releases/latest -s | jq .name -r)-next"
```
agusaldasoro, line 149: aren't we in a GitHub repository when running this? If so, then we should use git commands.

Also, if the latest version is `2.5.0` then it happens that `2.5.0-next < 2.5.0` acording semver. We should increase the minor version before adding the `-next`

agusaldasoro (changes requested):
aren't we in a GitHub repository when running this? If so, then we should use git commands.

Also, if the latest version is 2.5.0 then it happens that 2.5.0-next < 2.5.0 acording semver. We should increase the minor version before adding the -next

guidota:
`master` branch deleted, reopen with new default branch (`main`)

PR #805, opened 2021-12-20
fix: filter approved wearables

guidota:


lambdas/src/utils/TheGraphClient.ts
```diff
@@ -163,13 +163,17 @@ export class TheGraphClient {
...
+      { id: WearableId; isApproved: boolean }[]
+    > = {
       description: 'fetch wearables by owner',
       subgraph: subgraph,
       query: QUERY_WEARABLES_BY_OWNER,
-      mapper: (response) => response.nfts.map(({ urn }) => urn)
+      mapper: (response) => response.nfts.map(({ urn, collection }) => ({ id: urn, isApproved: collection.isApproved }))
     }
-    return this.paginatableQuery(query, { owner: owner.toLowerCase() })
+    const wearables = await this.paginatableQuery(query, { owner: owner.toLowerCase() })
```
nachomazzara, line 175: should we need to filter before paginating?

guidota: `paginatableQuery` will always retrieve all wearables (approved or not) and we are not paginating the result, so it doesn't make sense for the time being

agusaldasoro (approved):

PR #808, opened 2021-12-21
fix: attempt to fix uploads

menduz:


agusaldasoro (approved):
I'm approving to unblock, but in the next release I'd like to refactor the metrics so we can have only one with different tags:
- Kind: Synced/Organic
- Result: Success/Failed
- ErrorType: DownloadFile/Timeout/Etc
- EntityType

menduz:
sounds great lets do it

PR #828, opened 2021-12-28
Remove support for deploying legacy entities

agusaldasoro:


content/src/service/access/AccessCheckerForScenes.ts
```diff
@@ -52,8 +52,8 @@ export class AccessCheckerForScenes {
                 if (!hasAccess) {
                   errors.push(`The provided Eth Address does not have access to the following parcel: (${x},${y})`)
                 }
-              } catch (e) {
-                errors.push(`The provided Eth Address does not have access to the following parcel: (${x},${y}). ${e}`)
+              } catch {
```
menduz, line 55: Prevent sending the error to /dev/null printing it to the logger.

menduz (approved):
Lovely PR. One single comment for the people doing ops of the running servers trying to figure out why something doesn't work with the graph

PR #847, opened 2022-01-10
test: replace Jasmine with Jest as the testing framework

juanscolari:
## Description

Replace `Jasmine` testing framework with `Jest` in order to improve the development experience, and allow features like global setup and teardown functionalities, as well as parallel processing.

## Changes

- Remove `Jasmine` and replace with `Jest`
- Use `ts-jest` library to support `Jest` with TypeScript
- Add option to silence logs to the console to improve readability
- Create global setup and teardown for Docker containers in integration tests


.vscode/extensions.json
```diff
@@ -1,7 +1,6 @@
 {
   "recommendations": [
     "dbaeumer.vscode-eslint",
-    "esbenp.prettier-vscode",
-    "hbenl.vscode-jasmine-test-adapter"
+    "esbenp.prettier-vscode"
```
guidota, line 4: can we add Jest extension as recommended?

lambdas/src/utils/TimeRefreshedDataHolder.ts
```diff
@@ -5,7 +5,10 @@ export class TimeRefreshedDataHolder<T> {
   private valuePromise: Promise<T>
   private lastUpdateCall: Date
 
-  constructor(private readonly provider: () => Promise<T>, private readonly refreshTime: string) {}
+  constructor(
+    private readonly provider: () => Promise<T>,
+    private readonly refreshTime: string // private isStarted: boolean = true
```
agusaldasoro, line 10: Did you forget to delete commented code here?

content/jest.globalSetup.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (!isCI()) {
+    global.__POSTGRES_CONTAINER__ = await new GenericContainer('postgres', '12')
+      .withName('postgres_test')
+      .withEnv('POSTGRES_PASSWORD', DEFAULT_DATABASE_CONFIG.password)
+      .withEnv('POSTGRES_USER', DEFAULT_DATABASE_CONFIG.user)
+      .withExposedPorts(E2ETestEnvironment.POSTGRES_PORT)
+      .withWaitStrategy(new PostgresWaitStrategy())
+      .start()
+  }
+}
```
guidota, line 18: 👏

lambdas/src/utils/TimeRefreshedDataHolder.ts
pedrotambo: Thxs

.vscode/extensions.json
juanscolari: Done 👍

content/jest.globalSetup.ts
```diff
@@ -0,0 +1,53 @@
+import { GenericContainer } from 'testcontainers'
+import { Container } from 'testcontainers/dist/container'
+import { LogWaitStrategy } from 'testcontainers/dist/wait-strategy'
+import { DEFAULT_DATABASE_CONFIG } from './src/Environment'
+import { E2ETestEnvironment } from './test/integration/E2ETestEnvironment'
+import { isCI } from './test/integration/E2ETestUtils'
+
+export default async (): Promise<void> => {
```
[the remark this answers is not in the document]
menduz: After reading the whole PR, I'd like to suggest this: 

Wouldn't it be better to be more in control and use a high order function to test with a DB? These hooked files add a lot of magic and force us to have to separate unit and integration tests. Making coverage reports _very_ hard to unify.

```ts
export function describeWithDb(nameOfSuite: string, testFn: () => void) {
  describe(nameOfSuite, () => {
    it('initialize db', () => ...)
    testFn()
    it('teardown db', () => ...)
  })
}
```

With that function, the only thing needed would be to use `describeWithDb` instead of `describe`


```ts
describeWithDb('my test', () => {
  it('does something with the DB', () => ...)
})
```

content/jest.setupFilesAfterEnv.ts
```diff
@@ -0,0 +1,18 @@
...
+    const loggerMock = logComponentMock.getLogger('__test__')
+    loggerMock.debug = () => {}
+    loggerMock.error = () => {}
+    loggerMock.info = () => {}
+    loggerMock.log = () => {}
+    loggerMock.warn = () => {}
+    logComponentMock.getLogger = jest.fn(() => loggerMock)
+    return logComponentMock
+  })
+  jest.spyOn(process.stdout, 'write').mockImplementation(() => true)
```
menduz, line 17: Why is this necessary also?

content/jest.globalSetup.ts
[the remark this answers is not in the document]
menduz: Otherwise I'd suggest to move it back to the `E2ETestEnvironment` to not scatter the logic around many files handled by runners, that makes the debugging and understanding of the tests flow hard to follow

content/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "../tsconfig.json",
-  "include": ["src", "test"],
+  "include": ["src", "test", "./jest.*.ts"],
   "compilerOptions": {
+    "types": ["node", "jest"],
```
menduz, line 5: Types for "jest" should only be included in the `content/test/tsconfig.json`
```suggestion
    "types": ["node"],
```

jest.config.json
```diff
@@ -0,0 +1,7 @@
+{
```
menduz, line 1: Is this file necessary? if so, why?

commons/jest.config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "preset": "ts-jest",
+  "silent": true,
+  "testEnvironment": "node",
+  "testMatch": ["**/*.spec.(ts)"],
+  "verbose": true
```
menduz, line 6: https://github.com/well-known-components/template-server/blob/main/jest.config.js#L11-L12

Please add coverage reports to all `jest.config` calls, the `--coverage` arg is needed to make it work.

```
  coverageDirectory: "coverage",
  collectCoverageFrom: ["src/**/*.ts", "src/**/*.js"],
```

menduz (changes requested):

content/jest.globalSetup.ts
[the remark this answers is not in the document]
guidota: This change only applies to development env, preventing `E2ETestEnvironment` to spin up and delete a container every test, but not for initializing the DB. 
DB initialization is still happening in `E2ETestEnvironment` where it creates a database and run the migrations.

Also there is no problem to unify unit and integration tests, this change only ensure you have a postgres instance up and running, which is the same behavior as CircleCI having a service for postgres.

After this change:
- setup postgres instance
- run all tests (every test uses its own database)
- tear down

Before this change:
- for each test
  - start a container
  - run test (uses its own database)
  - delete container
This is causing a lot of noise when containers are not available or failing tests not deleting containers.

I agree if you are talking about removing `E2ETestEnvironment`, but it means a lot of refactor, probably worth to think it in another PR.

jest.config.json
juanscolari: Not necessary. I just removed it. 

Context: It was there to help with the debugging tools, but it wasn't loading the proper configuration for each directory. Solved [here](https://github.com/decentraland/catalyst/pull/847/commits/20fb1762cf9267ee80d5a0e3869c428eecd998e6)

content/jest.setupFilesAfterEnv.ts
```diff
@@ -0,0 +1,18 @@
...
+  jest.spyOn(logger, 'createLogComponent').mockImplementation(() => {
+    const logComponentMock = createLogComponent()
+    const loggerMock = logComponentMock.getLogger('__test__')
+    loggerMock.debug = () => {}
+    loggerMock.error = () => {}
+    loggerMock.info = () => {}
+    loggerMock.log = () => {}
+    loggerMock.warn = () => {}
+    logComponentMock.getLogger = jest.fn(() => loggerMock)
+    return logComponentMock
```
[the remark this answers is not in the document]
juanscolari: This is not ultimately necessary, but currently our CI logs are bloated with unnecessary data, making them hard to read like [here](https://app.circleci.com/pipelines/github/decentraland/catalyst/4484/workflows/50038e87-2322-43fa-9c2a-e5dd4d294ab2/jobs/24434). I believe this could be mitigated in the past by setting the `LOG_LEVEL` env variable, but I understand the current logger implementation is lacking that capability.
So, the point of these changes is to improve readability of the tests by mocking the current logger.

What would you suggest in this case?

content/jest.setupFilesAfterEnv.ts
juanscolari: On the same line as with the [logger mock](https://github.com/decentraland/catalyst/pull/847#discussion_r781440312) changes, this change is not ultimately necessary, but it aims to remove console logs done in the tests and improve readability of the CI logs.
What would you suggest in this case?

content/tsconfig.json
juanscolari: Removed it. It was a way to avoid come clashes with Jasmine in the meantime but it is not necessary anymore.

commons/jest.config.json
juanscolari: Added 👍

content/jest.globalSetup.ts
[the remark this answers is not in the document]
juanscolari: I completely agree on unifying coverage reports. On that line, I introduced Jest `projects` [here](https://github.com/decentraland/catalyst/pull/847/commits/dfa9f433c764375ea07a7ccee60ab4858717e8db). We can have a single config file for both unit and integration tests and have specific configurations for each. If we run them together, it will produce a unified coverage report.

A PR for unifying content tests with parallelism on the CI is being developed here: https://github.com/decentraland/catalyst/pull/850

content/jest.setupFilesAfterEnv.ts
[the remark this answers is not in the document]
menduz: I'd sacrifice readability in the console runner to benefit debug capabilities. Now that every test is passing ` bloated with unnecessary data` is a valid argument, but when you are trying to fix a hairy bug then you really need those logs.

I'd make it optional with an env var to enable local debugging with logs or add the log level env var to here https://github.com/well-known-components/logger/blob/2a1765ec3a3b5236c691f91a6083a9c5d1745694/src/helpers.ts#L12 if that suffices your use case

menduz:
🎉

PR #850, opened 2022-01-11
test: trying circleci parallelism with jest

guidota:
# Description

Adds parallelism to `build-content` job, splitting test by timings. It also stores the result which give us some insights around the tests.

Adds a job for coverage collection and report.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,11 +103,10 @@ jobs:
       - run:
           name: Run tests
           command: |
-            echo 'export JEST_JUNIT_OUTPUT_NAME=junit-${CIRCLE_NODE_INDEX:-0}.xml' >> $BASH_ENV
             cd content
-            TESTFILES=$(circleci tests glob "**/test/**/*.spec.ts" | circleci tests split --split-by=timings)
-            yarn test $TESTFILES --coverageDirectory=reports/${CIRCLE_NODE_INDEX:-0} --reporters=jest-junit
-            mv reports/${CIRCLE_NODE_INDEX:-0}/coverage-final.json reports/coverage-${CIRCLE_NODE_INDEX:-0}.json
+            TESTFILES=$(circleci tests glob "test/**/*.spec.ts" | circleci tests split --split-by=timings --timings-type=classname)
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -96,20 +96,45 @@ jobs:
           POSTGRES_DB: postgres_test
           POSTGRES_PASSWORD: '12345678'
           POSTGRES_USER: postgres
+    parallelism: 20
```
menduz, line 99: This makes CI expenses 20 times more expensive in exchange of  4x faster builds. I'd tune it to allow parallelism of 2,4 or 4 to optimize both values. Did you test other concurrency values?

content/jest.globalSetup.ts
```diff
@@ -1,19 +1,41 @@
...
+    // start postgres container and wait for it to be ready
+    const container = await new GenericContainer('postgres', '12')
+      .withName(postgresContainerName)
       .withEnv('POSTGRES_PASSWORD', DEFAULT_DATABASE_CONFIG.password)
       .withEnv('POSTGRES_USER', DEFAULT_DATABASE_CONFIG.user)
       .withExposedPorts(E2ETestEnvironment.POSTGRES_PORT)
       .withWaitStrategy(new PostgresWaitStrategy())
       .start()
+
+    global.__POSTGRES_CONTAINER__ = container
```
menduz, line 36: Prefer globalThis, global is non-standard

.circleci/config.yml
guidota: according to [this report](https://app.circleci.com/insights/github/decentraland/catalyst/workflows/build-deploy/jobs?branch=chore%2Fjest-circleci-with-parallelism), our slowest test time is **30s**.

it is currently splitting by names which is not the idea, if we find out how to split by timings I guess we can use lower values of parallelism and get the same result.

I will test with those values

guidota: parallelism of 4 and splitting by timings: 2.24s

see [workflow](https://app.circleci.com/pipelines/github/decentraland/catalyst/4707/workflows/1b93b4ef-3688-4d43-92f4-88d409647aff)

credits usage still low compared to `main` branch

agusaldasoro (approved):
Awesome work!

PR #854, opened 2022-01-13
chore: lock `faker.js` version on content `package.json`

juanscolari:
## Description

The last [faker.js](https://www.npmjs.com/package/faker) package version 6.6.6 removed all the original code and was replaced with an [empty repository](https://github.com/Marak/faker.js).

In order to prevent any other minor bump in the version, I locked the current `faker` version with the one in the `yarn.lock` (5.5.3)


guidota (approved):
LGTM 👍

PR #855, opened 2022-01-13
test

agusaldasoro:
## Description

Please include a summary of the change and which issue is fixed/closed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

Closes # (issue)

## Changes

Specify the behavior changes introduced by this pull request

i.e.

- Removed method X from class Y
- Return 400 instead of 500 in Z case
- Refactored class X into a new module

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] I have read the [**CONTRIBUTING**](https://github.com/decentraland/catalyst/blob/main/docs/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.
- [ ] All new and existing tests passed.

### DIFF
--- a/content/src/service/synchronization/SynchronizationManager.ts
+++ b/content/src/service/synchronization/SynchronizationManager.ts
@@ -87,7 +87,7 @@ export class ClusterSynchronizationManager implements SynchronizationManager, IS
         await bootstrapFromSnapshots(this.components, this.components.contentCluster)
       } catch (error) {
         ClusterSynchronizationManager.LOGGER.info(`Error bootstrapping: ${error}`)
-        process.exit(1)
+        throw error;
       }
       this.synchronizationState = SynchronizationState.SYNCED
     }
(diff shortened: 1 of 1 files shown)


content/src/service/synchronization/SynchronizationManager.ts
```diff
@@ -83,7 +83,12 @@ export class ClusterSynchronizationManager implements SynchronizationManager, IS
     bootstrap: {
       // Note: If any deployment was overwritten by the snapshots, then we never reach them
       ClusterSynchronizationManager.LOGGER.info(`Starting to bootstrap from snapshots`)
-      await bootstrapFromSnapshots(this.components, this.components.contentCluster)
+      try {
+        await bootstrapFromSnapshots(this.components, this.components.contentCluster)
+      } catch (error) {
+        ClusterSynchronizationManager.LOGGER.info(`Error bootstrapping: ${error}`)
+        process.exit(1)
```
menduz, line 90: `process.exit` should never be used, prefer throw instead and graceful shutdown.

Some reasons not to do it:
- Open DB connection may hang in the middle of execution of transaction yielding the database unusable.
- Downloaded files may be corrupt due to incomplete downloads
- Endpoints being served may result in data-loss from part of the clients

Alternative: simply "throw", since this function is called in the `start` method of the well-known-components lifecycle, the whole server will gracefully stop with exit code 1.

https://github.com/well-known-components/interfaces/blob/master/src/utils/lifecycle.ts#L108
followed by
https://github.com/well-known-components/interfaces/blob/069381a4e0f733a501cb14936ffd6359a5828a99/src/utils/lifecycle.ts#L249-L254


```suggestion
        throw error;
```

menduz (changes requested):

PR #860, opened 2022-01-14
Fail if not found catalysts in bootstrap

agusaldasoro:
Throw an error if the server couldn't bootstrap, also wait longer to get all servers in the cluster.


pentreathm (approved):

content/src/service/synchronization/bootstrapFromSnapshots.ts
```diff
@@ -13,11 +14,10 @@ type BootstrapComponents = Pick<
  * then iterates over all of the deployments to call the batch deployer for each deployed entity.
  */
 export async function bootstrapFromSnapshots(components: BootstrapComponents, cluster: ContentCluster): Promise<void> {
-  const catalystServers = await ensureListOfCatalysts(cluster, 10 /* retries */, 1000 /* wait time */)
+  const catalystServers = await ensureListOfCatalysts(cluster, 20 /* retries */, ms('30s') /* wait time */)
```
menduz, line 17: That change will halt the execution for up to 10 minutes. please consider only incrementing the amount of retries.

```suggestion
  const catalystServers = await ensureListOfCatalysts(cluster, 30 /* retries */, 1000 /* every 1sec */)
```

menduz (changes requested):

guidota (approved):

PR #862, opened 2022-01-16
fix: identity query

menduz:
Fixes the root cause of mitigation on #860 .

Makes self-identification more reliable and less consuming and dependant on the DAO.

Before:
- Hit all DAO catalysts to get their challenge
- Evaluate responses, if only one of the responses had the same challente, then that is us
- If more than one catalyst responds the same then fails
- Retry whole process N times in case of failure

After:
- Hit the endpoint specified in the CONTENT_SERVER_URL env var.
- If it belongs to the DAO and the challenge is the same. We found our identity
- If we got an answer and it does not belong to the DAO then converge identity as non-part of the dao
- Retry whole process N times in case of failure

This change reduces the amount of endpoint calls needed to converge with a valid identity (whether it belongs to DAO or not) to only one request.

If the public URL of CONTENT_SERVER_URL is not accessible, then the identity resolves to `undefined`

Attempts continue to exist to account for possible "warmup" of NGINX, cloudflare, or any other process or proxy.


agusaldasoro (approved):
Great work!

PR #865, opened 2022-01-17
Fix server killed when fails

agusaldasoro:
## Description

Lambdas server was killed when asked for content that was not present, for example: `/lambdas/images/abcde/256`


menduz (approved):
It may fix the problem right now but the code change is indeed a noop. All the logic for `existingDownloadsFutures` seems to be non-functional

guidota (approved):

PR #872, opened 2022-01-20
Modify get entities query to avoid unnecessary join

agusaldasoro:
- Modified query of getHistoricalDeployments to avoid making a self join to obtain `overwrittenBy` field, it is only calculated when `authChain` is requested in: `/pointer-changes` and `/audit/{entityType}/{entityId}`
- Removed `deployedBy` filter


comms/lighthouse/package.json
```diff
@@ -16,7 +16,7 @@
     "@dcl/archipelago": "1.0.0",
     "@dcl/catalyst-api-specs": "1.0.1",
     "cors": "2.8.5",
-    "dcl-catalyst-commons": "7.2.1",
+    "dcl-catalyst-commons": "tag-7.2.2-include-deployed-by",
```
jmoguilevsky, line 19: before merging to main, add the version here

content/src/logic/database-queries/deployments-queries.ts
```diff
@@ -140,47 +173,41 @@ export function getHistoricalDeploymentsQuery(
...
-
-  let where = SQL``
-  if (whereClause.length > 0) {
-    where = SQL` WHERE `.append(whereClause[0])
-    for (const condition of whereClause.slice(1)) {
-      where = where.append(' AND ').append(condition)
-    }
+function getBaseQuery(includeOverwrittenInfo?: boolean) {
+  // TODO: make another query for when deleter deployment is needed
+  if (!!includeOverwrittenInfo) {
```
jmoguilevsky, line 180: aren't both queries equal and the only difference the left join?

I think it can be added as are the other clauses

jmoguilevsky (approved):

PR #877, opened 2022-01-24
feat: update error when failed deployment retry fails

guidota:
## Description

Update failed deployment error description when retry task fails to fix it.

Fixing a failed deployment will be ignored when there are newer entities with same pointer and it will be removed from the cache.

### DIFF
--- a/content/src/logic/deployments.ts
+++ b/content/src/logic/deployments.ts
@@ -18,7 +18,14 @@ export async function isEntityDeployed(
 export async function retryFailedDeploymentExecution(
   components: Pick<
     AppComponents,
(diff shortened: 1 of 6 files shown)


content/src/logic/deployments.ts
```diff
@@ -44,6 +51,10 @@ export async function retryFailedDeploymentExecution(
           DeploymentContext.FIX_ATTEMPT
         )
       } catch (error) {
+        // it failed again, override failed deployment error description
```
agusaldasoro, line 54: Are you sure that if the deployment fails throws an exception? Can it fail without throwing an error?

agusaldasoro (approved):
This PR looks great, anyway consider changing the FailedDeploymentsCache to be indexed by pointer, so we don't store more than one failed deployment per pointer and we only store the latest one.

PR #882, opened 2022-01-25
chore: reduce off chain refresh time

juanscolari:
## Description

Deploying changes for the off chain wearables is taking a long time because results are cached for one day, making it difficult to check if what was deployed is correct, and also for testing and iterating. The goal of this PR is to reduce that time to a number that helps with these issues, while not impacting the catalyst performance.

## Changes

- Reduce the default time for off chain wearables cache refresh from `1 day` to `15 minutes`

### DIFF
--- a/lambdas/src/apis/collections/off-chain/OffChainWearablesManager.ts
+++ b/lambdas/src/apis/collections/off-chain/OffChainWearablesManager.ts
@@ -23,7 +23,7 @@ export class OffChainWearablesManager {
   }) {
     this.definitions = new TimeRefreshedDataHolder(
       () => OffChainWearablesManager.fetchOffChain(client, collections ?? DEFAULT_COLLECTIONS),
-      refreshTime ?? '1d'
+      refreshTime ?? '15m'
     )
   }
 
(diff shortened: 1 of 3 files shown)


lambdas/src/apis/collections/off-chain/OffChainWearablesManager.ts
```diff
@@ -23,7 +23,7 @@ export class OffChainWearablesManager {
   }) {
     this.definitions = new TimeRefreshedDataHolder(
       () => OffChainWearablesManager.fetchOffChain(client, collections ?? DEFAULT_COLLECTIONS),
-      refreshTime ?? '1d'
+      refreshTime ?? '15m'
```
agusaldasoro, line 26:
```suggestion
      refreshTime
```

And make `refreshTime` not optional parameter

agusaldasoro (changes requested):

lambdas/src/apis/collections/off-chain/OffChainWearablesManager.ts
juanscolari: Done

agusaldasoro (approved):

PR #883, opened 2022-01-25
chore: normalize storage

menduz:
## Description

Normalizes the storage implementation to remove friction on compressed assets

### DIFF
--- a/content/src/storage/ContentStorage.ts
+++ b/content/src/storage/ContentStorage.ts
@@ -5,6 +5,7 @@ export type ContentEncoding = 'gzip'
 
 export interface ContentStorage {
   storeStream(fileId: string, content: Readable): Promise<void>
+  storeStreamAndCompress(fileId: string, content: Readable): Promise<void>
   delete(fileIds: string[]): Promise<void>
   retrieve(fileId: string): Promise<ContentItem | undefined>
   exist(fileId: string): Promise<boolean>
@@ -60,7 +61,7 @@ export class SimpleContentItem implements ContentItem {
    * That may imply that the stream may be compressed, if so, the
    * compression encoding should be available in "content".
    */
-  async asRawStream() {
+  async asRawStream(): Promise<RawContent> {
     return {
       stream: await this.streamCreator(),
       encoding: this.encoding || null,
--- a/content/src/storage/ContentStorage.ts
+++ b/content/src/storage/ContentStorage.ts
@@ -1,24 +1,34 @@
-import { ContentFileHash } from 'dcl-catalyst-commons'
 import { Readable } from 'stream'
+import { createGunzip } from 'zlib'
 
 export type ContentEncoding = 'gzip'
 
 export interface ContentStorage {
-  storeStream(id: string, content: Readable): Promise<void>
-  /** @deprecated use storeStream instead */
-  store(id: string, content: Uint8Array): Promise<void>
-  delete(ids: string[]): Promise<void>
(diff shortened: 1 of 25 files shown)


content/src/storage/ContentStorage.ts
```diff
@@ -1,24 +1,34 @@
...
-  storeContent(fileHash: ContentFileHash, content: Uint8Array | Readable): Promise<void>
-  size(fileHash: ContentFileHash): Promise<number | undefined>
+  storeStream(fileId: string, content: Readable): Promise<void>
+  delete(fileIds: string[]): Promise<void>
+  retrieve(fileId: string): Promise<ContentItem | undefined>
+  exist(fileId: string): Promise<boolean>
+  existMultiple(fileIds: string[]): Promise<Map<string, boolean>>
+}
+
+export type RawContentItem = {
```
pedrotambo, line 14: nit: could it be possible to name it consistently with the `asRawStream`? Like `ContentRawStream`, `RawStream` or something alike?

content/src/storage/ContentStorage.ts
```diff
@@ -1,24 +1,34 @@
...
-  storeStream(id: string, content: Readable): Promise<void>
-  /** @deprecated use storeStream instead */
-  store(id: string, content: Uint8Array): Promise<void>
-  delete(ids: string[]): Promise<void>
-  retrieve(id: string): Promise<ContentItem | undefined>
-  exist(ids: string[]): Promise<Map<string, boolean>>
-  stats(id: string): Promise<{ size: number } | undefined>
-  storeContent(fileHash: ContentFileHash, content: Uint8Array | Readable): Promise<void>
-  size(fileHash: ContentFileHash): Promise<number | undefined>
+  storeStream(fileId: string, content: Readable): Promise<void>
```
pedrotambo, line 7: What about adding an `encoding` arg to signal if the content is already encoded or needs to be encoded before being stored, so we can get rid of the hack?
If not, the interface exposes the possibility to return a `RawStream` which could be encoded, but it doesn't have the ability to say that something is or should be encoded.

content/src/storage/ContentStorage.ts
```diff
@@ -32,16 +42,30 @@ export class SimpleContentItem implements ContentItem {
...
   }
 
-  async contentEncoding() {
-    return this.encoding ?? null
+  /**
+   * Used to get the raw stream, no matter how it is stored.
+   * That may imply that the stream may be compressed, if so, the
+   * compression encoding should be available in "content".
+   */
+  async asRawStream() {
```
pedrotambo, line 63: nit: add return type

content/src/storage/ContentStorage.ts
menduz: adding the encoding may generate problems with the file names. I've added a `storeStreamAndCompress` to remove the hack and also comply with this review

pedrotambo (approved):
Looks good!

pedrotambo:
> Would you test it in a production instance to prevent an accident by merging? I trust the coverage but I'm not sure about the integration e2e http tests https://app.circleci.com/pipelines/github/decentraland/catalyst/4780/workflows/fd00714a-b652-4006-9f36-b48d3097cbf3/jobs/26125

Yes, sure. Will give it a try and let you know

pedrotambo:
Update: seems to be working fine. Already deployed in `wc1`.

menduz:
:shipit:

PR #886, opened 2022-01-27
add catch to deploy entity

agusaldasoro:
- Adds dozens of logs and validations.
- Makes the deployments _not fail_ when repeated:
Screen Shot 2022-01-28 at 15 29 25

### DIFF
--- a/content/src/controller/Controller.ts
+++ b/content/src/controller/Controller.ts
@@ -149,7 +149,12 @@ export class Controller {
       }
     } catch (error) {
       this.components.metrics.increment('dcl_deployments_endpoint_counter', { kind: 'error' })
-      Controller.LOGGER.error(`POST /entities - Internal server error '${error}'`)
+      Controller.LOGGER.error(`POST /entities - Internal server error '${error}'`, {
+        entityId,
+        authChain: JSON.stringify(authChain),
+        ethAddress,
+        signature
+      })
       Controller.LOGGER.error(error)
       res.status(500).end()
     } finally {
--- a/content/src/controller/Controller.ts
+++ b/content/src/controller/Controller.ts
@@ -25,7 +25,12 @@ import { ContentAuthenticator } from '../service/auth/Authenticator'
 import { DeploymentOptions } from '../service/deployments/types'
 import { getPointerChanges } from '../service/pointers/pointers'
 import { DeploymentPointerChanges, PointerChangesFilters } from '../service/pointers/types'
-import { DeploymentContext, isSuccessfulDeployment, LocalDeploymentAuditInfo } from '../service/Service'
+import {
+  DeploymentContext,
+  isInvalidDeployment,
+  isSuccessfulDeployment,
+  LocalDeploymentAuditInfo
+} from '../service/Service'
 import { ContentItem } from '../storage/ContentStorage'
 import { AppComponents } from '../types'
 import { ControllerDeploymentFactory } from './ControllerDeploymentFactory'
@@ -134,10 +139,12 @@ export class Controller {
       if (isSuccessfulDeployment(deploymentResult)) {
         this.components.metrics.increment('dcl_deployments_endpoint_counter', { kind: 'success' })
         res.send({ creationTimestamp: deploymentResult })
-      } else {
+      } else if (isInvalidDeployment(deploymentResult)) {
         this.components.metrics.increment('dcl_deployments_endpoint_counter', { kind: 'validation_error' })
         Controller.LOGGER.error(`POST /entities - Deployment failed (${deploymentResult.errors.join(',')})`)
         res.status(400).send({ errors: deploymentResult.errors }).end()
+      } else {
+        throw new Error('deploymentResult is invalid')
       }
     } catch (error) {
       this.components.metrics.increment('dcl_deployments_endpoint_counter', { kind: 'error' })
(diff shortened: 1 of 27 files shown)


guidota (approved):

content/src/service/ServiceImpl.ts
```diff
@@ -74,43 +75,55 @@ export class ServiceImpl implements MetaverseContentService {
     context: DeploymentContext,
     task?: Database
   ): Promise<DeploymentResult> {
+    const deployedEntity = await this.getEntityById(entityId)
```
jmoguilevsky, line 78:
```suggestion
    const deployedEntity = await this.getEntityById(entityId, task)
```

content/src/controller/Controller.ts
```diff
@@ -134,14 +139,23 @@ export class Controller {
...
         res.send({ creationTimestamp: deploymentResult })
-      } else {
+      } else if (isInvalidDeployment(deploymentResult)) {
         this.components.metrics.increment('dcl_deployments_endpoint_counter', { kind: 'validation_error' })
-        Controller.LOGGER.error(`POST /entities - Returning error '${deploymentResult.errors.join('\n')}'`)
+        Controller.LOGGER.error(`POST /entities - Deployment failed (${deploymentResult.errors.join(',')})`)
         res.status(400).send({ errors: deploymentResult.errors }).end()
+      } else {
+        Controller.LOGGER.error(`deploymentResult is invalid ${JSON.stringify(deploymentResult)}`)
+        throw new Error('deploymentResult is invalid')
```
agusaldasoro, line 148: This error thrown will be caught to a response through WKC right?

content/src/logic/deployments.ts
```diff
@@ -52,10 +53,14 @@ export async function retryFailedDeploymentExecution(
         )
       } catch (error) {
         // it failed again, override failed deployment error description
-        const errorDescription = error.message
-        components.failedDeploymentsCache.reportFailure({ ...failedDeployment, errorDescription })
+        const errorDescription = error.message + ''
 
-        logger.info(`Failed to fix deployment of entity`, { entityId, entityType })
+        // TODO [mendez] this condition has no test coverage
+        if (!errorDescription.startsWith(IGNORING_FIX_ERROR)) {
```
agusaldasoro, line 59: The ignore of a failed deployment should correspond to fix it, not a fail: for example: if a failed deployment needs to be ignored that's because there's a new deployment for the same pointer correctly deployed

content/src/logic/deployments.ts
```diff
@@ -52,10 +53,14 @@ export async function retryFailedDeploymentExecution(
...
+          components.failedDeploymentsCache.reportFailure({ ...failedDeployment, errorDescription })
```
agusaldasoro, line 60: So, if the deployment is ignored it shouldn't be reported as a failure, right? It should be removed from the failed deployments instead.

content/src/logic/deployments.ts
```diff
@@ -52,10 +53,14 @@ export async function retryFailedDeploymentExecution(
         )
       } catch (error) {
         // it failed again, override failed deployment error description
-        const errorDescription = error.message
-        components.failedDeploymentsCache.reportFailure({ ...failedDeployment, errorDescription })
+        const errorDescription = error.message + ''
 
-        logger.info(`Failed to fix deployment of entity`, { entityId, entityType })
+        // TODO [mendez] this condition has no test coverage
```
agusaldasoro, line 58: I don't think this changes are necessary, read comments below

content/src/denylist/DenylistServiceDecorator.ts
```diff
@@ -102,16 +98,16 @@ export class DenylistServiceDecorator implements MetaverseContentService {
...
-      async (task) => {
-        // Validate the deployment
-        const hashedFiles = await this.validateDeployment(task.denylist, files, entityId, auditInfo)
-
-        // If all validations passed, then deploy the entity
-        return this.service.deployEntity(hashedFiles, entityId, auditInfo, context, task)
-      },
-      { priority: DB_REQUEST_PRIORITY.HIGH }
-    )
+    // return this.repository.task(
```
agusaldasoro, line 101: are we merging commented code?

content/src/service/ServiceImpl.ts
```diff
@@ -144,7 +172,19 @@ export class ServiceImpl implements MetaverseContentService {
...
       this.components.deployedEntitiesFilter.add(entity.id)
 
-      return storeResult.auditInfoComplete.localTimestamp
+      if (!storeResult.auditInfoComplete.localTimestamp) {
+        ServiceImpl.LOGGER.error(`auditInfoComplete is missbehaving`, {
+          auditInfoComplete: JSON.stringify(storeResult.auditInfoComplete)
+        })
+      }
+
+      // review this
```
agusaldasoro, line 181: Please add a `TODO:` here so we can search for this

content/src/logic/deployments.ts
menduz: Yes, that happens inside the deployEntity function

content/src/logic/deployments.ts
menduz: I added it because I've encountered an infinite loop of retries while debugging.

content/src/controller/Controller.ts
menduz: Catch of that throw is a couple of lines below. It sends a 500

content/src/logic/deployments.ts
agusaldasoro: Aaah, maybe that's you are learning the new deployment from every other catalyst?

PR #888, opened 2022-01-28
Migrate folder structure

### LINKED ISSUE #892: Add new file structure for content files
Context:
Currently, catalysts are storing all files in a big folder, which is, at the moment of writing, around 10 million files. This is causing an error in some catalysts that prevents the rename of a new file, thus preventing deployments from being done correctly

To store a content file the content server will:
1. Hash the filename (the CID of the content) and express it in hexadecimal
2. Use the first four chars as folder (this will allow for 65k folders)
3. Save the file keeping the name in the selected folder

Since the distribution of hashes among the 65k possibilities (0000 to FFFF) is equal in probability, no folder will have a number of files big enough for the rename problem to be seen again.

Other important changes that need to happen:
- The snapshot fetcher will need to store files in a tmp folder before storing the file in the corresponding one, to do this, the Storage component should be sent to the snapshot fetcher so that the logic is decoupled and not duplicated
- A file migration will need to be performed from the old file structure to the new one

jmoguilevsky:
## Description

This PR implements the new file structure, it implements the changes in the component storage and the corresponding migration

Closes #892

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] I have read the [**CONTRIBUTING**](https://github.com/decentraland/catalyst/blob/main/docs/CONTRIBUTING.md) document.
- [ ] All new and existing tests passed.


content/src/storage/FileSystemContentStorage.ts
```diff
@@ -83,6 +84,14 @@ export class FileSystemContentStorage implements ContentStorage {
...
-    return path.join(this.root, id)
+    // We are sharding the files using the first 4 digits of its sha1 hash, because it makes troubles
+    // for the file system to handle millions of files in the same directory.
+    // This way, asuming that sha1 hash distribution is ~uniform we are reducing by 16^4 the max amount of files in a directory.
+    const directoryPath = path.join(this.root, createHash('sha1').update(id).digest('hex').substring(0, 4))
+    if (!fs.existsSync(directoryPath)) {
+      fs.mkdirSync(directoryPath, { recursive: true })
+    }
+    const filePath = path.join(directoryPath, id)
+    return filePath
```
menduz, line 95: How does compressed files work?

```
hash(file+extension)[0..3] + / + file
```
or
```
hash(file)[0..3] + / + file + extension
```

content/src/types.ts
```diff
@@ -64,6 +65,7 @@ export type AppComponents = {
   catalystFetcher: Fetcher
   daoClient: DAOClient
   server: Server
+  contentFolderMigrationManager: ContentFolderMigrationManager
```
menduz, line 68: Please convert the MigrationManager into a function. It is not a port nor component nor manages anything

menduz (commented):
Looking good. Two minor things:
- I'd convert the manager into a function
- I want to understand the logic of compression, does it locate the file in a new folder?

menduz (approved):

content/test/integration/E2ETestEnvironment.ts
```diff
@@ -203,9 +203,7 @@ export function loadTestEnvironment(
     describe(name, () => {
       const testEnv = new E2ETestEnvironment()
 
-      it('starts the test environment', async () => {
-        await testEnv.start(overrideConfigs)
-      })
+      beforeEach(async () => { await testEnv.start(overrideConfigs) })
```
menduz, line 206: Please consider reverting this behavior. Granularity of tests should be executed at "describe" level in stateful flow based tests.

menduz (dismissed):

content/src/storage/FileSystemContentStorage.ts
pedrotambo: Option b: `hash(file)[0..3] + / + file + extension`

content/test/integration/E2ETestEnvironment.ts
pedrotambo: I read more about Jest and inspected the involved code and i think you're right, it doesn't feel correct to use `afterEach/beforeEach`. On the other hand, i found that using `beforeAll/afterAll` would work in the same way as using `it` but also would give use the chance to run a test singularly which i think is really useful for developing/debugging. Do you have any concerns about using `beforeAll/afterAll`?

menduz: A bit of story: Jasmine errors in `beforeAll/afterAll` were sent to /dev/null, yielding tests invalid with no information. That is the reason why `beforeAll/afterAll` were converted to `it`.

I think that if Jest does the same, we may keep them as `it` (Jest embeeds Jasmine, ouch). But I'd give it a try converting them to `beforeAll/afterAll` and putting a `throw` inside them to see what happens. If the suite fails then we are OK.

If errors are sent to /dev/null as well, I think keeping `it` is a best decision

content/src/storage/FileSystemContentStorage.ts
pedrotambo: The compression function just takes the file path of the file and save the compressed file in the same filepath + '.gzip'.
So if the `storeStreamAndCompress` method is called, it:
- Stores the uncompressed file in `this.getFilePath(id)`
- It tries to compress the file, if it success, then it places the compressed file in `this.getFilePath(id) + .gzip` (so it is in the same directory as the uncompressed file`
- Then deletes the uncompressed file in `this.getFilePath(id)`

menduz: Then I don't think this may work with compressed files. Please consider adding integration tests for it:

1. Write file to storage
2. Retrieve file must have encoding null
3. Compress stored file
4. 1) Retrieve file must have encoding gzip 2) Original file (uncompressed) must not exist anymore

content/test/integration/E2ETestEnvironment.ts
pedrotambo: Oh okay i see. As i read in some github issues some time ago that was an issue in Jest too (as it used jasmine in the darks), but now in `v27` jasmine was replaced by `jest-circus` and that issue was fixed, they say.
Anyway, i tested it myself and throwing an error within `beforeAll/afterAll` makes the test suite to fail and reports the errors correctly. So i think we're safe.

content/src/service/synchronization/deployRemoteEntity.ts
```diff
@@ -66,7 +69,7 @@ export async function deployDownloadedEntity(
   const deploymentTimeTimer = metrics.startTimer('dcl_deployment_time', { entity_type: entityType })
 
   try {
-    const entityFile = await fs.promises.readFile(path.join(components.staticConfigs.contentStorageFolder, entityId))
+    const entityFile = await fs.promises.readFile(path.join(components.staticConfigs.tmpDownloadFolder, entityId))
```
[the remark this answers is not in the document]
pedrotambo: Oh there it is what i was looking for hahah, thanks!

content/test/integration/E2ETestEnvironment.ts
menduz: Ok, I think it is safe to change to `beforeAll/afterAll` then

content/src/storage/FileSystemContentStorage.ts
menduz: I ment integration because it does not mock anything. Simply using the `FileSystemContentStorage` and real files.

.storeStreamAndCompress can be used safely for that test, I think.

If the unit test already uses the real FS without failing then we are covered

content/src/components.ts
```diff
@@ -40,8 +41,12 @@ export async function initComponentsWithEnv(env: Environment): Promise<AppCompon
   const repository = await RepositoryFactory.create({ env, metrics })
   const logs = createLogComponent()
   const fetcher = createFetchComponent()
+  const contentStorageFolder = path.join(env.getConfig(EnvironmentConfig.STORAGE_ROOT_FOLDER), 'contents', '_tmp')
+  const tmpDownloadFolder = path.join(contentStorageFolder, '_tmp')
```
menduz, line 45: I think tmpDownloadFolder is `xxx/contents/_tmp/_tmp`

content/src/metrics.ts
```diff
@@ -85,5 +85,10 @@ export const metricsDeclaration = validateMetricsDeclaration({
     help: 'Pending downloading jobs',
     type: 'gauge',
     labelNames: ['entity_type']
+  },
+  dcl_files_migrated: {
+    help: 'Files migrated to new folder structure',
+    type: 'counter',
+    labelNames: []
   }
```
menduz, line 93: This metric will not be useful since it doesn't change during the lifecycle of the program. A log will suffice

content/src/migrations/ContentFolderMigrationManager.ts
```diff
@@ -0,0 +1,84 @@
...
+import { join, resolve } from 'path'
+import { AppComponents } from '../../src/types'
+import { EnvironmentConfig } from '../Environment'
+
+export type ContentFolderMigrationComponents = Pick<AppComponents, 'logs' | 'env' | 'metrics' | 'storage'>
+
+export async function migrateContentFolderStructure(components: ContentFolderMigrationComponents) {
+  const queue = new PQueue({
+    concurrency: components.env.getConfig(EnvironmentConfig.FOLDER_MIGRATION_MAX_CONCURRENCY),
+    timeout: ms('30s')
```
menduz, line 15: Why do this job has a timeout?

content/src/migrations/ContentFolderMigrationManager.ts
```diff
@@ -0,0 +1,84 @@
...
+  await Promise.all(pending)
+
+  for (const file of failures) {
+    pending.push(
+      queue.add(async () => {
+        try {
+          await processFile(components, contentsFolder, file)
+        } catch (err) {
+          logs.error(`Retry for ${file} failed due to ${err}`)
+          failures.push(file)
```
menduz, line 56: What is the purpose of this push? `failures` is never cleared. Therefore if X fails, it is added to the failures. And if the retry fails it gets added again, rendering `failures=[X,X]`

content/src/service/synchronization/deployRemoteEntity.ts
```diff
@@ -46,17 +48,19 @@ async function downloadFullEntity(
       entityId,
       servers,
       serverLru,
-      components.staticConfigs.contentStorageFolder,
+      components.staticConfigs.tmpDownloadFolder,
       requestMaxRetries,
       requestRetryWaitTime
     )
+  } catch (e) {
+    console.log(`Error downloading file ${entityId}. Error: ${e}`)
```
menduz, line 56: This function is supposed to throw. Please catch the error in the caller

```suggestion
```

content/src/components.ts
pedrotambo: thxs, fixed.

content/test/integration/E2ETestEnvironment.ts
pedrotambo: sounds good

content/src/migrations/ContentFolderMigrationManager.ts
jmoguilevsky: But why throw? That would halt the whole process

content/src/helpers/fsWrapper.ts
```diff
@@ -0,0 +1,4 @@
+// This file was created just for testing purposes, mocking fs requires a third party lib like `mock-fs`
+// Since files like the content migrator uses both fs and fs/promises mocking it made the test code too hard to read
+// this adds no overhead and makes the test code simpler
+export * from 'fs'
```
menduz, line 4: Did you consider passing `createReadStream` to `migrateContentFolderStructure(components, { createReadStream })`?

menduz: That way you could pass a mocked function like this

```ts
migrateContentFolderStructure(components, { createReadStream: jest.mock() })
```

Preventing magic ✨ code injections and lifecycles

menduz: (this is not request for a change)

jmoguilevsky: I actually thought of this file to be used by everyone using `fs` to make simpler the tests for all files, but this also an option

menduz: It sounds like an anti-pattern and it is a code smell from my point of view. I think we should prevent adding these hacks when possible

content/package.json
```diff
@@ -19,7 +19,7 @@
     "@dcl/catalyst-api-specs": "1.0.1",
     "@dcl/content-validator": "1.0.1-20220128144241.commit-9e37746",
     "@dcl/schemas": "3.6.0",
-    "@dcl/snapshots-fetcher": "2.0.0",
+    "@dcl/snapshots-fetcher": "2.0.1-20220127235544.commit-66b24b2",
```
agusaldasoro, line 22: Remember to change this before merging

content/src/Environment.ts
```diff
@@ -9,6 +9,7 @@ import { AppComponents } from './types'
 export const CURRENT_CONTENT_VERSION: EntityVersion = EntityVersion.V3
 const DEFAULT_STORAGE_ROOT_FOLDER = 'storage'
 const DEFAULT_SERVER_PORT = 6969
+const DEFAULT_FOLDER_MIGRATION_MAX_CONCURRENCY = 1000
```
agusaldasoro, line 12: Is this the max amount of files per folder?

content/src/components.ts
```diff
@@ -41,8 +42,12 @@ export async function initComponentsWithEnv(env: Environment): Promise<AppCompon
   const repository = await RepositoryFactory.create({ env, metrics })
   const logs = createLogComponent()
   const fetcher = createFetchComponent()
+  const contentStorageFolder = path.join(env.getConfig(EnvironmentConfig.STORAGE_ROOT_FOLDER), 'contents')
+  const tmpDownloadFolder = path.join(contentStorageFolder, '_tmp')
+  await fs.promises.mkdir(tmpDownloadFolder, { recursive: true })
```
agusaldasoro, line 47: Does this fail if the folder already exists?

content/src/Environment.ts
jmoguilevsky: No, that's the default concurrency of migrations

content/src/components.ts
jmoguilevsky: Yes, it fails, @pedrotambo can we fix that and add a try catch
https://stackoverflow.com/questions/21194934/how-to-create-a-directory-if-it-doesnt-exist-using-node-js

pedrotambo: It fails but only when `recursive` is false.

agusaldasoro (approved):

PR #890, opened 2022-01-31
fix: failed deployments report

guidota:
## Description

Converts retry failed deployments job into a stoppable component.
Fixes condition to match with the ignore the fix error message.


agusaldasoro (approved):
Awesome work!

menduz (commented):
Epic

PR #894, opened 2022-02-03
Remove current denylist implentation

jmoguilevsky:
## Description

This PR removes the current denylist implementation, allowing for a new one to be built from the ground up


menduz (approved):
Amazing

PR #914, opened 2022-02-07
feat: add get active entities endpoint with request body (POST)

### LINKED ISSUE #901: Implement a new endpoint to retrieve active entities
TBD
Implement a new endpoint to retrieve active entities

guidota:
## Description

Adds `POST /entities/active` endpoint to retrieve all active entities by IDs or pointers.
Request body should contain IDs or pointers list (not both) to retrieve.

It includes a refactor on the entities cache, improving relation between pointers and active entity ids.

Updating `lru-cache` to v7, it has some performance improvements.

### To review

- It may impact the deprecated endpoint performance too since we are changing the cache strategy.
- Entities cache size.
- Test scenarios

Resolves #901


content/src/controller/Controller.ts
```diff
@@ -90,6 +94,41 @@ export class Controller {
...
+    ) {
+      res.status(400).send({ error: 'ids or pointers must be present, but not both' })
+      return
+    }
+
+    let entities: Entity[] = []
+    if (ids && ids.length > 0) {
+      entities = await this.components.deployer.getEntitiesByIds(ids)
+    } else if (pointers && pointers.length > 0) {
+      for (const type of Object.values(EntityType)) {
```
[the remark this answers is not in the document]
agusaldasoro: The entity type shouldn't be necessary anymore, as to do the query in the database is not mandatory.

content/src/controller/Controller.ts
```diff
@@ -90,6 +94,41 @@ export class Controller {
...
+        entities = [...entities, ...(await this.components.deployer.getEntitiesByPointers(type, pointers))]
```
agusaldasoro, line 123: This method should be `await this.components.deployer.getEntitiesByPointers(pointers)`

agusaldasoro (changes requested):
Great work, and great tests. Please revisit the `getEntititiesByPointers` method, maybe deprecate the old one and create a new one which doesn't include the filter by entityType?

content/src/controller/Controller.ts
guidota: changed cache to be by pointer (removing entity type)

content/src/controller/Controller.ts
```diff
@@ -85,14 +89,47 @@ export class Controller {
...
+    res: express.Response
+  ): Promise<void> {
+    // Method: POST
+    // Path: /entities/active
+    // Body: { ids: string[], pointers: string[]}
+
+    const ids: EntityId[] = req.body.ids
+    const pointers: Pointer[] = req.body.pointers
+
+    console.log(JSON.stringify(req.body))
```
menduz, line 111:
```suggestion
```

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,169 @@
...
+  components: Pick<AppComponents, 'database' | 'env' | 'logs' | 'metrics'>
+): ActiveEntities => {
+  const logger = components.logs.getLogger('ActiveEntities')
+  const cache = new LRU<EntityId, Entity>({
+    max: components.env.getConfig(EnvironmentConfig.ENTITIES_CACHE_SIZE)
+  })
+  const entityIdByPointers = new Map<Pointer, EntityId>()
+
+  const reportCacheAccess = (entity: Entity, result: 'hit' | 'miss') => {
+    components.metrics.increment('dcl_entities_cache_accesses_total', {
```
menduz, line 38: 💜

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,169 @@
...
+    const entity = cache.get(entityId)
+    if (entity) {
+      for (const pointer of entity.pointers) {
+        entityIdByPointers.delete(pointer)
+      }
+    }
+    cache.del(entityId)
+  }
+
+  const updateCacheAndReturnAsEntities = (deployments: PartialDeploymentHistory<Deployment>): Entity[] => {
```
menduz, line 58: I love how function names are clear enough to not require any extra comment in the code

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,169 @@
...
+    const deployments = await getDeployments(components, {
+      filters: { pointers, onlyCurrentlyPointed: true }
+    })
+    for (const deployment of deployments.deployments) {
+      deployment.pointers.forEach((pointer) => entityIdByPointers.set(pointer, deployment.entityId))
+    }
+    return updateCacheAndReturnAsEntities(deployments)
+  }
+
+  const getByIds = async (...entityIds: EntityId[]): Promise<Entity[]> => {
```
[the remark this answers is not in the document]
guidota: `findEntitiesByIds` goes to the DB to retrieve the deployments, where `getByIds` is checking the cache otherwise calling to `findEntitiesByIds` 

added some comments btw

content/src/controller/Controller.ts
```diff
@@ -85,14 +89,46 @@ export class Controller {
...
+    // Method: POST
+    // Path: /entities/active
+    // Body: { ids: string[], pointers: string[]}
+
+    const ids: EntityId[] = req.body.ids
+    const pointers: Pointer[] = req.body.pointers
+
+    // Validate pointers or ids are present, but not both
+    if (
+      (ids && ids.length > 0 && pointers && pointers.length > 0) || // both present
```
jmoguilevsky, line 113: instead of a comment, create a const that is clear on what this clause mean

content/src/controller/Controller.ts
```diff
@@ -85,14 +89,46 @@ export class Controller {
...
+      ((!ids || ids.length == 0) && (!pointers || pointers.length == 0)) // none present
```
jmoguilevsky, line 114: same for this clause

content/src/service/ServiceImpl.ts
```diff
@@ -341,51 +350,14 @@ export class ServiceImpl implements MetaverseContentService {
...
-        id: entityId,
-        type: entityType,
-        pointers,
-        timestamp: entityTimestamp,
-        content: content?.map(({ key, hash }) => ({ file: key, hash })),
-        metadata
-      })
-    )
+  async getEntitiesByPointers(pointers: Pointer[]): Promise<Entity[]> {
+    return this.components.activeEntities.withPointers(...pointers)
```
jmoguilevsky, line 357: You don't need to destructure the ids since the method already receives an array

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+    const { deployments } = await getDeployments(components, {
+      filters: { pointers, onlyCurrentlyPointed: true }
+    })
+    for (const deployment of deployments) {
+      deployment.pointers.forEach((pointer) => entityIdByPointers.set(pointer, deployment.entityId))
+    }
+    return updateCacheAndReturnAsEntities(deployments)
+  }
+
+  const withIds = async (...entityIds: EntityId[]): Promise<Entity[]> => {
```
jmoguilevsky, line 93: Do you ever plan to call the function as in `withIds(a,b,c,d)` otherwise just call `entitiIds` the param

content/src/controller/Controller.ts
```diff
@@ -85,14 +89,46 @@ export class Controller {
...
+    ) {
+      res.status(400).send({ error: 'ids or pointers must be present, but not both' })
+      return
+    }
+
+    let entities: Entity[] = []
+    if (ids && ids.length > 0) {
+      entities = await this.components.deployer.getEntitiesByIds(ids)
```
agusaldasoro, line 122: is it correct that the GET method is inside a deployer?

content/src/controller/Controller.ts
```diff
@@ -85,14 +89,46 @@ export class Controller {
...
+    } else if (pointers && pointers.length > 0) {
+      entities = await this.components.deployer.getEntitiesByPointers(pointers)
```
agusaldasoro, line 124: same as below

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+import { AppComponents } from '../types'
+
+export type ActiveEntities = {
+  withPointers(...pointers: Pointer[]): Promise<Entity[]>
+  withIds(...entityIds: EntityId[]): Promise<Entity[]>
+  inactivate(...pointers: Pointer[]): void
+  activate(pointer: Pointer, entity: Entity): void
+}
+
+const mapDeploymentsToEntities = (deployments: Deployment[]): Entity[] => {
```
agusaldasoro, line 14: This method is only for the content field? :(

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  const reportCacheAccess = (entity: Entity, result: 'hit' | 'miss') => {
+    components.metrics.increment('dcl_entities_cache_accesses_total', {
+      entity_type: entity.type,
+      result: result
+    })
+  }
+
+  const cacheEntity = (entity: Entity): void => {
+    cache.set(entity.id, entity)
```
agusaldasoro, line 43: should it have a different TTL according to entity type?

content/src/service/ServiceImpl.ts
```diff
@@ -341,51 +350,14 @@ export class ServiceImpl implements MetaverseContentService {
   }
 
   async getEntitiesByIds(ids: EntityId[]): Promise<Entity[]> {
```
jmoguilevsky, line 352: Delete this method and use the activeEntities component directly from the controller

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+    cache.del(entityId)
+  }
+
+  const updateCacheAndReturnAsEntities = (deployments: Deployment[]): Entity[] => {
+    const entities = mapDeploymentsToEntities(deployments)
+    // Save the calculated values
+    for (const entity of entities) {
+      cacheEntity(entity)
+      for (const pointer of entity.pointers) {
+        entityIdByPointers.set(pointer, entity.id)
```
agusaldasoro, line 62: shouldn't we invalidate entities that were active before?

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+      filters: { entityIds, onlyCurrentlyPointed: true }
+    })
+    return updateCacheAndReturnAsEntities(deployments)
+  }
+
+  /**
+   * Queries DB to retrieve deployments using the given pointers as filter and return them as entities
+   * It also updates the cache
+   */
+  const findEntitiesByPointers = async (pointers: Pointer[]): Promise<Entity[]> => {
```
agusaldasoro, line 83: This is a breaking change for `/deployments` right? As now there will not be a way to retrieve not active deployments.

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+    const { deployments } = await getDeployments(components, {
+      filters: { pointers, onlyCurrentlyPointed: true }
+    })
+    for (const deployment of deployments) {
+      deployment.pointers.forEach((pointer) => entityIdByPointers.set(pointer, deployment.entityId))
```
agusaldasoro, line 88: Isn't this done in `updateCacheAndReturnAsEntities` too?

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+export const createActiveEntitiesComponent = (
+  components: Pick<AppComponents, 'database' | 'env' | 'logs' | 'metrics'>
+): ActiveEntities => {
+  const logger = components.logs.getLogger('ActiveEntities')
+  const cache = new LRU<EntityId, Entity>({
+    max: components.env.getConfig(EnvironmentConfig.ENTITIES_CACHE_SIZE)
+  })
+  const entityIdByPointers = new Map<Pointer, EntityId>()
```
agusaldasoro, line 33: Will be a limit for each entity type here? It'd be a pity to store stale profiles or use that space that should be a scene.

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  return {
+    /**
+     * Retrieve active entities by their ids
+     */
+    withIds,
+    /**
+     * Retrieve active entities that are pointed by the given pointers
+     */
+    withPointers: async (...pointers) => {
```
agusaldasoro, line 128: Can we define this method in a constant before being redundant with `withIds`?

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
+import { Deployment, Entity, EntityId, Pointer } from 'dcl-catalyst-commons'
+import LRU from 'lru-cache'
+import { EnvironmentConfig } from '../Environment'
+import { getDeployments } from '../service/deployments/deployments'
+import { AppComponents } from '../types'
+
+export type ActiveEntities = {
+  withPointers(...pointers: Pointer[]): Promise<Entity[]>
+  withIds(...entityIds: EntityId[]): Promise<Entity[]>
+  inactivate(...pointers: Pointer[]): void
```
agusaldasoro, line 10: can we rename this method to `invalidate` to be more self-explained?

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+  activate(pointer: Pointer, entity: Entity): void
```
agusaldasoro, line 11: Same here, maybe `reportNewVersion` or something like that?

content/src/service/ServiceImpl.ts
```diff
@@ -38,6 +38,8 @@ export class ServiceImpl implements MetaverseContentService {
   private static LOGGER: ILoggerComponent.ILogger
   private readonly pointersBeingDeployed: Map<EntityType, Set<Pointer>> = new Map()
 
+  private readonly deploymentsCache: { cache: NodeCache; maxSize: number }
```
agusaldasoro, line 41: Is this the rate limit cache? Why do you need to change that here? Are these changes compatible with the ones in https://github.com/decentraland/catalyst/pull/917 ?

content/src/service/ServiceImpl.ts
```diff
@@ -171,9 +178,6 @@ export class ServiceImpl implements MetaverseContentService {
         })
         this.components.metrics.increment('total_deployments_count', { entity_type: entity.type }, 1)
 
-        // Invalidate cache for retrieving entities by id
-        storeResult.affectedPointers?.forEach((pointer) => this.cache.invalidate(entity.type, pointer))
```
agusaldasoro, line 175: shouldn't you call here the `inactivate` cache method too?

content/src/service/ServiceImpl.ts
```diff
@@ -288,7 +287,17 @@ export class ServiceImpl implements MetaverseContentService {
               deploymentId,
               entity
             )
-            affectedPointers = Array.from(pointersFromEntity.keys())
+
+            // Update pointers to entityId cache
+            for (const pointerChange of pointersFromEntity) {
+              if (pointerChange[1].after === DELTA_POINTER_RESULT.CLEARED) {
```
agusaldasoro, line 293: Please add a comment here explaining that this means that the entity has been overwritten

content/src/service/ServiceImpl.ts
```diff
@@ -341,51 +350,14 @@ export class ServiceImpl implements MetaverseContentService {
   }
 
   async getEntitiesByIds(ids: EntityId[]): Promise<Entity[]> {
-    const deployments = await getDeployments(this.components, {
-      filters: { entityIds: ids, onlyCurrentlyPointed: true }
-    })
-    return this.mapDeploymentsToEntities(deployments)
+    return this.components.activeEntities.withIds(...ids)
```
agusaldasoro, line 353: niceee

content/src/service/ServiceImpl.ts
```diff
@@ -341,51 +350,14 @@ export class ServiceImpl implements MetaverseContentService {
...
-      ({ entityVersion, entityId, entityType, pointers, entityTimestamp, content, metadata }) => ({
-        version: entityVersion,
-        id: entityId,
-        type: entityType,
-        pointers,
-        timestamp: entityTimestamp,
-        content: content?.map(({ key, hash }) => ({ file: key, hash })),
-        metadata
-      })
-    )
```
agusaldasoro, line 386: 👏

content/src/ports/activeEntities.ts
pedrotambo, line 88: Is this still necessary? I think it's already done within `updateCacheAndReturnAsEntities` called below

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,175 @@
...
+        onCache.push(entity)
+        reportCacheAccess(entity, 'hit')
+      } else {
+        logger.debug('Entity not found on cache', { entityId })
+        missing.push(entityId)
+      }
+    }
+
+    // calculate values for those missing keys
+    const calculated: Entity[] = missing.length > 0 ? await findEntitiesByIds(missing) : []
```
pedrotambo, line 110: `notOnCache`? :P

content/src/ports/activeEntities.ts
guidota: to get not active deployments we can still use `getDeployments` without `onlyCurrentlyPointed` filter.

this should only retrieve active entities and store them in the cache

content/src/service/ServiceImpl.ts
guidota: I just removed the `ServiceFactory` so I moved some logic to the constructor here, will catch up with #917 changes.

content/src/ports/activeEntities.ts
guidota: yes :(

content/src/service/ServiceImpl.ts
```diff
@@ -329,7 +324,24 @@ export class ServiceImpl implements MetaverseContentService {
     // Mark deployment as successful (this does nothing it if hadn't failed on the first place)
     this.components.failedDeploymentsCache.removeFailedDeployment(entity.id)
 
-    return { auditInfoComplete, wasEntityDeployed: !isEntityAlreadyDeployed, affectedPointers }
+    return { auditInfoComplete, wasEntityDeployed: !isEntityAlreadyDeployed }
+  }
+
+  private updateActiveEntities(pointersFromEntity: PointerChanges, entity: Entity) {
+    const { cleared, setted } = Array.from(pointersFromEntity).reduce(
```
pedrotambo, line 331: past tense of set is set

content/src/service/ServiceImpl.ts
```diff
@@ -14,15 +14,16 @@ import {
...
 import { bufferToStream, ContentItem } from '../ports/contentStorage/contentStorage'
 import { FailedDeployment, FailureReason } from '../ports/failedDeploymentsCache'
 import { Database } from '../repository/Database'
 import { DB_REQUEST_PRIORITY } from '../repository/RepositoryQueue'
 import { AppComponents } from '../types'
-import { CacheByType } from './caching/Cache'
 import { getDeployments } from './deployments/deployments'
 import { DeploymentOptions } from './deployments/types'
 import { EntityFactory } from './EntityFactory'
+import { DELTA_POINTER_RESULT, DeploymentResult as PointerChanges } from './pointers/PointerManager'
```
pedrotambo, line 26: Why is `DeploymentResult` imported as `PointerChanges`? Note that `PointerChanges` is already defined in `types.ts` and generates confusion

content/src/service/ServiceImpl.ts
pedrotambo, line 331: could we change the name `cleared` and `setted` to  `clearedPointers` and `setPointers` or add its types?

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,241 @@
...
+  /**
+   * Retrieve active entities by their ids
+   * Note: result is cached, even if the id has no active entity
+   */
+  withIds(entityIds: EntityId[]): Promise<Entity[]>
+  /**
+   * Save entityId for given pointer and store the entity in the cache,
+   * useful to retrieve entities by pointers
+   */
+  update(pointers: Pointer[], entity: EntityCacheResult): void
```
pedrotambo, line 30: "Clearing" the pointers means importing the `NOT_ACTIVE` type from this module and passing that as a parameter to the update function. `NOT_ACTIVE` feels like an internal object to me (as we're not returning it anywhere, just omitting the pointers that does not have an active entity)

What do you think about having two methods to interact with the `ActiveEntities`? 

I was thinking in something sooomehow like:
`update(pointers: Pointer[], entity: Entity[])`
`clear(pointers)`

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,241 @@
...
+  /**
+   * Returns true if there is no active entity with given id
+   * Note: testing purposes
+   */
+  isEntityIdNotActive(entityId: EntityId): boolean
+  /**
+   * Returns true if there is no active entity with given pointer
+   * Note: testing purposes
+   */
+  hasNoActiveEntity(pointer: Pointer): boolean
```
pedrotambo, line 57: Same for this?

content/src/ports/activeEntities.ts
```diff
@@ -0,0 +1,241 @@
...
+   */
+  update(pointers: Pointer[], entity: EntityCacheResult): void
+  /**
+   * Returns the cached result:
+   *  - entity id if there is an active entity
+   *  - NONE if there is no active entity
+   *  - undefined if there is no cached result
+   * Note: testing purposes
+   */
+  getCachedActiveEntityOrNone(pointer: Pointer): PointerToEntityId | undefined
```
pedrotambo, line 42: I understand the motivation of exhausting tools for testing this key component. In addition to exposing internal behavior of having a cache (ActiveEntities is not a cache as far as i understand), what if we inject the `cache` and `entityIdByPointers` to the creation method so then in the test we can make the assertions/mock what we want?

content/src/ports/activeEntities.ts
```diff
@@ -39,22 +40,7 @@ export type ActiveEntities = {
...
-   * Returns true if there is no active entity with given id
-   * Note: testing purposes
-   */
-  isEntityIdNotActive(entityId: EntityId): boolean
-  /**
-   * Returns true if there is no active entity with given pointer
-   * Note: testing purposes
-   */
-  hasNoActiveEntity(pointer: Pointer): boolean
+  getCachedEntity(pointer: Pointer): EntityId | NotActiveEntity | undefined
```
pedrotambo, line 43: nit: `idOrPointer`

pedrotambo (approved):
Looks nice!

Issue #896, opened 2022-02-03
Support partial fetch to download content 


pentreathm:
In order to implement a Service Worker on the client side to try to cache videos locally and reduce the DataTransfer cost in CloudFlare, Partial Fetch support is needed in the Content Server, and these type of requests should return HTTP Status code 206 and support the range header. 

The HTTP 206 **Partial Content success status response** code indicates that the request has succeeded and the body contains the requested ranges of data, as described in the Range header of the request

juanscolari:
On Standby because of the incident: https://github.com/decentraland/rca/blob/main/Incidents/incident-2.md

PRs:
- https://github.com/decentraland/catalyst/pull/915
- https://github.com/decentraland/catalyst-api-specs/pull/23

PR #915, opened 2022-02-08
feat: add support for partial content

juanscolari:
## Description

Support partial fetch to download content when sending the `Range` header

Ignore the ranges when they are invalid, or the content is encoded

Closes [#896](https://github.com/decentraland/catalyst/issues/896)

Docs: https://github.com/decentraland/catalyst-api-specs/pull/23

## Changes

- Status Code
  - Set Status Code `206` only when requesting partial content. (`200` is still returned for the whole file)
- Request Headers
  - Parse `Range` header to support partial content
- Response Headers
  - Set `Accept-Ranges: bytes` header to indicate the client that there is support por partial content
  - Set `Content-Length` header accordingly. Por partial content, it's the size of the chunk. Por full files it's the size of the whole file
  - Set `Content-Range` header for partial content, describing the chunk and the total file size
- Response
  - Return the partial response when a `Range` header is provided

## Out of Scope

- Lack of support for [416](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416) Status Code errors. The current behavior is to return the whole file, as if no `Range` was provided.

## Supporting Documentation

- [Range Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests)
- [Partial Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206)
- [Range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range)
- [Accept-Ranges](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Ranges)
- [Content-Range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Range)
- [w3 Org Spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35.1)

## Types of changes

- [x] New feature (non-breaking change which adds functionality)
- [x] I have read the [**CONTRIBUTING**](https://github.com/decentraland/catalyst/blob/main/docs/CONTRIBUTING.md) document.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.


content/src/controller/Controller.ts
```diff
@@ -227,6 +232,28 @@ export class Controller {
...
+
+    if (!rangeHeader || !BYTES_RANGE_REGEXP.test(rangeHeader)) {
+      return
+    }
+
+    // Parse `start` and `end` ranges
+    // Set default values when they are not provided
+    const ranges = rangeHeader.substring(6, rangeHeader.length).split('-')
+    const start = ranges[0] ? Number(ranges[0]) : 0
+    const end = ranges[1] ? Number(ranges[1]) : Number.MAX_SAFE_INTEGER
```
agusaldasoro, line 247: what happens when the end is bigger than the file size?

content/src/storage/FileSystemContentStorage.ts
```diff
@@ -53,19 +54,31 @@ export class FileSystemContentStorage implements ContentStorage {
...
+    range?: ContentRange
+  ): Promise<ContentItem | undefined> {
     const extension = encoding ? '.' + encoding : ''
     const filePath = this.getFilePath(id) + extension
 
     if (await existPath(filePath)) {
       const stat = await fs.promises.stat(filePath)
+      if (range && stat.size) {
+        const safeRange = {
+          start: Math.min(range.start, stat.size - 1),
```
agusaldasoro, line 69: shouldn't this be `Math.max(range.start, 0)` ?

agusaldasoro (approved):

content/src/storage/FileSystemContentStorage.ts
```diff
@@ -53,19 +54,42 @@ export class FileSystemContentStorage implements ContentStorage {
...
     const extension = encoding ? '.' + encoding : ''
     const filePath = this.getFilePath(id) + extension
 
     if (await existPath(filePath)) {
       const stat = await fs.promises.stat(filePath)
+
+      if (range && stat.size) {
+        // Set default values for ranges when they are not defined
+        const rangeStart = range.start === undefined ? 0 : range.start
+        const rangeEnd = range.end === undefined ? stat.size - 1 : range.end
```
guidota, line 71:
```suggestion
        const rangeEnd = range.end ?? stat.size - 1
```

content/src/controller/Controller.ts
```diff
@@ -669,12 +691,21 @@ async function setContentFileHeaders(content: RawContent, hashId: string, res: e
...
   res.setHeader('Access-Control-Expose-Headers', 'ETag')
   res.setHeader('Cache-Control', 'public,max-age=31536000,s-maxage=31536000,immutable')
+  res.setHeader('Accept-Ranges', 'bytes')
 
   if (content.encoding) {
     res.setHeader('Content-Encoding', content.encoding)
   }
 
-  if (content.size) {
+  if (content.size && content.range?.start !== undefined && content.range?.end !== undefined) {
```
guidota, line 700: is it possible to `start` or `end` be undefined but still a valid range?

content/src/storage/FileSystemContentStorage.ts
guidota: nit:

guidota (commented):
nice 🙌

content/src/controller/Controller.ts
guidota: ok, it is not possible but maybe is a good idea to have the optional range but required `start`/`end` when you create the `ContentItem`

content/src/storage/FileSystemContentStorage.ts
juanscolari: I made it that way initially to prevent having a `range.start` bigger than the file size. I've now refactored the code to make it clearer, setting some default values when the ranges are not provided and validating them later

content/src/controller/Controller.ts
juanscolari: Thanks for pointing out this. I was able to come up with a more elegant solution here: https://github.com/decentraland/catalyst/pull/915/commits/67397820ff56ce06a1a7db63887c6a532a9d703f

The point was that the `range` that is used to create the headers is one that is calculated when reading the file. So, we have both start and end properties or none of them.

menduz:
Quick question, I'm AFK so I couldn't go through the code. 

will it work the same for compressed assets like the snapshot files?

juanscolari:
@menduz 
> will it work the same for compressed assets like the snapshot files?

Good catch. I've removed range support for files with encoding as it didn't work. I've also removed the header indicating support for range request for those files.

The current behavior  for files with encoding is to ignore the range headers and return the whole file with a `200` status code.

https://github.com/decentraland/catalyst/pull/915/commits/40ed52c6131897350e8f753be091ccc51d0fb631

PR #917, opened 2022-02-09
Rate limit when deploying any entity type

agusaldasoro:
Expand the rate limit when deploying functionality to all entity types, it was used only for profiles. 

It has a default configuration, but it can be overwritten by the owner of the catalyst.

### DIFF
--- a/content/src/Environment.ts
+++ b/content/src/Environment.ts
@@ -1,10 +1,10 @@
 import { DECENTRALAND_ADDRESS } from '@catalyst/commons'
-import { EntityVersion } from 'dcl-catalyst-commons'
+import { EntityType, EntityVersion } from 'dcl-catalyst-commons'
 import log4js from 'log4js'
 import ms from 'ms'
 import { initComponentsWithEnv } from './components'
 import { RepositoryQueue } from './repository/RepositoryQueue'
-import { AppComponents } from './types'
+import { AppComponents, parseEntityType } from './types'
 
 export const CURRENT_CONTENT_VERSION: EntityVersion = EntityVersion.V3
 const DEFAULT_STORAGE_ROOT_FOLDER = 'storage'
@@ -91,8 +91,8 @@ export enum EnvironmentConfig {
   SYNC_WITH_SERVERS_INTERVAL,
   CHECK_SYNC_RANGE,
   DECENTRALAND_ADDRESS,
-  DEPLOYMENTS_RATE_LIMIT_TTL,
-  DEPLOYMENTS_RATE_LIMIT_MAX,
+  DEPLOYMENTS_DEFAULT_RATE_LIMIT_TTL,
+  DEPLOYMENTS_DEFAULT_RATE_LIMIT_MAX,
   ETH_NETWORK,
   LOG_LEVEL,
   FETCH_REQUEST_TIMEOUT,
@@ -126,7 +126,9 @@ export enum EnvironmentConfig {
   BLOCKS_L2_SUBGRAPH_URL,
   VALIDATE_API,
   FOLDER_MIGRATION_MAX_CONCURRENCY,
-  RETRY_FAILED_DEPLOYMENTS_DELAY_TIME
+  RETRY_FAILED_DEPLOYMENTS_DELAY_TIME,
+  DEPLOYMENT_RATE_LIMIT_TTL,
+  DEPLOYMENT_RATE_LIMIT_MAX
 }
 
 export class EnvironmentBuilder {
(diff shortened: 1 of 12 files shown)


pedrotambo:
Looks nice!
What do you think about adding a simple metric, something like `rate_limit_deployments_total{entity_type=profile|wearable|..., limited=true|false` ?

content/src/ports/rateLimitDeploymentCacheMap.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+}
+
+function generateDeploymentCacheMap(
+  logs: ILoggerComponent.ILogger,
+  rateLimitConfig: RateLimitConfig
+): Map<EntityType, { cache: NodeCache; maxSize: number }> {
+  const envMaxPerEntity: Map<EntityType, number> = getMaxPerEntityMapConfig(rateLimitConfig.entitiesConfigMax)
+  const envTtlPerEntity: Map<EntityType, number> = getTtlPerEntityMapConfig(rateLimitConfig.entitiesConfigTtl)
```
menduz, line 63: Could those two functions be merged into one function returning a `Record<EntityName, { max: number, ttl: number }>` instead of two maps?

content/src/types.ts
```diff
@@ -90,3 +92,13 @@ export type StatusProbeResult = {
 export type IStatusCapableComponent = {
   getComponentStatus(): Promise<StatusProbeResult>
 }
+
+// TODO: Move this to catalyst-commons
+export function parseEntityType(strType: string): EntityType {
+  if (strType.endsWith('s')) {
+    strType = strType.slice(0, -1)
+  }
```
menduz, line 100: I'd love to remove that `if` if possible

agusaldasoro: Do you mean the logic of removing the s, or using an if instead of a regex?

content/src/ports/rateLimitDeploymentCacheMap.ts
agusaldasoro: That really makes sense! Will do it

content/src/Environment.ts
```diff
@@ -361,6 +363,37 @@ export class EnvironmentBuilder {
...
+      const rateLimitMaxConfig: Map<EntityType, number> = new Map(
+        Object.entries(process.env)
+          .filter(([name, value]) => name.startsWith('DEPLOYMENT_RATE_LIMIT_MAX_') && !!value)
+          .map(([name, value]) => [
+            parseEntityType(name.replace('DEPLOYMENT_RATE_LIMIT_MAX_', '')) as EntityType,
+            value as any as number
+          ])
+      )
+      return rateLimitMaxConfig ?? new Map()
+    })
```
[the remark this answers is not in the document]
agusaldasoro: Yes, it's something like a json?

content/src/types.ts
agusaldasoro: mm, I've only moved the method from another file. Doing that would result in a breaking change for the entire API

content/src/ports/rateLimitDeploymentCacheMap.ts
```diff
@@ -0,0 +1,134 @@
...
+  rateLimitConfig: RateLimitConfig
+): IRateLimitDeploymentCacheMapComponent {
+  const logs: ILoggerComponent.ILogger = components.logs.getLogger('RateLimitDeploymentCacheMapComponent')
+
+  const deploymentCacheMap: Map<EntityType, { cache: NodeCache; maxSize: number }> = generateDeploymentCacheMap(
+    logs,
+    rateLimitConfig
+  )
+
+  function getFromCache(entityType: EntityType): { cache: NodeCache; maxSize: number } {
```
pedrotambo, line 30: nit: `getCacheFromEntityType`?

content/src/ports/rateLimitDeploymentCacheMap.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+export type RateLimitConfig = {
+  defaultTtl: number
+  defaultMax: number
+  entitiesConfigTtl: Map<EntityType, number>
+  entitiesConfigMax: Map<EntityType, number>
+}
+
+export function createRateLimitDeploymentCacheMap(
```
pedrotambo, line 19: nit: `createDeploymentsRateLimiter`?

content/src/service/validations/server.ts
```diff
@@ -39,8 +43,10 @@ const localChecks = async (entity: Entity, serviceCalls: ServiceCalls): Promise<
     return `This entity was already deployed. You can't redeploy it`
 
   /** Validate the deployment is not rate limited */
-  if (await serviceCalls.isEntityRateLimited(entity))
+  if (await serviceCalls.isEntityRateLimited(entity)) {
+    components.metrics.increment('dcl_content_rate_limited_deployments_total', { entity_type: entity.type })
```
pedrotambo, line 47: Do we have another metric to count the total number of deployments (rate limited or not), so we can't track the rate? If so, that's fine. If not, wouldn't it be nice to have a label that says if the deployment was rate limited or no?

pedrotambo: That way we can keep track of a the real load in order to adjust the limits, if necessary

pedrotambo (approved):

content/src/service/validations/server.ts
agusaldasoro: There are several different metrics referring to the number of deployments made (failing or not / from the sync or organic), and they should be only one with different tags. Anyway, that implies a refactor not only in code but also in Grafana dashboards, so I'd leave that to another issue.

PR #922, opened 2022-02-14
Feat/add db queries observability

pedrotambo:
## Description

This PR adds a histogram metric of the durations of the queries to the database identified by name.
The motivation is to be able to identify performance issues with the database associated with specific queries.
For this reason, the instrumentation of the queries is optional. 

It was out the scope the deprecation of legacy mechanisms to access the database such as the ones made by `Repository.ts` or `Database.ts`.  The idea was to add as little as possible code here as we're planning to deprecated them.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [X] I have read the [**CONTRIBUTING**](https://github.com/decentraland/catalyst/blob/main/docs/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.


content/src/ports/postgres.ts
```diff
@@ -52,6 +53,15 @@ export async function createDatabaseComponent(
   // Config
   const pool: Pool = new Pool(finalOptions)
 
+  /*
+  Please, don't remove this listener. From node-postgres documentation:
+  "It is important you add an event listener to the pool to catch errors.  Just like other event emitters,  if a pool
+  emits an error event and no listeners are added node will emit an uncaught error and potentially exit."
+  */
+  pool.on('error', (err) => {
```
menduz, line 61: Which kind of errors are we actively ignoring with this lines?

content/src/instrument.ts
```diff
@@ -0,0 +1,24 @@
...
+    return new Promise<any>((res) => {
+      res(functionToInstrument(...args))
+    })
+      .catch((err) => {
+        queryStatus = 'error'
+        throw new Error(err)
+      })
+      .finally(() => {
+        endTimer({ status: queryStatus })
+      })
```
menduz, line 22: May I suggest a more readable version?

```suggestion
    try {
       const res = await functionToInstrument(...args)
       endTimer({ status: 'success' })
       return res
    } catch (err) {
      throw err
    }
    return new Promise<any>((res) => {
      res(functionToInstrument(...args))
    })
      .catch((err) => {
        queryStatus = 'error'
        throw new Error(err)
      })
      .finally(() => {
        endTimer({ status: queryStatus })
      })
```

menduz (commented):
Some comments in the code

content/src/instrument.ts
```diff
@@ -0,0 +1,24 @@
+import { AppComponents } from './types'
+
+export function instrumentWithQueryDurationMetric<F extends (...args: any[]) => any>(
+  components: Pick<AppComponents, 'metrics'>,
+  queryNameLabel: string,
+  functionToInstrument: F
+): F {
```
menduz, line 7: May this be simplified? It may reduce the final code using closures

```diff
      ? await instrumentWithQueryDurationMetric(
          components,
          durationQueryNameLabel,
-         (sql: SQLStatement): Promise<IDatabase.IQueryResult<T[]>> => {
+         async () => {
            return pool.query<T[]>(sql)
          }
-       )(sql)
+       )
```

```suggestion
export function instrumentWithQueryDurationMetric<T>(
  components: Pick<AppComponents, 'metrics'>,
  queryNameLabel: string,
  functionToInstrument: () => Promise<T>
): Promise<T> {
```

content/src/ports/postgres.ts
pedrotambo: Connection-level errors. Running the integration tests locally, all the tests pass, but the process running them exists with error: 
`Error: Unhandled error. (error: terminating connection due to administrator command
    at Parser.parseErrorMessage (/Users/pedrotamborindeguy/repos/catalyst/node_modules/pg-protocol/src/parser.ts:369:69)...`
After some investigation, i found that the error is happening when the postgres container is shutting down in `jest.globalTeardown.ts`, AFTER the database component is stopped. 
I tried others workaround but i couldn't work it out.
I can remove this lines for this PR and continue working on it separately

content/src/instrument.ts
pedrotambo: Yes, sure. I don't know why I thought it was a better idea to wrap the function only, for re-usage maybe. Anyways, refactored.

content/src/instrument.ts
pedrotambo: appreciated

menduz (approved):
Wonderful, long needed measurements!

Issue #926, opened 2022-02-15
Add atribute to health response to easily identify if all services are healthy


juanscolari:
This can help Load Balancers identify easily if all services are healthy, as they can find that data in a substring

Issue #925, opened 2022-02-15
Set content server as unhealthy while bootstrapping


juanscolari:
While bootstrapping, the content server is not able to provide all the necessary content to its clients, leading to errors in integrations. It should be marked as "Unhealthy" until it advances to the next state

PR #927, opened 2022-02-15
feat: set content server as healthy only when it is Syncing

juanscolari:
## Description

Prevent client issues trying to fetch content from a server that is bootstrapping

Closes https://github.com/decentraland/catalyst/issues/925
Closes https://github.com/decentraland/catalyst/issues/926

## Changes

- Mark content server as `Healthy` only when Syncing. Otherwise it would be `Unhealthy`, like when Bootstrapping
- Add attribute `readyToUse` to the health route to make it easy for a load balancer to understand the health of the whole system

### Minor changes
- Make calls to the comms, lambda, and content server in parallel instead of serial
- Create `PeerStatus` type for the peer health response, as it is not possible to create some type with the form `Record<string, HealthStatus> & { readyToUse: boolean }`

### Doc changes

- https://github.com/decentraland/catalyst-api-specs/pull/24

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] I have read the [**CONTRIBUTING**](https://github.com/decentraland/catalyst/blob/main/docs/CONTRIBUTING.md) document.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/lambdas/src/apis/status/health.ts
+++ b/lambdas/src/apis/status/health.ts
@@ -29,7 +29,9 @@ export async function refreshContentServerStatus(
 
     const obtainDeploymentTimeIsTooLong = obtainDeploymentTime > ms(maxDeploymentObtentionTime)
 
-    if (hasOldInformation || obtainDeploymentTimeIsTooLong) {
+    const isBootstrapping = (serverStatus as any).synchronizationStatus?.synchronizationState === 'Bootstrapping'
+
+    if (hasOldInformation || obtainDeploymentTimeIsTooLong || isBootstrapping) {
       healthStatus = HealthStatus.UNHEALTHY
     } else {
       healthStatus = HealthStatus.HEALTHY
--- a/lambdas/src/apis/status/health.ts
+++ b/lambdas/src/apis/status/health.ts
@@ -29,9 +29,10 @@ export async function refreshContentServerStatus(
 
     const obtainDeploymentTimeIsTooLong = obtainDeploymentTime > ms(maxDeploymentObtentionTime)
 
-    const isBootstrapping = (serverStatus as any).synchronizationStatus?.synchronizationState === 'Bootstrapping'
+    // This is the only valid syncronization state that ensures content is being served up to date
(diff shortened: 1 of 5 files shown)


lambdas/src/apis/status/health.ts
```diff
@@ -29,7 +29,9 @@ export async function refreshContentServerStatus(
 
     const obtainDeploymentTimeIsTooLong = obtainDeploymentTime > ms(maxDeploymentObtentionTime)
 
-    if (hasOldInformation || obtainDeploymentTimeIsTooLong) {
+    const isBootstrapping = (serverStatus as any).synchronizationStatus?.synchronizationState === 'Bootstrapping'
```
menduz, line 32: Hmm, I'd revert the condition, and consider the content server healthy if and only if `synchronizationState == "Syncing"` instead of `synchronizationState != "Bootstrapping"`. WDYT?

agusaldasoro: Agree on that, anyway I don't love the semantic on that, as being `Bootstrapping` is a healthy status but not ready to receive traffic yet 🤔 .

juanscolari: We also have the `"Synced"` [state](https://github.com/decentraland/catalyst/blob/97054c9f5ee093f61b978b9329d178039d930940/content/src/service/synchronization/SynchronizationManager.ts#L8-L12). Should I also include it? 

`synchronizationState == "Syncing" || synchronizationState == "Synced"`

menduz: `Synced` was removed with the new sync. we may delete it from the enum. Returning "Bootstrapping" instead of "Unhealthy" also sounds reasonable in my head I'm ok with both statuses as long as readyToUse==false if it is Bootstrapping

agusaldasoro (approved):

PR #966, opened 2022-03-14
Split images in dockerhub Chore/split images and comms

hugoArregui:
## Description

This PR removes lighthouse from this repo into https://github.com/decentraland/lighthouse and also builds and publish two different images for content and lambdas.


guidota (approved):
LGTM 👍

PR #970, opened 2022-03-16
feat: remote denylist and reload every two minutes

menduz:


juanscolari (approved):

content/src/Environment.ts
```diff
@@ -10,6 +10,8 @@ export const CURRENT_CONTENT_VERSION: EntityVersion = EntityVersion.V3
 const DEFAULT_STORAGE_ROOT_FOLDER = 'storage'
 const DEFAULT_SERVER_PORT = 6969
 const DEFAULT_DENYLIST_FILE_NAME = 'denylist.txt'
+const DEFAULT_DENYLIST_URLS = 'https://config.decentraland.org/denylist'
```
agusaldasoro, line 13: can we make this a configurable environment var?

agusaldasoro: If urlS are expected, shouldn't this be an array? Or what is the format for the variable?

content/src/ports/denylist.ts
```diff
@@ -1,50 +1,104 @@
+import { IBaseComponent } from '@well-known-components/interfaces'
 import { resolve } from 'path'
 import { createInterface } from 'readline'
+import { URL } from 'url'
 import { EnvironmentConfig } from '../Environment'
 import { AppComponents } from '../types'
+import { FSComponent } from './fs'
 
-export interface DenylistComponent {
```
agusaldasoro, line 6: The convention for ports is having Components defined, we should define them all in a similar way: https://github.com/decentraland/catalyst/blob/main/content/src/ports/bloomFilter.ts#L3

content/src/ports/denylist.ts
```diff
@@ -1,50 +1,104 @@
...
-      input: content,
-      crlfDelay: Infinity
+  const denylistsUrlsRaw = components.env.getConfig<string>(EnvironmentConfig.DENYLIST_URLS)?.split(/[\r\n\s]+/gm) ?? []
+  const denylistsUrls: URL[] = denylistsUrlsRaw
+    .filter((url) => {
+      try {
+        new URL(url)
+        return true
+      } catch {
+        return false
```
agusaldasoro, line 32: Please log something on failure

content/src/ports/denylist.ts
```diff
@@ -1,50 +1,104 @@
...
+      }
     })
+    .map((url) => new URL(url))
+  logger.info(`File: ${fileName}`)
```
agusaldasoro, line 36: Can we make these logs more descriptive about what are they printing?

content/src/ports/denylist.ts
```diff
@@ -1,50 +1,104 @@
...
+  const loadDenylists = async () => {
+    try {
+      const originalSize = deniedContentIdentifiers.size
+      await loadDenylistFromFile()
+
+      for (const url of denylistsUrls) {
+        await loadDenylistFromUrl(url)
+      }
+      if (deniedContentIdentifiers.size > originalSize) {
+        logger.info('New elements where added')
```
agusaldasoro, line 86:
```suggestion
        logger.info('New elements were added')
```

agusaldasoro, line 86: Can it happen that the elements changed but the size didn't?

content/src/ports/denylist.ts
pedrotambo: sure

content/src/ports/denylist.ts
pedrotambo: Yes, you're right... didn't want to log every 2 minutes in each reload as it would overflood the logs.. will delete it

content/src/ports/denylist.ts
menduz: Please let me know if there anything in particular to fix in the docs. I'm not sure what convention are you talking about here

content/src/Environment.ts
pedrotambo: When the `Environment` is built it reads the `process.env.DENYLIST_URLS` env var, if not exists, it takes the default. Or you are saying of doing it differently?
The format is a string of urls separated by `space|newline|tab`, as it would also be configurable by an envvar.

content/src/ports/denylist.ts
pedrotambo: Nothing wrong with the docs. We're discussing if there's a convention to add the keyword `Component` at the end of every component or not.

PR #990, opened 2022-04-01
Feat: add name to tp collections

nachomazzara:
## Description

Add `name` to third-party collections lambda and remove the `id` inside the `metadata` which is not being used. 

The explorer need it: 
Screen Shot 2022-04-01 at 18 50 29


guidota (approved):
lgtm 👍

PR #996, opened 2022-04-05
Remove changes field in pointer changes from /pointer-changes endpoint

agusaldasoro:
Necessary changes to be compliant with: https://github.com/decentraland/catalyst-api-specs/pull/29

Refactor the endpoint `/content/pointer-changes` to remove the field `changes` in the response as it wasn't being used and the calculation of it consumes many resources. Also, added `pointers` and `authChain` to standardize the endpoints used for synchronization.


lambdas/package.json
```diff
@@ -11,7 +11,7 @@
     "test": "jest --runInBand --forceExit --coverage"
   },
   "dependencies": {
-    "@dcl/catalyst-api-specs": "1.0.1",
+    "@dcl/catalyst-api-specs": "tag-test-new-pointer-changes",
```
menduz, line 14: change this before merge

content/src/controller/Controller.ts
```diff
@@ -717,13 +716,7 @@ export enum DeploymentField {
...
 }
 
-export type ControllerPointerChanges = Omit<DeploymentPointerChanges, 'changes'> & {
-  changes: {
-    pointer: Pointer
-    before: EntityId | undefined
-    after: EntityId | undefined
-  }[]
-}
+export type ControllerPointerChanges = Omit<DeploymentPointerChanges, 'changes'>
```
pedrotambo, line 719: I think the `ControllerPointerChanges` is not necessary right now? Or at least the `omit` thing?

content/src/controller/Controller.ts
```diff
@@ -410,8 +410,7 @@ export class Controller {
       })
     )
     const controllerPointerChanges: ControllerPointerChanges[] = deltas.map((delta) => ({
```
pedrotambo, line 412: If we use the `DeploymentPointerChanges` we could skip this mapping

content/src/controller/Controller.ts
pedrotambo: Or the thing to be omitted is the new field `pointers`?

content/src/service/pointers/pointers.ts
```diff
@@ -25,17 +22,12 @@ export async function getPointerChanges(
...
-  const deployments = deploymentsWithExtra.slice(0, curatedLimit)
-  const deploymentIds = deployments.map(({ deploymentId }) => deploymentId)
-  const deltasForDeployments = await getPointerChangesForDeployments(components, deploymentIds)
   const pointerChanges: DeploymentPointerChanges[] = deployments.map(
-    ({ deploymentId, entityId, entityType, localTimestamp, authChain }) => {
-      const delta = deltasForDeployments.get(deploymentId) ?? new Map()
-      const changes = transformPointerChanges(entityId, delta)
-      return { entityType, entityId, localTimestamp, changes, authChain }
+    ({ entityId, entityType, localTimestamp, pointers, authChain }) => {
+      return { entityId, entityType, localTimestamp, pointers, authChain }
```
pedrotambo, line 30: Some here, seems that it is mapping to the same thing

pedrotambo (changes requested):
Some comments that might require changes or not

content/src/controller/Controller.ts
agusaldasoro: I've removed the ControllerPointerChanges as now it's the same as DeploymentPointerChanges

content/src/service/pointers/pointers.ts
agusaldasoro: done

content/src/controller/Controller.ts
```diff
@@ -396,34 +396,28 @@ export class Controller {
...
       pagination.next = this.calculateNextRelativePathForPointer(lastPointerChange, pagination.limit, filters)
     }
 
-    res.send({ deltas: controllerPointerChanges, filters, pagination })
+    res.send({ deltas: pointerChanges, filters, pagination })
   }
 
   private calculateNextRelativePathForPointer(
-    lastPointerChange: ControllerPointerChanges,
+    lastPointerChange: DeploymentPointerChanges,
```
menduz, line 420: `DeploymentWithAuthChain` is the same type as  `DeploymentPointerChanges` can we keep only `DeploymentWithAuthChain` for maximum compatibility with snapshot-fetcher and to prevent accidental divergences?

agusaldasoro: Sure, done!

menduz (approved):

PR #997, opened 2022-04-05
feat: ignore entity types

menduz:
Selectively ignore entity types on sync using environment variable

### DIFF
--- a/content/src/service/synchronization/batchDeployer.ts
+++ b/content/src/service/synchronization/batchDeployer.ts
@@ -46,13 +46,13 @@ export function createBatchDeployerComponent(
    * servers only. Local deployments using POST /entities _ARE NOT_ filtered by this function.
    */
   async function shouldRemoteEntityDeploymentBeIgnored(entity: RemoteEntityDeployment): Promise<boolean> {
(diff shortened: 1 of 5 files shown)


content/src/service/synchronization/batchDeployer.ts
```diff
@@ -42,6 +45,11 @@ export function createBatchDeployerComponent(
     // ignore entities if those were successfully deployed during this execution
     if (successfulDeployments.has(entity.entityId)) return true
 
+    // ignore specific entity types using EnvironmentConfig.SYNC_IGNORED_ENTITY_TYPES
+    if (syncOptions.ignoredTypes.has(entity.entityType)) {
```
agusaldasoro, line 49: Can this be the first check? As it seems like the less expensive

agusaldasoro (approved):

agusaldasoro:
Can we add a disclaimer of the configuration to `Use at your own risk`? As we currently don't offer a mechanism to find entities in other servers if they weren't synced

content/src/service/synchronization/batchDeployer.ts
menduz: yes, makes sense

Issue #991, opened 2022-04-04
Delete Old snapshots


pedrotambo:
Delete old snapshots from Catalyst's storage. One tome function that runs at the bootstrapping of a Catalyst and goes through all the largest files and deletes them if they are snapshots.

PR #1005, opened 2022-04-11
Feat/snapshot cleaner

pedrotambo:
## Description

This PR adds a function to run only on startup that deletes all the snapshots with the current and old structure in the content storage directory. 
It sweeps the directory looking for files with a size greater than 50 MB, checks if a file is a snapshot and it's not being referenced in the content_table of the db and deletes it.
[Linked issue](https://github.com/decentraland/catalyst/issues/991#issue-1191799309)

## Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [X] I have read the [**CONTRIBUTING**](https://github.com/decentraland/catalyst/blob/main/docs/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.
- [X] All new and existing tests passed.


content/src/ports/gzipCompressor.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export interface FileCompressor {
+  compress(sourceFilepath: string, destinationFilepath: string): Promise<CompressionResult>
+  decompress(sourceFilepath: string, destinationFilepath: string): Promise<boolean>
+}
+
+export function createGzipCompressor(components: Pick<AppComponents, 'logs' | 'fs'>): FileCompressor {
+  return {
+    async compress(sourceFilepath: string, destinationFilepath: string) {
```
menduz, line 21: Since the compressor doesn't hold any state or lifecycle, I recommend you to move these functions to the `logic` folder and receiving the components as argument. If possible, prevent creating components that don't hold state or lifecycle hooks

```ts
export async function compress(components: Pick<AppComponents,'fs'>, sourceFilepath: string, destinationFilepath: string)
```

content/src/logic/snapshot-cleaner.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (stderr) {
+      throw new Error(`Error deleting old snapshots: ${stderr}`)
+    }
+    const relativeFilepaths = stdout ? stdout.trim().split('\n') : []
+    return relativeFilepaths
+  }
+
+  const resolvedPath = path.resolve(storageDirectory)
+  const pathsOfBigFiles = await getPathsOfBigFiles(resolvedPath)
+  const bigFilesHashes = pathsOfBigFiles.map((filepath) => filepath.substring(filepath.lastIndexOf('/') + 1))
```
menduz, line 82:
```suggestion
  const bigFilesHashes = pathsOfBigFiles.map((filepath) => basename(filepath).replace(/\.gzip/, ''))
```

content/src/logic/snapshot-cleaner.ts
```diff
@@ -0,0 +1,95 @@
...
+        await components.fs.unlink(potentialSnapshotFilepath)
+      }
+    } catch (error) {
+      logger.error(error, { filepath })
+    }
+  }
+
+  const getPathsOfBigFiles = async (directory: string): Promise<string[]> => {
+    const { stdout, stderr } = await executeCommand(
+      `find ${directory} -type f -size +${minimumSnapshotSizeInBytes - 1}c`
```
menduz, line 71: This doesn't work in windows

content/src/logic/snapshot-cleaner.ts
```diff
@@ -0,0 +1,95 @@
...
+    fs: Pick<FSComponent, 'unlink' | 'createReadStream'>
+  },
+  executeCommand: (command: string) => Promise<{ stdout: string; stderr: string }>,
+  storageDirectory: string,
+  minimumSnapshotSizeInBytes: number
+): Promise<void> {
+  const logger = components.logs.getLogger('Snapshot Cleaner')
+
+  const isModernSnapshotHeader = (header: string): boolean => header.startsWith('### Decentraland json snapshot')
+  const isLegacySnapshotHeader = (header: string): boolean => header.match(/\[\[\"Qm.+?(?=\",\[\")/) !== null
```
menduz, line 34: Can this two conditions be removed and only remove big files that are not part of any active deployment? This is tooooo specific and I see more value on having this same code but with fewer conditions.

Anyways, as part of GC, we want to delete non-important files, regardless if they are snapshots or not. And that's already being done by `filterContentFileHashes`

menduz (changes requested):

content/src/service.ts
```diff
@@ -1,14 +1,14 @@
...
-import { migrateContentFolderStructure } from './migrations/ContentFolderMigrationManager'
+import { cleanSnapshots } from './logic/snapshot-cleaner'
 import { bootstrapFromSnapshots } from './service/synchronization/bootstrapFromSnapshots'
 import { AppComponents } from './types'
 
 // this function wires the business logic (adapters & controllers) with the components (ports)
 export async function main(program: Lifecycle.EntryPointParameters<AppComponents>): Promise<void> {
   const { components, startComponents } = program
 
-  await migrateContentFolderStructure(components)
```
menduz, line 11: I don't think this is a good moment to remove the `migrateContentFolderStructure(components)`, it still migrates hundreds of files every time the service starts. Do we know why it happens?

pedrotambo: I didn't know. Will get it back and investigate why.

content/src/logic/snapshot-cleaner.ts
menduz: I think that would simplify _a lot_ the proposed implementation without losing any value

content/src/logic/snapshot-cleaner.ts
pedrotambo: Won't work even if the docker container is running linux?

menduz: Regarding windows, we had issues created by folks running catalysts in windows. I don't think we should close that door.

Mmm I can think about some alternatives, but not a clear path ahead...

- how long does it take to execute that `find ...` command in the catalyst?
- how long would `glob('*/*')` take? (maybe it may be part of the storage component, for consistency and abstraction)

those answers may lead to different implementation questions like:

- how long would take to run `storage.allEntries().filter(db.isActiveFile)`, skipping the stat+uncompress?
- how long `let bloom = bloomFilter(db.activeFiles); storage.allEntries().filter(not in bloom).filter(db.isActiveFile).map(storage.deleteFile)` would take?

content/src/logic/snapshot-cleaner.ts
pedrotambo: Yes, it would be simpler but less secure in some way. I still don't have the expertise on the content server to evaluate the security of that decision, but i believe in you :)

menduz: I will defer to the team here for a final decision. From my understanding, as long as the process runs _before_ you start the server component or bind the router, then it should be safe. Because there are no other parts of the code touching these files and the server is definitely not serving the files to anyone

content/src/logic/database-queries/snapshot-cleaner-queries.ts
```diff
@@ -0,0 +1,18 @@
+import SQL from 'sql-template-strings'
+import { AppComponents } from '../../types'
+
+export async function filterContentFileHashes(
```
agusaldasoro, line 4: the name of this function may be ambiguous, is it keeping or filtering out the content file's hashes?

content/src/logic/snapshot-cleaner.ts
```diff
@@ -0,0 +1,95 @@
...
+    fs: Pick<FSComponent, 'unlink' | 'createReadStream'>
+  },
+  storageDirectory: string,
+  minimumSnapshotSizeInBytes: number
+): Promise<void> {
+  return cleanSnapshotsWithExec(components, promisify(exec), storageDirectory, minimumSnapshotSizeInBytes)
+}
+
+// This function that takes exectueCommand that is for injecting a mock and being able to unit test.
+// Didn't found a way to mock exec/promisify functions.
```
agusaldasoro, line 22:
```suggestion
// The function executeCommand is for injecting a mock and being able to unit test.
// Didn't find a way to mock exec/promisify functions.
```

agusaldasoro, line 22:
```suggestion
// The function executeCommand is for injecting a mock and being able to unit test.
// Didn't find a way to mock exec/promisify functions.
```

content/src/logic/snapshot-cleaner.ts
agusaldasoro: I prefer to delete first only the snapshots files, as there are many and they take up lots of space. After that, we can analyze if there are still big files unreferenced and what to do with them. In the case we need to delete them as well, it'd be as easy as removing these checks on the next release.

content/src/logic/snapshot-cleaner.ts
agusaldasoro: This script is a one-time run, as we need to delete some forgotten snapshots files, but with the bump of the snapshots-fetcher version, we won't get any more of them. So, as long as this code doesn't crash on Windows we are good to go. Maybe add a try/catch?

content/src/logic/snapshot-cleaner.ts
```diff
@@ -0,0 +1,95 @@
...
+      logger.error(error, { filepath })
+    }
+  }
+
+  const getPathsOfBigFiles = async (directory: string): Promise<string[]> => {
+    const { stdout, stderr } = await executeCommand(
+      `find ${directory} -type f -size +${minimumSnapshotSizeInBytes - 1}c`
+    )
+    if (stderr) {
+      throw new Error(`Error deleting old snapshots: ${stderr}`)
```
agusaldasoro, line 74: Does this error stops the Content Server execution? We may want to ignore this crash

agusaldasoro, line 74: Does this error stops the Content Server execution? We may want to ignore this crash

content/src/service.ts
agusaldasoro: I agree that we should remove that in another PR, anyway I think that the logger says that you are migrating files but there's nothing to do. We'll have to validate that.

agusaldasoro (changes requested):

agusaldasoro:
Duplicated by https://github.com/decentraland/catalyst/pull/1008

PR #1006, opened 2022-04-11
Validate urn

agusaldasoro:
Modify `/entitites/active/currently-pointed/{urnPrefix}` endpoint to only accept urns that are Third Party Wearables collections and rename it to `/entities/active/collections/{collectionUrn}`.

This PR depends on:
- https://github.com/decentraland/urn-resolver/pull/118 
- https://github.com/decentraland/catalyst-api-specs/pull/31


content/src/controller/Controller.ts
```diff
@@ -142,8 +143,7 @@ export class Controller {
     // Path: /entities/currently-pointed/{urnPrefix}
     const urnPrefix: string = req.params.urnPrefix
 
-    const regex = /^[a-zA-Z0-9_.:,-]+$/g
-    if (!regex.test(urnPrefix)) {
+    if (!(await isUrnPrefixValid(urnPrefix))) {
       return res
         .status(400)
         .send({ errors: `Invalid Urn prefix param: '${urnPrefix}'` })
```
pedrotambo, line 149: Could we here a more detailed error message like `Urn prefix param should be a valid urn prefix of a 3rd party collection` or something like that? (Assuming the name of the endpoint can't be changed)

content/package.json
```diff
@@ -15,14 +15,14 @@
...
-    "@dcl/catalyst-api-specs": "1.2.1",
+    "@dcl/catalyst-api-specs": "tag-rename-urn-prefix",
     "@dcl/catalyst-contracts": "1.0.2",
     "@dcl/catalyst-node-commons": "1.0.1",
     "@dcl/content-validator": "1.4.2",
     "@dcl/hashing": "^1.1.0",
     "@dcl/snapshots-fetcher": "3.0.2",
     "@dcl/schemas": "4.3.0",
-    "@dcl/urn-resolver": "1.3.0",
+    "@dcl/urn-resolver": "tag-tpw-collections-2",
```
pedrotambo, line 25: Just reminder to update this

pedrotambo (approved):

PR #1017, opened 2022-04-25
Fix/fix deployment filter

pedrotambo:
The current implementation of the BloomFilter doesn't work properly: after adding ~5M of elements, it has a high rate of false positives despite the size of the bitset. It could be that the hash function is clustering the elements.
This PR replaces the bloom filter with another one from a library.


content/src/logic/deployments.ts
```diff
@@ -8,13 +8,24 @@ import { AppComponents } from '../types'
...
   entityId: string
-) {
+): Promise<boolean> {
   // this condition should be carefully handled:
   // 1) it first uses the bloom filter to know wheter or not an entity may exist or definitely don't exist (.check)
   // 2) then it checks against the DB (deploymentExists)
-  return (await components.deployedEntitiesFilter.check(entityId)) && (await deploymentExists(components, entityId))
+  if (await components.deployedEntitiesFilter.check(entityId)) {
+    if (await deploymentExists(components, entityId)) {
+      components.metrics.increment('dcl_deployed_entities_filter_checks_total', { false_positive: 'no' })
```
menduz, line 19: can `false_positive` become `bloom_hit` to prevent double negation?

content/src/service/synchronization/batchDeployer.ts
```diff
@@ -106,7 +106,6 @@ export function createBatchDeployerComponent(
               DeploymentContext.SYNCED
             )
 
-            components.deployedEntitiesFilter.add(entity.entityId)
```
menduz, line 109: I think this was necessary. Was it moved somewhere else?

menduz (changes requested):

content/src/logic/deployments.ts
pedrotambo: yes, it felt weird. `hit` alone would be good? prefer not to expose which kind of filter it is

content/src/service/synchronization/batchDeployer.ts
pedrotambo: tl;dr: it's called twice
The `batchDeployer` calls `deployEntityFromRemoteServer` just the line above, which then calls `deployDownloadedEntity` that finally calls `deployEntity` where the `add` is done. This seems good to me because there's only one point where the component is mutated.

content/src/logic/deployments.ts
menduz: From a devexp point of view, the fact that it is a bloom filter tells you a lot about the implementation and the code you are reading, if filter is the name of the component and bloom filter isn't, somebody may think that it is simply a filter that returns true and false with 100% accuracy, which bloom filters don't do.

menduz (approved):

content/src/logic/deployments.ts
pedrotambo: Hmmm that's a good point... Now i want to add that to the metric and the variable:
`deployedEntitiesFilter` -> `deployedEntitiesBloomFilter`
`dcl_deployed_entities_filter_checks_total` -> `dcl_deployed_entities_bloom_filter_checks_total`

PR #1019, opened 2022-04-26
fix: TPW integration

agusaldasoro:
Bug fixing:
- Showing only approved third party collections
- Filtering from TP ownership only matching collections


pedrotambo (approved):
LGTM

PR #1050, opened 2022-05-05
fix: fix file fds leak and small refactor to prevent future ones related

pedrotambo:
This PR fixes a file descriptor leak associated to files that were store and compressed in the contentStorage.
After compressing the content, the code checked if it was effectively compressed, but for that opened an unnecessary read stream which was never destroyed. The small refactor includes the ability to know the encoding of the file without the need to open a file descriptor.


agusaldasoro (approved):
great work!!

PR #1061, opened 2022-05-13
fix: correct handling when deploying overlapping scenes with a new owner

marianogoldman:
## Description

New tests have been added in `deployment-entity-overlap.spec.ts` to verify some scenarios that could happen during scene deployments overlapping with previous ones:

- When new scene is deployed on overlapping parcels, then new deployment removes previous scenes from orphaned parcels
- When scene is deployed, then server checks for permissions
- When parcel changes owner, then new deployment by new owner succeeds and removes previous scenes from orphaned parcels

## Types of changes

- All new and existing tests passed.

### DIFF
--- a/content/test/integration/service/deployments/deployment-entity-overlap.spec.ts
+++ b/content/test/integration/service/deployments/deployment-entity-overlap.spec.ts
@@ -112,18 +112,12 @@ loadStandaloneTestEnvironment()("Integration - Deployment with Entity Overlaps",
       makeNoopValidator(components);
       makeNoopServerValidator(components);
 
-      // stub(components.externalCalls, 'queryGraph')
-      //   .callsFake((...args: any) => {
-      //     console.log("queryGraph called with", args);
-      //     return Promise.resolve({});
-      //   });
-
       // Deploy E1 on P1, P2
       await deploy(components, E1);
       await assertDeploymentsAre(components, E1);
 
       // Change ownership of P2
-      // Nothing to do really, as the mock above already allows it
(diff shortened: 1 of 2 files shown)


content/test/integration/service/deployments/deployment-entity-overlap.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+
+      // Deploy E2 on P1
+      await expect(deploy(components, E2)).rejects.toThrow('The provided Eth Address does not have access to the following parcel: (0,0),The provided Eth Address does not have access to the following parcel: (0,1)');
+      await assertDeploymentsAre(components, E1); // E2 should have not been deployed
+    }
+  );
+
+  testCaseWithComponents(
+    testEnv,
+    "When parcel changes owner, then new deployment by new owner succeeds and removes previous scenes from orphaned parcels",
```
agusaldasoro, line 109: isn't this the same test case as the first one?

agusaldasoro (commented):
Can we add a check that fails when trying to deploy a scene with parcels that you are not the owner but are overwritten?

content/test/integration/service/deployments/deployment-entity-overlap.spec.ts
marianogoldman: In a sense... yes. Except for the 2 scenes belong to different identities. But giving the permission validation is mocked, they are somewhat very similar. In this case the overlap is also different (as it uses on overlapping parcel + one different one).

agusaldasoro (approved):

PR #1116, opened 2022-05-31
fix: tpw url and remove log4j logger

agusaldasoro:
- Remove log4j lib to log as it wasn't working
- Fix the generation of TPW url


juanscolari (approved):
Thx for providing some extra tests 👍

PR #1159, opened 2022-06-30
feat: add post endpoint to retrieve profiles

agusaldasoro:
The explorer needs to retrieve many profiles at once for the chat functionality, the GET endpoint has a limit on the number of ids as they are sent as query parameters. 

This PR exposes a new endpoint to retrieve profiles with a body with ids.


lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -88,6 +88,39 @@ export async function getIndividualProfileById(
...
+  res: Response
+): Promise<Response<any, Record<string, any>> | undefined> {
+  // Method: POST
+  // Path: /lambdas/profiles
+  // Body: { ids: string[]}
+
+  const profileIds: string[] = req.body.ids ?? []
+
+  console.debug(req.body)
+  console.log(`ids: ${profileIds}`)
```
jmoguilevsky, line 107:
```suggestion
```

menduz:
Does this use the new `POST /entities/active`? or it is still using the previous endpoint?

agusaldasoro:
> Does this use the new `POST /entities/active`? or it is still using the previous endpoint? Can this be replaced by a call to `POST /entities/active` the content server directly? to remove responsibilities from lambdas

It can't be moved to the content server directly, as the Kernel needs the profile sanitized (for example: filter out no more owned wearables). Anyway, it's true that the lambdas endpoint should call POST /entities/active from content server.

agusaldasoro:
After @menduz feedback, this issue has been created: https://github.com/decentraland/catalyst/issues/1160 as the refactor needed exceeds the profiles use cases.

menduz:
what will be the impact of this endpoint merged as-is? will lambdas & content survive a `ab http://lambdas/profiles` ?

lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -88,6 +88,36 @@ export async function getIndividualProfileById(
...
+): Promise<Response<any, Record<string, any>> | undefined> {
+  // Method: POST
+  // Path: /lambdas/profiles
+  // Body: { ids: string[] }
+
+  const profileIds: string[] = req.body.ids ?? []
+
+  if (!profileIds) {
+    return res.status(400).send({ error: 'You must specify at least one profile id' })
+  }
```
mrbaguvix, line 108:
```suggestion
  if (profileIds.length === 0) {
    return res.status(400).send({ error: 'You must specify at least one profile id' })
  }
```

lambdas/src/apis/profiles/controllers/profiles.ts
```diff
@@ -88,6 +88,36 @@ export async function getIndividualProfileById(
...
+  wearables: WearablesOwnership,
+  profilesCacheTTL: number,
+  req: Request,
+  res: Response
+): Promise<Response<any, Record<string, any>> | undefined> {
+  // Method: POST
+  // Path: /lambdas/profiles
+  // Body: { ids: string[] }
+
+  const profileIds: string[] = req.body.ids ?? []
```
mrbaguvix, line 104: This would break the code if `ids` param is not passed in
```suggestion
  const profileIds: string[] = req.body?.ids ?? []
```

mrbaguvix (changes requested):

lambdas/src/apis/profiles/controllers/profiles.ts
agusaldasoro: fixed

PR #1156, opened 2022-06-30
refactor: calculate overwrites using deployments table

marianogoldman:
First part of #1112.
This PR reimplements the code for `calculateOverwrites` by running a query on the `deployments` table, and does not use `pointer_history` table anymore. This table will soon be deleted.

Issue #1112, opened 2022-05-27
Remove redudant tables


agusaldasoro:
The tables `last_deployed_pointers` and `pointer_history` are used to calculate the overwrites of the deployments. This implies lots of transactions when doing a POST /entities. Remove those tables and do the same logic using the other existing tables.

PR #1162, opened 2022-07-01
refactor: get last active deployments using deployments table

marianogoldman:
Second part of https://github.com/decentraland/catalyst/issues/1112.

This PR reimplements the code for `getLastActiveDeploymentsOnPointers()` by running a query on the deployments table, and does not use `last_deployed_pointers` table anymore. `last_deployed_pointers` will soon be deleted. Class `LastDeployedPointersRepository` is deleted.

It also removes `PointerHistoryRepository` as the content was no longer used since #1156.


pedrotambo (approved):

content/src/service/pointers/PointerManager.ts
```diff
@@ -12,72 +9,38 @@ export class PointerManager {
    * Commit a new entity, and return the 'before & after' generated by the deployment
    */
   async referenceEntityFromPointers(
-    lastDeployedPointersRepo: LastDeployedPointersRepository,
+    deploymentsRepo: DeploymentsRepository,
     deploymentId: DeploymentId,
-    entity: Entity
+    entity: Entity,
+    overwrote: Set<number>,
+    overwritten: boolean
```
agusaldasoro, line 16: what is this parameter used for? How is it calculated?

content/src/service/pointers/PointerManager.ts
```diff
@@ -12,72 +9,38 @@ export class PointerManager {
...
-    // Prepare variables
     const result: DeploymentResult = new Map()
-    const overwrite: Set<string> = new Set()
 
-    lastDeployments.forEach((lastDeployment) => {
-      // Calculate the intersection of pointers between the last deployment and the new one
-      const intersection: Set<string> = intersect(lastDeployment.pointers, entity.pointers)
+    if (!overwritten) {
+      // Fetch overwritten deployments from the DB
+      const overwrittenDeployments = await deploymentsRepo.getDeployments(overwrote)
```
agusaldasoro, line 22: Is it correct to name this `overwritten` if we are querying by `overwrote`?

content/src/service/pointers/PointerManager.ts
```diff
@@ -12,72 +9,38 @@ export class PointerManager {
    * Commit a new entity, and return the 'before & after' generated by the deployment
    */
   async referenceEntityFromPointers(
```
agusaldasoro, line 11: Can we add some more comments here explaining what does this method do in general and in each line?

agusaldasoro: Does this endpoint make an update in the db or is just a read only?

marianogoldman: It's read-only

content/src/service/pointers/PointerManager.ts
marianogoldman: Renamed and documented

PR #1168, opened 2022-07-04
refactor: group routes

manumena:
## Description

Group all routes as in well-known-components structure


lambdas/src/controllers/handlers/profiles/handlers.ts
```diff
@@ -304,3 +304,34 @@ export type ProfileMetadataForSnapshots = {
 type AvatarForSnapshots = {
   snapshots: AvatarSnapshots
 }
+
+function asyncHandler(handler: (req: Request, res: Response, next: NextFunction) => Promise<void>): RequestHandler {
```
agusaldasoro, line 308: can we remove this method and put all the code inside `createProfileHandler` instead?

lambdas/src/controllers/routes.ts
```diff
@@ -0,0 +1,126 @@
...
+  )
+  router.get('/collections/', (req, res) => getCollectionsHandler(theGraphClient, req, res))
+  router.get('/collections/wearables-by-owner/:owner', (req, res) =>
+    getWearablesByOwnerEndpoint(contentClient, theGraphClient, req, res)
+  )
+  router.get('/collections/wearables', (req, res) =>
+    getWearablesEndpoint(contentClient, theGraphClient, offChainManager, req, res)
+  )
+
+  // Functionality for Explore use case
```
agusaldasoro, line 117:
```suggestion
  // Functionality for Explorer use case
```

PR #1191, opened 2022-07-19
fix: bump snapshots-fetcher and content-validator that contain several fixes

marianogoldman:
# Description
This PR has a few changes and several bumps of dependencies. The goal of this is to fix all failed deployments that came up from starting a fresh content server from scrach.
These are the main fixes:
* It downloads the files referenced in snapshots part of the profile that are not included in content.
* It only checks that entity are v3 for deployments after ADR-45 timestamp.
* It changes the determination of blocks for wearable ownership checks.


content/test/integration/service/deployments/metadata-schema-validations.spec.ts
```diff
@@ -107,8 +107,8 @@ loadStandaloneTestEnvironment()("Integration - Deployment with metadata validati
       expect(await deployEntity(components, E1))
         .toEqual({
           errors: [
-            "The provided Eth Address does not have access to the following parcel: (0,0)",
-            "The provided Eth Address does not have access to the following parcel: (0,1)"
+            "The provided Eth Address does not have access to the following parcel: (0,0). Error: Unknown metric subgraph_errors_total",
```
marianogoldman, line 110: Undo this change

pedrotambo (approved):
LGTM

PR #1205, opened 2022-07-26
feat: add content info metrics

pedrotambo:
This PR adds multiple metrics involving current display information of the content-servers in the monitor.
Metrics added:
- dcl_content_server_build_info: static build info.
- dcl_content_server_snapshot_entities: number of entities by type in each snapshot. It is calculated in each snapshot generation.
- dcl_content_server_sync_state: BOOTSTRAPPING/SYNCING/SYNCED
- dcl_content_server_failed_deployments: current failed deployments.


marianogoldman (approved):

package.json
```diff
@@ -48,7 +48,8 @@
     "eslint-plugin-prettier": "4.0.0",
     "lint-staged": "12.4.1",
     "prettier": "2.6.2",
-    "typescript": "^4.4.4"
+    "typescript": "^4.4.4",
+    "syncpack": "8.2.4"
```
menduz, line 52: are we exploring syncpack?

pedrotambo: Hmmm not sure if i understand the question. We're currently using it to build `content` and `lambdas` with the same dependency versions. Just adding it to `devDependencies` so we don't have to install between every `yarn install, yarn build` cycle.

PR #1234, opened 2022-08-11
fix: normalize pointers for active entities

menduz:
## Description

Fixes active entities cache keys to always use lowercase instead of uppercase or a mix.


marianogoldman (approved):

aleortega (approved):

aleortega:
The first commit I pushed fixed the tests. The issue was that we were persisting the keys in the cache with uppercases while we were trying to retrieve them only with lowercases. Having said so, I thought that maybe it is worth updating all the trips around the cache to always use lowercases. Therefore I made a second commit to solve this but I'm not sure if this is something we want as well, since the tests were already working with the first commit.

marianogoldman (approved):

PR #1242, opened 2022-08-19
fix: the 60secs timeout of a malformed request

menduz:
Fixes the 60secs timeout of a malformed request manually handling exceptions

### DIFF
--- a/content/src/controller/Controller.ts
+++ b/content/src/controller/Controller.ts
@@ -52,69 +52,81 @@ export class Controller {
    * this endpoint will be deprecated in favor of `getActiveEntities`
    */
   async getEntities(req: express.Request, res: express.Response): Promise<void> {
-    // Method: GET
-    // Path: /entities/:type
-    // Query String: ?{filter}&fields={fieldList}
-    const type: EntityType = parseEntityType(req.params.type)
-    const pointers: string[] = this.asArray<string>(req.query.pointer as string)?.map((p) => p.toLowerCase()) ?? []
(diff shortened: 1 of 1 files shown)


pedrotambo (approved):

marianogoldman:
Just asking, but is it possible to do this treatment using a middleware? I imagine we want to to exactly the same for any endpoints...

menduz:
It should be possible.. but the http server impl uses a custom middleware enhancer/composer that is not unit tested. It requires so much testing effort and I can only afford fixing the issue at hand right now. Feel free to dismiss the PR

content/src/controller/Controller.ts
```diff
@@ -52,69 +52,81 @@ export class Controller {
...
+        ids.length > 0
+          ? await this.components.activeEntities.withIds(ids)
+          : await this.components.activeEntities.withPointers(pointers)
 
-    const maskedEntities: Entity[] = entities.map((entity) => ControllerEntityFactory.maskEntity(entity, enumFields))
-    res.send(maskedEntities)
+      const maskedEntities: Entity[] = entities.map((entity) => ControllerEntityFactory.maskEntity(entity, enumFields))
+      res.send(maskedEntities)
+    } catch (error) {
+      Controller.LOGGER.error(`POST /entities/active - Internal server error '${error}'`)
```
menduz, line 91:
```suggestion
      Controller.LOGGER.error(`POST /entities/:type - Internal server error '${error}'`)
```

marianogoldman (approved):

PR #1245, opened 2022-08-24
chore: use catalyst storage

menduz:
## Description

Uses https://github.com/decentraland/catalyst-storage to handle storage, compatible with S3


content/src/logic/delete-unreferenced-files.ts
```diff
@@ -11,46 +12,61 @@ export async function deleteUnreferencedFiles(
...
+      }
     }
+    logger.info(`Deleted ${numberOfDeletedFiles} files`)
+  } else {
+    logger.error('deleteUnreferencedFiles only works with FolderBasedContentStorage')
   }
-  logger.info(`Deleted ${numberOfDeletedFiles} files`)
+}
+
+function isFolderBasedStorage(x: any): x is FolderBasedContentStorage {
```
pedrotambo, line 69: Could we move this responsibility to the `catalyst-storage` ?

content/src/service/ServiceImpl.ts
```diff
@@ -1,10 +1,11 @@
+import { ContentItem } from '@dcl/catalyst-storage'
 import { AuthChain, Authenticator } from '@dcl/crypto'
 import { Entity, EntityType, IPFSv2 } from '@dcl/schemas'
 import { ILoggerComponent } from '@well-known-components/interfaces'
 import { EnvironmentConfig } from '../Environment'
 import { runReportingQueryDurationMetric } from '../instrument'
 import { calculateDeprecatedHashes, calculateIPFSHashes } from '../logic/hashing'
-import { bufferToStream, ContentItem } from '../ports/contentStorage/contentStorage'
+import { bufferToStream } from '@dcl/catalyst-storage/dist/content-item'
```
pedrotambo, line 8: Is it necessary to import from `.../dist/...`?

content/src/components.ts
```diff
@@ -76,7 +76,7 @@ export async function initComponentsWithEnv(env: Environment): Promise<AppCompon
 
   const catalystFetcher = FetcherFactory.create({ env })
   const contentFolder = path.join(env.getConfig(EnvironmentConfig.STORAGE_ROOT_FOLDER), 'contents')
-  const storage = await createFileSystemContentStorage({ fs }, contentFolder)
+  const storage = await createFolderBasedFileSystemContentStorage({ fs }, contentFolder)
```
pedrotambo, line 79: Could we move the `fs` interface + implementation to `catalyst-storage` too? (Or remove it within `catalyst-storage`, so it's not duplicated)

menduz: Yes, will do

content/src/service/ServiceImpl.ts
menduz: ts is nasty sometimes.. having a src + dist folders is the best way I've found to deal with conflicts. since bufferToStream and SimpleContentItem are internal utils for testing, those are not exported by index.ts

content/src/logic/delete-unreferenced-files.ts
menduz: which responsibility? the helper function?

pedrotambo: exactly

content/src/service/ServiceImpl.ts
pedrotambo: But if we're using them here, they're not only for internal utils 🤔

content/src/components.ts
pedrotambo: 👍

PR #1249, opened 2022-08-26
feat: deprecate old repository db

pedrotambo:
This PR deletes the old abstraction to make queries to the database using `pgPromise` and migrates the queries to the new postgres client. In addition, a new abstraction to make `transactions` was needed, so it was created as the `pg` package doesn't provide one.


content/src/Environment.ts
```diff
@@ -393,24 +392,6 @@ export class EnvironmentBuilder {
...
-      EnvironmentConfig.REPOSITORY_QUEUE_MAX_QUEUED,
-      () => process.env.REPOSITORY_QUEUE_MAX_QUEUED ?? RepositoryQueue.DEFAULT_MAX_QUEUED
-    )
-
-    this.registerConfigIfNotAlreadySet(
-      env,
-      EnvironmentConfig.REPOSITORY_QUEUE_TIMEOUT,
-      () => process.env.REPOSITORY_QUEUE_TIMEOUT ?? RepositoryQueue.DEFAULT_TIMEOUT
-    )
-
```
agusaldasoro, line 413: Is there an easy way to configure manually those values now? Those vars were used for troubleshooting in the catalyst owner

content/src/components.ts
```diff
@@ -160,7 +154,7 @@ export async function initComponentsWithEnv(env: Environment): Promise<AppCompon
   )
 
   const garbageCollectionManager = new GarbageCollectionManager(
-    { repository, deployer, systemProperties, metrics, logs, storage },
+    { deployer, systemProperties, metrics, logs, storage, database },
```
agusaldasoro, line 157: ❤️

content/src/instrument.ts
```diff
@@ -1,42 +1,4 @@
 import { ILoggerComponent } from '@well-known-components/interfaces'
```
agusaldasoro, line 1: Is this file being used now?

content/src/logic/database-queries/deployments-queries.ts
```diff
@@ -239,9 +248,150 @@ export async function getActiveDeploymentsByContentHash(
...
+          WHERE dep1.entity_type = ${entity.type} AND
+              dep1.entity_pointers && ${entity.pointers} AND
+              (dep1.entity_timestamp < to_timestamp(${entity.timestamp} / 1000.0) OR (dep1.entity_timestamp = to_timestamp(${entity.timestamp} / 1000.0) AND dep1.entity_id < ${entity.id})) AND
+              (dep2.id IS NULL OR dep2.entity_timestamp > to_timestamp(${entity.timestamp} / 1000.0) OR (dep2.entity_timestamp = to_timestamp(${entity.timestamp} / 1000.0) AND dep2.entity_id > ${entity.id}))
+          ORDER BY dep1.entity_timestamp DESC, dep1.entity_id DESC`
+    )
+  ).rows.map((row) => row.id)
+}
+
+export async function calculateOverwrittenByMany1(
```
agusaldasoro, line 360: is this name correct?

content/src/logic/database-queries/deployments-queries.ts
```diff
@@ -239,9 +248,150 @@ export async function getActiveDeploymentsByContentHash(
...
+  pointers.forEach((pointer) => q.append(pointer))
+  q.append(SQL`)
+          AND deployments.entity_type = ${entity.type}
+          AND (deployments.entity_timestamp > to_timestamp(${entity.timestamp} / 1000.0) OR (deployments.entity_timestamp = to_timestamp(${entity.timestamp} / 1000.0) AND deployments.entity_id > ${entity.id}))
+        ORDER BY deployments.entity_timestamp, deployments.entity_id
+        LIMIT 1`)
+  return (await database.queryWithValues<{ id: number }>(q)).rows
+}
+
+export async function calculateOverwrittenByMany2(
```
agusaldasoro, line 381: Is this name correct?

content/src/service/ServiceImpl.ts
```diff
@@ -63,10 +60,9 @@ export class ServiceImpl implements MetaverseContentService {
     files: DeploymentFiles,
     entityId: string,
     auditInfo: LocalDeploymentAuditInfo,
-    context: DeploymentContext,
-    task?: Database
+    context: DeploymentContext
   ): Promise<DeploymentResult> {
-    const deployedEntity = await this.getEntityById(entityId, task)
+    const deployedEntity = await getEntityById(this.components, entityId)
```
agusaldasoro, line 65: shouldn't this be a method from the component instead of sending the component as a parameter? Something like `this.components.database.getEntityById(entityId)`?

content/src/service/garbage-collection/GarbageCollectionManager.ts
```diff
@@ -47,24 +47,17 @@ export class GarbageCollectionManager {
...
-      await this.components.repository.tx(
-        async (transaction) => {
-          const { end: endTimer } = this.components.metrics.startTimer('dcl_content_garbage_collection_time')
+      const hashes = await findContentHashesNotBeingUsedAnymore(this.components, this.lastTimeOfCollection)
 
-          const hashes = await transaction.content.findContentHashesNotBeingUsedAnymore(this.lastTimeOfCollection)
+      this.components.metrics.increment('dcl_content_garbage_collection_items_total', {}, hashes.length)
 
-          this.components.metrics.increment('dcl_content_garbage_collection_items_total', {}, hashes.length)
+      this.LOGGER.debug(`Hashes to delete are: (${hashes.join(',')})`)
```
agusaldasoro, line 56: is debug logging working okay now? If not, would that be too much noise?

content/test/integration/E2ETestEnvironment.ts
```diff
@@ -49,7 +49,7 @@ export class E2ETestEnvironment {
       })
     })
     this.database = await createDatabaseComponent({ logs, env: this.sharedEnv, metrics })
-    this.database.start()
+    if (this.database.start) this.database.start()
```
agusaldasoro, line 52: is this correct? Shouldn't it be the other way out?

content/test/unit/service/Service.spec.ts
```diff
@@ -242,7 +232,7 @@ describe('Service', function () {
     const storage = new MockedStorage()
     const pointerManager = NoOpPointerManager.build()
     const authenticator = new ContentAuthenticator(new HTTPProvider("https://rpc.decentraland.org/mainnet?project=catalyst-ci"), DECENTRALAND_ADDRESS)
-    const database = await createDatabaseComponent({ logs, env, metrics })
+    // const database = await createDatabaseComponent({ logs, env, metrics })
```
agusaldasoro, line 235: should we delete this line?

content/src/Environment.ts
pedrotambo: 🤔  those values configured the old repository (now deleted). It's still possible to configure query timeouts for the new client tho. would that be enough?

content/src/instrument.ts
pedrotambo: yeah, there's still one remaining function

content/src/logic/database-queries/deployments-queries.ts
pedrotambo: will improve

content/src/service/ServiceImpl.ts
pedrotambo: That would follow the old pattern where the "`repository`" component implemented a method for each query. Currently we're following other pattern in all the queries inside `database-queries`.

content/src/service/garbage-collection/GarbageCollectionManager.ts
pedrotambo: It logs a considerable amount of text but every long period of time i think. I can change it if you think it's better

content/test/integration/E2ETestEnvironment.ts
pedrotambo: The `database` now implements the WKC `IBaseComponent` interface where `start` and `stop` are optional methods. I don't think there's another better way out to do that.

content/src/components.ts
pedrotambo: 🕺

content/src/Environment.ts
agusaldasoro: yes it is, as long as you can do it through `.env` file right?

content/src/service/ServiceImpl.ts
agusaldasoro: can we send only the necessary components as parameters then?

content/src/Environment.ts
pedrotambo: yes, it's possible with `PG_QUERY_TIMEOUT` and `PG_STREAM_QUERY_TIMEOUT`

agusaldasoro (approved):
Awesome job!

content/src/Environment.ts
agusaldasoro: can we document that in https://github.com/decentraland/catalyst/blob/main/content/README.md ?

content/src/service/ServiceImpl.ts
menduz: This is an emerging pattern in WKC, which is in line with tolerant readers https://martinfowler.com/bliki/TolerantReader.html: the functions only specify the components required to make it work but a broader _and possible more complete_ set of components can be passed on for clarity. the snapshots fetcher uses that a lot

PR #1289, opened 2022-09-15
break: fast bootstrapping

pedrotambo:
This PR refactors how a Content Server bootstraps from snapshots. Now, instead of having one big snapshot per node, it's splitter in multiple smaller snapshots. A Content Server remembers all the snapshots that it has processed in the past and doesn't process them again every time it restarts. For more details, please read the [RFC - Fast Bootstrapping](https://docs.google.com/document/d/1V8HUCbsnb7O8O03nUnpvDzu5OXia9KWU3qwQs8FLYC4/edit).

This PR should be read in conjunction with [this PR](https://github.com/decentraland/snapshots-fetcher/pull/68) from `snapshots-fetcher`.

Notes:
- Failed deployments are now persisted. This was done to simplify the logic of marking a snapshot as processed. Instead of waiting until all the failed deployments for a specific snapshot are deployed, we mark a snapshot as processed once it has already read all of its deployments inside. If someone failed, it is persisted and the deployment is tried afterward.
- A `Clock` component was introduced for testing purposes.
- The old snapshot generation is still generated to be compatible with catalysts outdated. Once all were migrated and some time passed, it will be removed.
- The old snapshot generation is now done once time on startup to reduce CPU consumption.
- The old snapshot generation was enforced to run before the new snapshot generation as there were race conditions when running at the same time. Sometimes, they try to save the empty snapshot at the same time and an exception is thrown in the `storage` (it happens mostly in tests).

Open questions:
- Should periodic sync of snapshots be introduced? Once a Catalyst is bootstrapped, it will generate daily snapshots every day, but it only reads snapshots from others on bootstrap. If periodic (daily ideally) sync of snapshots is introduced, it will be kept updated and if it restarts, it won't have to process multiple snapshots from other Catalyst (covering all the time it was `UP`). Moreover, if it reads the daily snapshots when it's `UP`, it will be benefited from the bloom filter of deployed entities (probably the entities were already deployed by the `pointer-changes` instant sync).


content/src/logic/database-queries/snapshots-queries.ts
```diff
@@ -33,3 +35,103 @@ export async function* streamActiveDeployments(
...
+  const query = SQL`
+  SELECT
+    hash,
+    date_part('epoch', init_timestamp)  AS "initTimestampSecs",
+    date_part('epoch', end_timestamp)  AS "endTimestampSecs",
+    replaced_hashes as replacedSnapshotHashes,
+    number_of_entities as numberOfEntities
+  FROM snapshots
+  WHERE to_timestamp(${timerange.initTimestampSecs}) <= init_timestamp
+  AND end_timestamp <= to_timestamp(${timerange.endTimestampSecs})
```
menduz, line 83: https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-between/

menduz: The committed version is inverted and shouldn't work heh that's why I suggested the BETWEEN operator to prevent this kind of typing errors

menduz: Oh nevermind those are different fields

content/src/logic/database-queries/snapshots-queries.ts
```diff
@@ -33,3 +35,182 @@ export async function* streamActiveDeployments(
...
+): Promise<void> {
+  if (processedSnapshotHashesToDelete.length == 0) {
+    return
+  }
+  const query = SQL`DELETE FROM processed_snapshots WHERE hash IN (`
+  const hashes = processedSnapshotHashesToDelete.map((h, i) =>
+    i < processedSnapshotHashesToDelete.length - 1 ? SQL`${h},` : SQL`${h}`
+  )
+  hashes.forEach((hash) => query.append(hash))
+  query.append(`);`)
```
menduz, line 196: it is about time that we create a helper for this 😂

content/src/ports/fileWriter.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const file = components.fs.createWriteStream(filePath)
+
+  const fileClosedFuture = new Promise<void>((resolve, reject) => {
+    file.on('finish', resolve)
+    file.on('end', resolve)
+    file.on('error', reject)
+  })
+
+  // the follosing lines exist to naively emulate a buffering algorithm to reduce disk IO
```
menduz, line 37:
```suggestion
  // the following lines exist to naively emulate a buffering algorithm to reduce disk IO
```

content/src/ports/snapshotGenerator.ts
```diff
@@ -0,0 +1,62 @@
...
+  let isStopped = false
+  let runningGeneration: Promise<void> = Promise.resolve()
+  let nextGenerationTimeout: NodeJS.Timeout
+  let currentSnapshots: NewSnapshotMetadata[]
+
+  async function runGenerationAndScheduleNext() {
+    isRunningGeneration = true
+    try {
+      currentSnapshots = await generateSnapshotsInMultipleTimeRanges(components, {
+        // IT IS IMPORTANT THIS TIMESTAMP NEVER CHANGES; IF IT DOES, THE WHOLE SNAPSHOTS SET WILL BE REGENERATED.
```
menduz, line 28: This is scary, why is it like that?

content/src/service/synchronization/batchDeployer.ts
```diff
@@ -110,14 +112,19 @@ export function createBatchDeployerComponent(
...
-            components.deployer.reportErrorDuringSync(
-              entity.entityType as any,
-              entity.entityId,
-              FailureReason.DEPLOYMENT_ERROR,
-              entity.authChain,
-              err.toString()
+            const errorDescription = err.toString()
+            logs.warn(
+              `Deployment of entity (${entity.entityType}, ${entity.entityId}) failed. Reason was: '${errorDescription}'`
             )
```
menduz, line 118: Prefer structured logs
```suggestion
            const errorDescription = err.toString()
            logs.warn(`Entity deployment failed`, {
              entityType: entity.entityType,
              entityId: entity.entityId,
              reason: errorDescription
            })
```

content/test/unit/logic/time-range.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 49), 'MWWIIIIIII')
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 50), 'MWWWI')
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 56), 'MWWWIIIIIII')
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 57), 'MWWWWI')
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 63), 'MWWWWIIIIIII')
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 64), 'MMWI')
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 364), 'MMMMMMMMMMMMWWWIIIIIII')
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 365), 'MMMMMMMMMMMMWWWWI')
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 370), 'MMMMMMMMMMMMWWWWIIIIII')
+  expectTimeDivisionMatchesRepresentation(timeRangeOfDays(1640995200000, 371), 'MMMMMMMMMMMMWWWWIIIIIII')
```
menduz, line 147: what is this? would you add some comments on the test?

content/src/service/Server.ts
```diff
@@ -100,6 +100,7 @@ export class Server implements IBaseComponent {
     this.registerRoute('/pointer-changes', controller, controller.getPointerChanges)
     this.registerRoute('/snapshot/:type', controller, controller.getSnapshot) // TODO: Deprecate
     this.registerRoute('/snapshot', controller, controller.getAllSnapshots)
```
agusaldasoro, line 102:
```suggestion
    this.registerRoute('/snapshot', controller, controller.getAllSnapshots) // TODO: Deprecate
```

content/src/logic/database-queries/snapshots-queries.ts
```diff
@@ -33,3 +35,182 @@ export async function* streamActiveDeployments(
...
+  INSERT INTO processed_snapshots
+  (hash, process_time)
+  VALUES
+  (${processedSnapshotHash}, to_timestamp(${processTimestampSecs} / 1000.0))
+  RETURNING hash
+  `
+  await database.queryWithValues(query, 'save_processed_snapshot')
+}
+
+export async function deleteProcessedSnapshots(
```
agusaldasoro, line 184: why you should ever need to delete a processed snapshot?

content/src/logic/database-queries/snapshots-queries.ts
```diff
@@ -33,3 +35,182 @@ export async function* streamActiveDeployments(
     }
   }
 }
+
+export async function* streamActiveDeploymentsInTimeRange(
+  components: Pick<AppComponents, 'database'>,
+  timeRange: TimeRange
+): AsyncIterable<DeploymentWithAuthChain> {
+  // IT IS IMPORTANT THAT THIS QUERY NEVER CHANGES. ORDER IS NOT GUARANTEED  `
```
agusaldasoro, line 43: why can't we guarantee the order?

content/src/logic/snapshots.ts
```diff
@@ -0,0 +1,107 @@
...
+import { divideTimeInYearsMonthsWeeksAndDays, intervalSizeLabel, isTimeRangeCoveredBy, TimeRange } from './time-range'
+
+export type NewSnapshotMetadata = {
+  hash: string
+  timeRange: TimeRange
+  numberOfEntities: number
+  replacedSnapshotHashes?: string[]
+}
+
+export async function generateAndStoreSnapshot(
```
agusaldasoro, line 19: the format of the old snapshot and new files is the same, right?

content/src/logic/time-range.ts
```diff
@@ -0,0 +1,109 @@
...
+ *    - timeRanges: [[t1, t1 + 6], [t1 + 5, t2]]
+ *  Result: True, even though as the intervals in timeRanges are overlapped, the sum of all the elements in timeRanges cover the interval of timerange.
+ *
+ * EXAMPLE 4
+ *  Input:
+ *    - timerange: [t1, t2]
+ *    - timeRanges: [[t1 - 1, t1 + 5], [t1 + 5, t2 + 3]]
+ *  Result: True, even though the sum of all the elements in timeRanges extends the limits of timerange, it strictly contains it.
+ */
+export function isTimeRangeCoveredBy(timerange: TimeRange, timeRanges: TimeRange[]) {
```
agusaldasoro, line 65: is this function only for testing?

content/src/logic/time-range.ts
```diff
@@ -0,0 +1,109 @@
...
+  for (const t of timeRanges) {
+    if (t.initTimestamp > currentMaxTimestamp) {
+      return false
+    }
+    currentMaxTimestamp = Math.max(currentMaxTimestamp, t.endTimestamp)
+  }
+  return minTimestamp <= timerange.initTimestamp && currentMaxTimestamp >= timerange.endTimestamp
+}
+
+export function divideTimeInYearsMonthsWeeksAndDays(timeRange: TimeRange): TimeRangeDivision {
```
agusaldasoro, line 78: can you add docs here?

content/src/ports/failedDeployments.ts
```diff
@@ -0,0 +1,78 @@
...
+import { EntityType } from '@dcl/schemas'
+import {
+  deleteFailedDeployment,
+  getFailedDeployments,
+  saveFailedDeployment
+} from '../logic/database-queries/failed-deployments-queries'
+import { AppComponents } from '../types'
+
+export enum FailureReason {
+  DEPLOYMENT_ERROR = 'Deployment error' // During sync, there was an error during deployment. Could be due to a validation
```
agusaldasoro, line 11: do we need this? It seems a little bit redundant as failed deployments only came from Sync

agusaldasoro (approved):
Excellent work!

content/src/logic/database-queries/snapshots-queries.ts
pedrotambo: when you're replacing 7 daily snapshots for a weekly snapshot for example

content/src/logic/database-queries/snapshots-queries.ts
pedrotambo: we can but the query could take more time

content/src/logic/snapshots.ts
pedrotambo: yes

content/src/logic/time-range.ts
pedrotambo: nope, it's also used in `snapshots.ts`

content/src/logic/time-range.ts
pedrotambo: yes

content/src/ports/failedDeployments.ts
pedrotambo: just copied from last implementation

PR #1303, opened 2022-09-27
fix: Support invalid urn

agusaldasoro:


lambdas/src/apis/collections/Utils.ts
```diff
@@ -25,8 +25,12 @@ export function preferEnglish(i18ns: I18N[]): string | undefined {
   return (i18nInEnglish ?? i18ns[0])?.text
 }
 
-export function translateEntityIntoWearable(client: SmartContentClient, entity: Entity): LambdasWearable {
-  const metadata: Wearable = entity.metadata!
+export function translateEntityIntoWearable(client: SmartContentClient, entity: Entity): LambdasWearable | undefined {
+  if (!entity.metadata || !entity.metadata.representations) {
```
pedrotambo, line 29: shouldn't this be `!(entity.metadata.data && entity.metadata.data.representations)` or something like that?

lambdas/src/apis/collections/controllers/emotes.ts
```diff
@@ -10,7 +10,7 @@ import { TheGraphClient } from '../../../utils/TheGraphClient'
 import { ItemFilters, ItemPagination, LambdasEmote } from '../types'
 import { translateEntityIntoEmote } from '../Utils'
 
-const LOGGER = log4js.getLogger('TheGraphClient')
+const LOGGER = log4js.getLogger('TheGraphClientEmotes')
```
pedrotambo, line 13: `EmotesHandler`?

lambdas/src/apis/collections/Utils.ts
agusaldasoro: That's logically the same, right?

pedrotambo (approved):

PR #1311, opened 2022-10-03
fix: emotes ownership

pedrotambo:


lambdas/src/utils/SmartContentClient.ts
```diff
@@ -112,8 +112,8 @@ export class SmartContentClient implements ContentAPI {
       try {
         const fetcher = new Fetcher()
         await fetcher.fetchJson(`${SmartContentClient.INTERNAL_CONTENT_SERVER_URL}/status`, {
-          attempts: 6,
-          waitTime: '10s'
+          attempts: 1,
```
agusaldasoro, line 115: should we undo this change?

pedrotambo: done

agusaldasoro (approved):

PR #1312, opened 2022-10-03
feat: persist failed deployments

pedrotambo:
This PR is for reviewing a small part of a bigger stuff. This won't be merged directly to `main`. It'll be merged to `feat/fast-bootstrapping` and then in the future will be reviewed as a whole.
It is for persisting failed deployments during sync.

After reading a snapshot file and deploying the entities inside it, we need to mark it as processed and save the snapshot hash in the database so in the next bootstrap the server doesn't have to re-processed the snapshot file.
During sync, when a deployment fails, it's queued in a `Map` to be retried in the future and continues with the next one. Currently, we cannot mark a snapshot as processed if there are failed deployments because they only live in memory and will be lost if the server restarts.

We decided to consider a snapshot as `processed` when all the file was read and its entities where tried to deploy and persist the failed deployments to be re-tried in the future. Why?
- The readability of the logic its way more clear. Adding logic to mark a snapshot as processed after being all the failed deployments in the `Map` successfully deployed adds a lot of coupling and nested code.
- There are cases were a deployment is successfully deployed in a server but when it's synced during bootstrap in the others servers, it fails. In that case, the snapshot would be re-processed every time the server restart and that could lead to an instant degrade in the bootstrapping time.


agusaldasoro (approved):
Nice job!

PR #1380, opened 2022-12-23
fix: pointers and ids retrieval from content by normalizing cases

aleortega:
## Description

It fixes how the content server stores and retrieves Entities by Ids or Pointers.

Reason of the issue: we had keys mixing uppercases and lowercases on the cache.

Also, it fixes a corner case where on (`/content/entities/active` endpoint):
* If you try to get an entity by Id, the client needs to send that Id always in lowercases
* If the client at some point miss parsing the Id to lowercase and send it with at least one letter in uppercase, then the cache breaks
* Now the client can send the id in lower or uppercases and the content server internally will treat them as they were always in lowercases, so the cache should not break again because of this

### DIFF
--- a/content/src/ports/activeEntities.ts
+++ b/content/src/ports/activeEntities.ts
@@ -246,8 +246,8 @@ export const createActiveEntitiesComponent = (
     clear,
 
     getCachedEntity: (idOrPointer) => {
-      if (cache.has(idOrPointer)) {
-        const cachedEntity = cache.get(idOrPointer)
+      if (cache.has(idOrPointer.toLowerCase())) {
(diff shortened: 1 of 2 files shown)


menduz (approved):
Changes look good, since no tests were broken for this change, would you add some tests to prevent accidentally breaking it?

aleortega:
> Changes look good, since no tests were broken for this change, would you add some tests to prevent accidentally breaking it?

Nice catch, I'm on it

menduz:
My mistake, some tests did fail!

aleortega:
> My mistake, some tests did fail!

Just realized it as well, I'm on it. I'll update the map implementation as well, thx!

marianogoldman (approved):

PR #1405, opened 2023-01-06
fix: save own snapshots as processed, so they dont get deleted

pedrotambo:


agusaldasoro (approved):

content/src/logic/database-queries/snapshots-queries.ts
```diff
@@ -116,6 +116,14 @@ export async function saveSnapshot(
   await database.queryWithValues(query, 'save_snapshot')
 }
 
+export async function hasSnapshot(components: Pick<AppComponents, 'database'>, snapshotHash: string): Promise<boolean> {
```
agusaldasoro, line 119: shouldnt this be something like `isLocalSnapshot`?

agusaldasoro: own snapshot

pedrotambo: own snapshot sound good

PR #1421, opened 2023-01-20
fix: memory leak while syncing from scratch

### LINKED ISSUE #1427: Content Server:  Memory Leaks while syncing from scratch
While developing the on-chain validations, some memory leaks were detected on the content server.

hugoArregui:
Fixes #1427

### DIFF
--- a/content/src/service/synchronization/batchDeployer.ts
+++ b/content/src/service/synchronization/batchDeployer.ts
@@ -114,8 +114,9 @@ export function createBatchDeployerComponent(
 
       const operationPriority = priorityBasedOnEntityType(entity.entityType)
 
-      parallelDeploymentJobs
-        .scheduleJobWithPriority(async () => {
+      try {
+        await parallelDeploymentJobs.onSizeLessThan(1000)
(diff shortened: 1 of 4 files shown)


content/src/service/synchronization/batchDeployer.ts
```diff
@@ -109,81 +109,88 @@ export function createBatchDeployerComponent(
...
-                components,
-                entity.entityId,
-                entity.entityType,
-                entity.authChain,
-                elementInMap.servers,
-                DeploymentContext.SYNCED
-              )
-              wasEntityProcessed = true
+      try {
+        await parallelDeploymentJobs.onSizeLessThan(1000)
```
menduz, line 116: I think this condition will render `operationPriority` useless, should this condition be "priority aware"?
```suggestion
        await parallelDeploymentJobs.onSizeLessThan(1000, operationPriority)
```

hugoArregui: I don't think so, this is just to put a bound on it, the 1000 elements have their priority.

marianogoldman (approved):

PR #1443, opened 2023-02-17
fix: unhandled when tx timeouts

pedrotambo:


content/src/ports/postgres.ts
```diff
@@ -184,6 +184,9 @@ export async function createDatabase(
          * No need to dispose the client.
          */
         const client: PoolClient = await pool.connect()
+        client.on('error', (err) => {
+          logger.error(`There was an error with the database connection: ${err.message}`)
```
menduz, line 188: This is an unrecoverable error, it should kill the service!

PR #1507, opened 2023-04-20
chore: intelligent bloom filter loading

pedrotambo:
This PR works in conjunction with the [snapshots-fetcher PR](https://github.com/decentraland/snapshots-fetcher/pull/122).

- Removes the loading of the bloomFiler with all the deployments from the database with only the last 20 minutes.
- Adds to the bloomFilter interface the ability to load deployments from the database for a certain time interval.
- Adds the warmup in the batchDeployer, loading the bloomFilter only with the intervals of the snapshots to be processed. (see snapshots-fetcher PR).


content/test/unit/logic/time-range.spec.ts
```diff
@@ -155,3 +155,106 @@ function timeRangeOfDays(timestampSecsStartingAt: number, numberOfDays: number):
...
+    expect(result).toEqual([newTimeRange(0, 4)])
+  })
+
+  /*
+  * A:        [      ]
+  * B:            [     ]
+  * C:              [  ]
+  * D:                    [ ]
+  * Result:   [         ] [ ]
+  */
```
menduz, line 244: I love this

content/src/logic/time-range.ts
```diff
@@ -104,3 +104,34 @@ export function divideTimeInYearsMonthsWeeksAndDays(timeRange: TimeRange): TimeR
     }
   }
 }
+
+export function joinOverlappedTimeRanges(timeRanges: TimeRange[]): TimeRange[] {
```
hugoArregui, line 108: would be possible to add a comment here explaining the expected result of the function?

hugoArregui: I also want to point out that there is a very good node library for handling all time stuff, in this case intervals: https://date-fns.org/v2.30.0/docs/Interval, not saying you should use it or that it has everything you need, but I remember the utilities were very expressive and the code using the library was very easy to follow

hugoArregui (approved):

content/src/logic/time-range.ts
pedrotambo: Will add a comment here and take a look that that lib 👍

PR #1516, opened 2023-04-25
feat: add getEntityImage, getEntityThumbnail and getERC721Entity endpoints

### LINKED ISSUE #1513: Migrate to content endpoints in which lambdas just transforms a content response
/lambdas/collections/contents/{urn}/thumbnail
/lambdas/collections/contents/{urn}/image
/lambdas/collections/standard/erc721/{chainId}/{contract}/{option}/{emission}

hugoArregui:
Fixes #1513

### DIFF
--- a/content/src/service/Server.ts
+++ b/content/src/service/Server.ts
@@ -87,16 +87,23 @@ export class Server implements IBaseComponent {
       '/entities/active/entity/:pointer/thumbnail',
       controller,
       controller.getEntityThumbnail,
+      HttpMethod.HEAD
+    )
+    this.registerRoute(
+      '/entities/active/entity/:pointer/thumbnail',
+      controller,
+      controller.getEntityThumbnail,
       HttpMethod.GET
     )
+    this.registerRoute('/entities/active/entity/:pointer/image', controller, controller.getEntityImage, HttpMethod.HEAD)
+    this.registerRoute('/entities/active/entity/:pointer/image', controller, controller.getEntityImage, HttpMethod.GET)
     this.registerRoute(
       '/entities/active/erc721/:chainId/:contract/:option/:emission',
       controller,
       controller.getERC721Entity,
       HttpMethod.GET
     )
-    this.registerRoute('/entities/active/entity/:pointer/image', controller, controller.getEntityImage, HttpMethod.GET)
-    this.registerRoute('/contents/:hashId', controller, controller.headContent, HttpMethod.HEAD) // Register before GET
+    this.registerRoute('/contents/:hashId', controller, controller.getContent, HttpMethod.HEAD) // Register before GET
     this.registerRoute('/contents/:hashId', controller, controller.getContent, HttpMethod.GET)
     this.registerRoute('/available-content', controller, controller.getAvailableContent)
     this.registerRoute('/audit/:type/:entityId', controller, controller.getAudit)
--- a/content/src/service/Server.ts
+++ b/content/src/service/Server.ts
@@ -98,7 +98,7 @@ export class Server implements IBaseComponent {
     this.registerRoute('/entities/active/items/:pointer/image', controller, controller.getEntityImage, HttpMethod.HEAD)
     this.registerRoute('/entities/active/items/:pointer/image', controller, controller.getEntityImage, HttpMethod.GET)
     this.registerRoute(
-      '/erc721/:chainId/:contract/:option/:emission',
+      '/erc721/:chainId/:contract/:option/:emission?',
       controller,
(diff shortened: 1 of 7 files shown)


content/src/service/Server.ts
```diff
@@ -83,7 +83,27 @@ export class Server implements IBaseComponent {
...
+    )
+    this.registerRoute('/entities/active/entity/:pointer/image', controller, controller.getEntityImage, HttpMethod.HEAD)
+    this.registerRoute('/entities/active/entity/:pointer/image', controller, controller.getEntityImage, HttpMethod.GET)
+    this.registerRoute(
+      '/entities/active/erc721/:chainId/:contract/:option/:emission',
+      controller,
+      controller.getERC721Entity,
+      HttpMethod.GET
+    )
+    this.registerRoute('/contents/:hashId', controller, controller.getContent, HttpMethod.HEAD) // Register before GET
```
menduz, line 106: What is the decision framework to decide that these endpoints are part of the content server instead of lamb2?

menduz (changes requested):

content/src/service/Server.ts
hugoArregui: The fact that this is purely a content transformation and the fact that is part of the protocol, since it's referenced by contracts and subgraphs

menduz: since those are all queries over the state of the content server, may I suggest to scope them under `/queries/`? It is to prevent tinting the content server content API with more ephemeral queries

hugoArregui: I'm open to the idea, but not quite sure to understand the reasoning here, I put them under /active/entities because it's the same query

pedrotambo: I'd like to add my two cents here:

The way I see it, the functionality of the `thumbnail` and `image` endpoints are not content transformations but a way to get content that is stored _as it is_ in the Content Server. It's just a shortcut to:
- wearable = GET /entities/active/pointer1
- GET /contents/wearable.metadata.image

That's why I don't think it's wrong to add the functionalities here. Also `thumbnail` and `image` only make sense for wearables/emotes, so I'd suggest an endpoint like `/entities/active/wearables/:pointer/thumbnail` and `/entities/active/wearables/:pointer/thumbnail` or one condensing both in `/entities/active/items/:pointer/thumbnail`.

With respect to the erc721 endpoint, I still have doubts about this one. It doesn't return things that are actually stored in the Content Server but something that can be calculable in a client. I think this transformation is part of the protocol but it is not very related to the content service. It might be handy to have the functionality in the same server for performance matters, but I'd put it in a separate endpoint and expose it through nginx in a different prefix than `/content`.

hugoArregui: I like your suggestion Peter, I changed it. Regarding the erc721 keep in mind it requires the entity itself, so I think this belongs here (I mean, ideally we won't have it and we would just fetch the entity from content, but since it's now part of the protocol because it's referenced in contracts), then we will need to keep it.

hugoArregui: @menduz do you want to explain the idea behind putting it under `/queries`?

hugoArregui: @pedrotambo @menduz explained to me that he wants the `/queries` prefix (or something else) to have a clear distinction between what is necessary for the sync protocol and what's needed for other stuff. So I took both of your ideas and I decided for: `/queries/items...` is that ok with you?

pedrotambo: After thinking it over carefully, I take back what I said before 🤣 . If someone wants to get a `thumbnail`/`image`, they have to get a `pointer`. To get a pointer they must have obtained the metadata beforehand. If you have the metadata, you have all the information to get the thumbnail => GET /contents/metadata.data.thumbnail. I can't imagine a use case where a client calls these endpoints without having obtained the metadata beforehand. I don't even think the `thumbnail`/`image` endpoints are protocol.
My order of preference is:
1. Deprecate: Implement in clients and set a sunset for the `thumbnail`/`image` APIs. ERC721 one is protocol but again not content stored in Content Server. 
2. Move to Lamb2: Document, educate (tell why it's unnecessary), and add functionality to CatalystClient.
3. Move to `/queries/items`.

hugoArregui: @pedrotambo (1) was also my preferred solution, but the problem is those 3 endpoints are referenced in smart contracts and indexed in subgraphs, that's why, sadly, I think we need to consider them protocol. We could move them to `lamb2` but I choose to keep them here because this is content really, if these endpoints were not in use we could just call active entities, they just serve the data in a different way.

PR #1548, opened 2023-05-15
[WIP] feat: refactor code to bump catalyst-client to version 17.1.0

aleortega:
WIP

## Description

Please include a summary of the change and which issue is fixed/closed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

Closes # (issue)

## Changes

Specify the behavior changes introduced by this pull request

i.e.

- Removed method X from class Y
- Return 400 instead of 500 in Z case
- Refactored class X into a new module

## Types of changes

What types of changes does your code introduce? Remove the lines that don't that apply:

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to change). If there's an API change, link the pull request in the [API spec repository](https://github.com/decentraland/catalyst-api-specs) and the accepted [DAO governance poll](https://governance.decentraland.org/)
- I have read the [**CONTRIBUTING**](https://github.com/decentraland/catalyst/blob/main/docs/CONTRIBUTING.md) document.
- My code follows the code style of this project.
- My change requires a change to the documentation.
- I have updated the documentation accordingly.
- I have added tests to cover my changes.
- All new and existing tests passed.


content/src/service/garbage-collection/GarbageCollectionManager.ts
```diff
@@ -78,7 +77,7 @@ export class GarbageCollectionManager {
   private waitUntilSyncFinishes(): Promise<void> {
     return new Promise(async (resolve) => {
       while (this.sweeping === true) {
-        await delay('1s')
+        await setTimeout(() => {}, 1000)
```
hugoArregui, line 80: not quite sure if this is right, it seems the classic timeout, not the one that returns the promise

content/src/helpers/FetcherFactory.ts
```diff
@@ -1,18 +1,24 @@
-import { Fetcher } from 'dcl-catalyst-commons'
+import { createFetchComponent } from '@well-known-components/fetch-component'
+import { IFetchComponent, RequestOptions } from '@well-known-components/interfaces'
+import ms from 'ms'
+import * as fetch from 'node-fetch'
```
hugoArregui, line 4: should we still import this?

pedrotambo (approved):

content/src/helpers/FetcherFactory.ts
aleortega: We do need it to decorate the fetcher, mostly because the Response (I don't want to cast that Response to another type). Anyways, these RequestInit and Response types could be exported from `interfaces` or even from `fetch-component`. Both ways, we will have node-fetch as a dep or chained dep.

The scope of this PR is just to stop using custom fetchers (and bumps new catalyst-client), I'm still thinking that there is room for improvements, for example: I think we don't need a FetcherFactory and we should support default options when creating the component fetcher. But this change will require a new version of fetch-component that I can release afterwards while integrating the latest version of `catalyst-client` (which currently is outdated by 3 major updates, that's why my eager to merge this PR :).

Please let me know if you think we should take care of these improvements on this PR.

hugoArregui: this is fine, I just though that interfaces was exposing this as well, not sure why is not doing so.

hugoArregui (approved):

content/src/helpers/FetcherFactory.ts
aleortega: I'll add these types there, thank you!

PR #1553, opened 2023-05-17
fix: array query params handling to support param[] format

aleortega:
## Description

This PR adds support to different array query params formats:
* param=1&param=2 => [1, 2]
* param[]=1&param[]=2 => [1,2]
* param[1]=1&param[0]=2 => [2,1]


content/src/logic/query-params.ts
```diff
@@ -1,4 +1,23 @@
+import { isArray } from 'eth-connect'
```
marianogoldman, line 1: I think we need `Array.isArray()` here.

marianogoldman (approved):

marianogoldman (approved):

PR #1570, opened 2023-05-25
refactor: EntityFileFactory

aleortega:
## Description

This PR refactors `EntityFactory` to avoid exposing a static class in charge of re-creating entities. Instead, it became a module that exports a single function.


marianogoldman (approved):

content/src/logic/entity-parser.ts
```diff
@@ -0,0 +1,66 @@
+import { EntityContentItemReference } from '@dcl/hashing'
+import { Entity, EntityType } from '@dcl/schemas'
+import { AnyObject } from '../types'
+
+function getObjectEntityFromBuffer(buffer: Uint8Array): AnyObject {
+  try {
+    return JSON.parse(new TextDecoder().decode(buffer))
```
hugoArregui, line 7: I would instantiate the TextDecoder once at the top of the file and reuse it, so we only do it once

PR #1571, opened 2023-05-25
refactor: DAOClient and its factories

aleortega:
## Description

This PR refactors the DAO related Factories and re-implement their usage using plain functions.


content/src/ports/dao-servers-getter.ts
```diff
@@ -0,0 +1,57 @@
...
+    getAllContentServers,
+    getAllServers
+  }
+}
+
+export function createDAOComponent(components: Pick<AppComponents, 'env' | 'l1Provider'>): DAOComponent {
+  const { env, l1Provider } = components
+  const customDAO: string = env.getConfig(EnvironmentConfig.CUSTOM_DAO) ?? ''
+
+  if (usingCustomDAO(customDAO)) {
```
hugoArregui, line 40: nitpick but maybe instead of this, in components we can check the configuration and create a customDaoImplementation there, instead this component

hugoArregui (approved):

PR #1729, opened 2024-01-30
chore: clear stale profiles

hugoArregui:
# Approach 

## GC
The GC process will run no matter the status of the GARBAGE_COLLECTION flag. It will:

1. it will clean the old profiles active entities, which means:
   1. it will delete the old profiles (+1year) from the active entities cache
   2. it will delete the old profiles (+1year) from the active pointers table
2. if GARBAGE_COLLECTION is enabled, it will remove the old profiles deployments, the content_files's table entries for those deployments, and the files themselves if they are not used by a newer profile (<1 year)
3. if GARBAGE_COLLECTION is enabled, it will run the old GC as well
   
## Sync

No old profiles will be synced, even if they are included in a snapshot


content/src/service/garbage-collection/GarbageCollectionManager.ts
```diff
@@ -1,11 +1,25 @@
...
 import { findContentHashesNotBeingUsedAnymore } from '../../logic/database-queries/content-files-queries'
 import { SYSTEM_PROPERTIES } from '../../ports/system-properties'
-import { AppComponents } from '../../types'
+import { AppComponents, PROFILE_DURATION } from '../../types'
+import SQL from 'sql-template-strings'
+
+const PROFILE_CLEANUP_LIMIT = 10000
+
+export type GCStaleProfilesResult = {
+  hashesDeleted: Set<string>
```
marianogoldman, line 10:
```suggestion
  deletedHashes: Set<string>
```

content/src/service/garbage-collection/GarbageCollectionManager.ts
```diff
@@ -1,11 +1,25 @@
...
+  deploymentsDeleted: Set<string>
```
marianogoldman, line 11:
```suggestion
  deletedDeployments: Set<string>
```

marianogoldman (commented):
All looks good!

marianogoldman (approved):

PR #1768, opened 2024-04-23
chore: use GH actions

hugoArregui:


Dockerfile
```diff
@@ -0,0 +1,58 @@
+FROM node:16-alpine as base
```
marianogoldman, line 1:
```suggestion
FROM node:18-alpine as base
```

hugoArregui: I would do this as part of this PR, but I agree we should update this

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+  push:
+    branches:
+      - 'main'
+  pull_request:
+
+jobs:
+  validations:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
marianogoldman, line 13:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Use Node.js 18.x
+        uses: actions/setup-node@v3
```
marianogoldman, line 15:
```suggestion
        uses: actions/setup-node@v4
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+          POSTGRES_PASSWORD: 12345678
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+    steps:
+      - uses: actions/checkout@v3
```
marianogoldman, line 43:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Use Node.js 16.16.0
+        uses: actions/setup-node@v3
```
marianogoldman, line 45:
```suggestion
        uses: actions/setup-node@v4
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          node-version: 16.16.0
```
marianogoldman, line 47: Doesn't this work with node 18?

marianogoldman (approved):

PR #1796, opened 2024-08-05
feat: third party ownership validation

marianogoldman:


content/src/logic/third-party-contract-registry.ts
```diff
@@ -0,0 +1,128 @@
...
+      const r = await sendSingle(provider, await contract.balanceOf.toRPCMessage(contract.address, 0))
+
+      if (!r.result) {
+        return false
+      }
+      return !!contract.balanceOf.unpackOutput(toData(r.result))
+    } catch (_) {
+      return false
+    }
+  }
```
kevinszuchet, line 99: WDYT about creating a new function `checkContractType` (maybe) with the shared logic? (non-blocker IMO)

content/src/logic/third-party-item-checker.ts
```diff
@@ -0,0 +1,110 @@
...
+    result.forEach((r: any, idx: number) => {
+      if (!r.result) {
+        filteredAssets[idx].result = false
+      } else {
+        const data = toData(r.result)
+        if (thirdPartyContractRegistry.isErc721(filteredAssets[idx].contract!)) {
+          filteredAssets[idx].result =
+            (data === '0x' ? '' : contracts[idx].ownerOf.unpackOutput(data).toLowerCase()) === ethAddress.toLowerCase()
+        } else if (thirdPartyContractRegistry.isErc1155(filteredAssets[idx].contract!)) {
+          filteredAssets[idx].result = (data === '0x' ? 0 : contracts[idx].balanceOf.unpackOutput(data)) > 0
```
kevinszuchet, line 99: Would you mind creating a const storing the value '0x' 🫣? I think it could help with readability.

kevinszuchet (approved):

content/src/logic/third-party-contract-registry.ts
marianogoldman: I'm not really sure there's much to be reused... other than the try-catch, the rest has subtle differences in each line. Open to a suggestion, may be I'm not seeing it.

kevinszuchet: Yes, I think it's not worth the new abstraction. I thought they shared much more logic, but it's mostly code.

This was the only thing that I could think of:

```ts
 async function checkContractType(
  contractFactory: ContractFactory,
  contractAddress: ContractAddress,
  responseValidator: (response: any, contract: any, method: string) => boolean,
  method: string,
  args: any[]
): Promise<boolean> {
  try {
    const contract = await contractFactory.at(contractAddress);
    const response = await sendSingle(
      provider,
      await contract[method].toRPCMessage(...args)
    );

    return responseValidator(response, contract, method);
  } catch (error) {
    console.error(`Error checking contract type for ${contractAddress}`, error);
    return false;
  }
}

function defaultValidation(
  response: any,
  contract: any,
  method: string
): boolean {
  if (!response.result) return false;
  return !!contract[method].unpackOutput(toData(response.result));
}

async function checkIfErc721(
  contractFactory: ContractFactory,
  contractAddress: ContractAddress
): Promise<boolean> {
  const responseValidator: (
    response: any,
    contract: any,
    method: string
  ) => boolean = (response, contract) => {
    // Custom check for ERC721 (if any)
    if (response.error?.code === 3) {
      return true; // NFT id doesn't exist, but it's still an ERC721
    }
    return defaultValidation(response, contract, "ownerOf");
  };

  return checkContractType(
    contractFactory,
    contractAddress,
    responseValidator,
    "ownerOf",
    [0]
  );
}

async function checkIfErc1155(
  contractFactory: ContractFactory,
  contractAddress: ContractAddress
): Promise<boolean> {
  return checkContractType(
    contractFactory,
    contractAddress,
    defaultValidation,
    "balanceOf",
    [contractAddress, 0]
  );
}

```

Definitely not worth the added complexity 🙏

PR #1815, opened 2024-10-10
fix: profiles sync deployment

aleortega:
This PR bumps `content-validator` and `snapshot-fetcher` dependencies version to latest. Basically what this PR pretends to do is to validate the hash of the deployment files before admitting the deployment on the Catalyst.
This validation was already part of the `catalyst-client` library but was never included on the server.


braianj (approved):
LGTM

PR #1829, opened 2024-11-13
feat: set fixed cache for active scenes and add all scenes getter

leanmendoza:
This change introduces a handler to retrieve all the cached scenes. Also, the scene entities are pre-cached at the start and unaffected by the LRU logic.

### DIFF
--- a/content/test/unit/ports/activeEntities.spec.ts
+++ b/content/test/unit/ports/activeEntities.spec.ts
@@ -332,6 +332,62 @@ describe('activeEntities', () => {
       expect(firstResult).toMatchObject(fourthResult)
     })
   })
+
+  describe('after initialization', () => {
+    it('should populate cache with scenes during initialization', async () => {
+      const components = await buildComponents()
+      const sceneDeployment = {
+        ...fakeDeployment,
+        entityType: EntityType.SCENE,
+        entityId: 'scene1'
+      }
+
+      sut.mockImplementation(() => Promise.resolve([sceneDeployment]))
+
+      await components.activeEntities.initialize(components.database)
+
+      const cachedScenes = components.activeEntities.getAllCachedScenes()
+      expect(cachedScenes).toHaveLength(1)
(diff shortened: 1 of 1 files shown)


content/src/ports/activeEntities.ts
```diff
@@ -94,6 +101,47 @@ export function createActiveEntitiesComponent(
...
+  })
+  const fixedCache = new Map<string, Entity | NotActiveEntity>()
+
+  const createLRUandFixedCache = (maxLRU: number, fixedTypes: EntityType[]) => {
+    return {
+      get(entityId: string) {
+        return cache.get(entityId) || fixedCache.get(entityId)
+      },
+      set(entityId: string, entity: Entity | NotActiveEntity) {
+        const isFixed = fixedCache.has(entityId) || (typeof entity !== 'string' && fixedTypes.includes(entity.type))
```
aleortega, line 115: Ain't the second part of the condition enough to use the fixedCache? (I mean if entity type is SCENE we should already store it regardless if it is already stored there, right?)

content/src/ports/activeEntities.ts
```diff
@@ -322,7 +385,27 @@ export function createActiveEntitiesComponent(
...
+      }
+      logger.info('Active entities cache initialized')
+    } catch (error) {
+      logger.error('Error initializing active entities cache', { error })
+    }
+  }
+
+  function getAllCachedScenes() {
+    return Array.from(fixedCache.values()).filter(
+      (entity): entity is Entity => typeof entity !== 'string' && entity.type === EntityType.SCENE
```
aleortega, line 407: I'm not familiar with this check, but can we prevent `typeof entity !== string` on setting time instead of retrieve time?

leanmendoza: The LRU cache was already with this workaround for unsetting pointers with no longer active entities. I haven't had the intention of modifying this current behavior

PR #1882, opened 2026-04-03
feat: add HTTP range request support for content serving

LautaroPetaccio:
## Why

The [unity-explorer](https://github.com/decentraland/unity-explorer) relies on QuickTime for video playback on macOS. QuickTime requires servers to support HTTP range requests — it reads the MP4 `moov` atom from the end of the file via suffix ranges (`bytes=-N`), then uses `Content-Range` totals to calculate seek offsets. Without this, video textures fail to load (e.g., animated banners in the game arena appear black on macOS).

## What changed

**Range request handling** — The content, entity image, and entity thumbnail handlers now support HTTP range requests per RFC 9110:
- `bytes=start-end` and `bytes=start-` (standard ranges)
- `bytes=-N` (suffix ranges, used by QuickTime to read MP4 trailers)
- Returns **206 Partial Content** with correct `Content-Range` and `Content-Length` for valid ranges
- Returns **416 Range Not Satisfiable** with `Content-Range: bytes */size` for unsatisfiable ranges
- Falls back to **200** for malformed or unrecognized range formats (RFC-compliant)

**CORS headers** — `Access-Control-Expose-Headers` now includes `Content-Range`, `Accept-Ranges`, and `Content-Length` on both 206 and 416 responses, so browser-based players can read them.

**Correct total size for gzip-stored files** — Uses `contentSize` from `fileInfo()` (the uncompressed content size, added in [catalyst-storage#96](https://github.com/decentraland/catalyst-storage/pull/96)) for range validation and `Content-Range` totals. Falls back to `size` for backward compatibility. This prevents incorrect totals and silent data truncation when serving ranges from gzip-compressed files (used for content synced from other Catalyst nodes).

**`Accept-Ranges: bytes`** — Advertised in all content response headers so clients know range requests are supported.


content/src/controller/handlers/get-entity-thumbnail-handler.ts
```diff
@@ -22,14 +21,28 @@ export async function getEntityThumbnailHandler(
...
+  if (!result) {
     throw new NotFoundError('Entity has no thumbnail.')
   }
 
+  if (result.status === 416) {
+    return {
+      status: 416,
+      headers: result.rangeHeaders
+    }
+  }
```
braianj, line 35: should we return result + range headers?

content/package.json
```diff
@@ -23,7 +23,7 @@
     "@dcl/block-indexer": "^1.1.2",
     "@dcl/catalyst-api-specs": "^3.8.0",
     "@dcl/catalyst-contracts": "^4.4.2",
-    "@dcl/catalyst-storage": "^4.3.1",
+    "@dcl/catalyst-storage": "https://sdk-team-cdn.decentraland.org/@dcl/catalyst-storage/branch/feat/add-content-size-to-file-info/dcl-catalyst-storage-4.4.1-24264600601.commit-1b3a4b1.tgz",
```
braianj: just a reminder

braianj (approved):
👍🏽

content/src/controller/handlers/get-entity-thumbnail-handler.ts
LautaroPetaccio: The HTTP 416 status code means `Range Not Satisfiable`. There's no result for it. The client has requested a range that is not correct.

PR #1892, opened 2026-04-13
feat: add If-None-Match conditional request support for content endpoints

LautaroPetaccio:
## Summary

The three content-serving handlers (`get-content-handler`, `get-entity-image-handler`, `get-entity-thumbnail-handler`) set `ETag` and `Cache-Control` headers on every response, but never checked the `If-None-Match` request header. This meant clients couldn't use conditional GET requests — every repeat request re-downloaded the full content even when both sides knew it hadn't changed.

Each handler now checks `If-None-Match` against the computed ETag after building the response headers. When they match, the handler returns `304 Not Modified` with headers but no body, saving bandwidth for clients that cache content.

## Test plan

- [ ] Added unit tests covering: matching ETag returns 304 with no body, non-matching ETag returns full response, absent header returns full response
- [ ] Verify with `curl -H 'If-None-Match: "Qm..."' http://localhost:6969/contents/Qm...` returns 304


kevinszuchet (approved):
Nice! LGTM

PR #1902, opened 2026-04-14
fix: validate CID format in images handler

LautaroPetaccio:
## Summary

- Add input validation for the `cid` route parameter in the lambdas image resize handler to prevent path traversal attacks

## Why

The `cid` parameter from `req.params` was interpolated directly into a filesystem path (`${storageLocation}/${cid}_${size}`) and into an upstream content server URL (`/contents/${cid}`) without any format check. A crafted CID containing path separators (e.g. `../../`) could escape the storage directory during file reads/writes, and could manipulate the upstream fetch URL.

## How

IPFS CIDs use strictly alphanumeric character sets — base58 for CIDv0 (`Qm...`) and base32/base36 for CIDv1 (`bafy...`). A simple `^[a-zA-Z0-9]+$` regex rejects any input containing path separators, dots, or other special characters before the value reaches any I/O code path. The validation reuses the existing `ServiceError` pattern to return a 400 response on invalid input.

## Test plan

- [ ] `GET /images/<valid-cid>/256` continues to work as before
- [ ] `GET /images/../../etc/passwd/256` returns 400 "Invalid CID"
- [ ] `GET /images/abc..def/256` returns 400 "Invalid CID"

### DIFF
--- a/lambdas/src/apis/images/controllers/images.ts
+++ b/lambdas/src/apis/images/controllers/images.ts
@@ -12,14 +12,8 @@ const LOGGER = log4js.getLogger('ImagesController')
 
 const validSizes = ['128', '256', '512']
 
-// CIDv0 (Qm + base58) and CIDv1 (bafy + base32) are strictly alphanumeric
-const CID_PATTERN = /^[a-zA-Z0-9]+$/
-
-function validateCid(cid: string) {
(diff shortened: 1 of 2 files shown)


kevinszuchet (approved):
LGTM!

PR #1923, opened 2026-05-01
refactor: wkc layout — phase 13 (drain non-cache ports)

LautaroPetaccio:
## Summary

- migrates the three legacy `src/ports/` files that don't depend on an in-memory cache out into proper wkc adapter folders.
- repoints all import sites in `src/` and `test/` and relocates the postgres / snapshot-storage / snapshot-generator specs to mirror the new layout.
- defers `failedDeployments.ts` and `processedSnapshotStorage.ts` to the next slice — they depend on the `@dcl/memory-cache-component` minor bump (decentraland/core-components#88).

## Migration map

| legacy | new |
| --- | --- |
| `ports/postgres.ts` | `adapters/database/` (component + types + index) |
| `ports/snapshotStorage.ts` | `adapters/snapshot-storage/` (component + types + index) |
| `ports/snapshotGenerator.ts` | `adapters/snapshot-generator/` (component + types + index) |

`createTestDatabaseComponent` moves to `test/mocks/database-component-mock.ts` so the production database adapter no longer ships a test stub. `IDatabaseComponent`, `DatabaseClient`, `DatabaseTransactionalClient` live in `adapters/database/types.ts`.

## After this PR

`src/ports/` contains only `failedDeployments.ts` and `processedSnapshotStorage.ts`. Both are cache-backed and will move once `@dcl/memory-cache-component` ships a configurable cap/ttl (the upstream pr is decentraland/core-components#88).

## Test plan

- [x] `tsc --noEmit` clean
- [x] `npm run test:unit` — 233/233 passing locally
- [ ] CI green
- [ ] integration suite green


LautaroPetaccio:
Pushed `5a494d2` addressing both P2 comments:

**(1) Two overlapping test DB helpers** — added a JSDoc block at the top of `test/mocks/database-component-mock.ts` calling out which helper fits which test shape. `createDatabaseMockedComponent` is for tests that spy on / stub a specific call (`database.transaction.mockImplementation(...)`), `createTestDatabaseComponent` is for tests that wire the database in without expecting to touch it — the throw-on-call shape surfaces unintended use loudly instead of resolving to `undefined`.

**(2) Non-relative imports** — normalized both occurrences:
- `src/logic/entities.ts`: `'src/adapters/database'` → `'../adapters/database'`. While there I also dropped a redundant `'../logic/active-entities'` (the file is already inside `logic/`) for `'./active-entities'`.
- `test/helpers/logic/server-validator/NoOpValidator.ts`: `'src/logic/server-validator'` → `'../../../../src/logic/server-validator'`, matching the depth of the `AppComponents` import on the line below.

`grep "from 'src/"` over `src/` and `test/` now returns zero matches. 233/233 unit tests still passing locally.

PR #1924, opened 2026-05-01
refactor: wkc layout — phase 14 (drain final ports + reporter)

LautaroPetaccio:
> Stacked on #1923. Once that merges this can rebase onto main.

## Summary

- migrates the last two legacy `src/ports/` files (`failedDeployments`, `processedSnapshotStorage`) into proper wkc adapter folders.
- introduces `logic/failed-deployments-reporter/` to own the `database.transaction(...)` for the snapshot-failure-already-cached path.
- removes `src/ports/` entirely.

## Highlights

- **`adapters/failed-deployments-cache/`** is backed by `@dcl/memory-cache-component@^2.3.0` configured with `{ max: 1_000_000, ttl: 0 }` so the cache stays a complete mirror of the `failed_deployments` table for the lifetime of the process. `getAllFailedDeployments` enumerates the cache, so implicit ttl/lru eviction would be a regression — the cache-mirror unit test guards against it.
- **`adapters/processed-snapshot-storage/`** uses default `@dcl/memory-cache-component` options. Cache misses already fall through to the repository, so lru / 1 h ttl is fine. `reset()` is now async to match the cache component api; `resetServer` awaits accordingly.
- **`logic/failed-deployments-reporter/`** owns the transaction. Cache adapters no longer call `database.transaction(...)`; repos honor the wkc rule that logic owns transactions and adapters call repos with a `db: DatabaseClient` per call.
- **`batch-deployer`**, **`logic/deployments`**, and **`retry-failed-deployments`** are updated to inject `failedDeploymentsReporter` and call `reportFailure` through it.

## Test layout

- `test/unit/adapters/failed-deployments-cache/` — cache spec rewritten against the testing standards (describe-per-context, beforeEach for inputs, scoped mocks). Adds a "many warmed deployments" test that asserts no implicit eviction.
- `test/unit/logic/failed-deployments-reporter/` — new spec covering the three branches (snapshot+cached → tx, snapshot+not-cached → no tx, non-snapshot → cache only).
- `test/integration/adapters/failed-deployments-cache.spec.ts` and `test/integration/logic/failed-deployments-reporter.spec.ts` — equivalents against a real database.
- `test/integration/ports/` and `test/unit/ports/{failedDeployments,processedSnapshotStorage}.spec.ts` removed.

## After this PR

`src/ports/` does not exist. `find src test -name '*.ts' | xargs grep -l "ports/"` returns only the unrelated `test/unit/ports/contentStorage/` test fixture path.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `npm run test:unit` — 248/248 passing locally (up from 233 in #1923 thanks to the new specs)
- [ ] CI green
- [ ] integration suite green


LautaroPetaccio:
Pushed `26c8001c` addressing comments 1 and 2:

**(1) Async-heavy `getAllFailedDeployments` & friends.** Switched the failed-deployments cache to hash storage in `@dcl/memory-cache-component`. A single hash (key = `'failed-deployments'`, fields = entityIds) holds every cached deployment. As a result:

- `getAllFailedDeployments` is now `Object.values(await cache.getAllHashFields(...))` — one async call.
- `findFailedDeployment` / `removeFailedDeployment` / `cacheFailedDeployment` are still O(1) per call but go through `getFromHash` / `removeFromHash` / `setInHash` instead of round-tripping through `keys()`.
- `observeSize` reuses the hash payload (`Object.keys(all).length`) instead of doing a separate `keys()` lookup.

**(2) `reset()` in processed-snapshot-storage.** I checked the `@dcl/memory-cache-component` API — it has `get/set/remove/keys`, hash ops, and lock primitives, but no `clear()`. Rather than open another upstream bump, I applied the same hash-storage trick: all processed snapshot hashes are fields of one cache hash, and `reset()` is a single `cache.remove(PROCESSED_SNAPSHOTS_HASH)`. As a side benefit, `filterProcessedSnapshotsFrom` is one `getAllHashFields` + sync `in`-check across the input array instead of N parallel `get`s. Cache-miss fall-through to the repository is preserved.

**(3) 3-file pattern on processed-snapshot-storage.** I'd actually like to keep this one. The split is mandatory in this repo — `content/CLAUDE.md` makes the four-file layout a hard rule for every adapter and logic component, no exceptions for thin pass-throughs. Other thin adapters (e.g. `snapshot-storage`, with the same shape) already follow it; carving out an exception here would create the more confusing inconsistency the convention exists to prevent. Open to revisit at the convention level if the project decides bare-bones adapters should collapse to one file, but I'd rather not do that ad-hoc on this PR.

248/248 unit tests still passing locally — the existing specs cover the abstract behavior so the storage-shape swap is transparent.

LautaroPetaccio:
applied the same fix across the other five repositories. dropped the bare exports of the query functions in `active-entities-repository`, `pointers-repository`, `content-files-repository`, `snapshots-repository`, and `deployments-repository` — each `index.ts` now exposes only the factory and types. consumers now obtain the repository via dependency injection (`pick<appcomponents, 'xrepository'>`) and call methods on the instance instead of importing query functions with a database argument.

`deployments-repository` also dropped `getHistoricalDeploymentsQuery` from the interface — it is a pure sql builder, not i/o, and stays as a bare export alongside `createOrClause`.

call sites widened their `pick` deps to thread the new repos through: `deployment-service`, `deployments`, `snapshots`, `active-entities`, `garbage-collection`, `batch-deployer`, `pointer-manager`, `processed-snapshot-storage`, `snapshot-storage`, `snapshot-generator`, `failed-deployments-cache`, `deployed-entities-bloom-filter`, `retry-failed-deployments`, and the audit / deployments / pointer-changes / active-entities-by-hash controllers. `maintenancecomponents` and the `run-maintenance` / `run-content-files-fixer` entrypoints carry the new repos through. tests rewired to mock the repository interfaces instead of module-level spies.

unit tests pass (248/248) and tsc is clean.

PR #1937, opened 2026-06-11
refactor: native-fetch migration (@dcl/http-server v2) + storage bump + upload hardening

### LINKED ISSUE #1935: POST /entities/active: arrays not capped (unauthenticated DoS)
## Summary
`src/controllers/handlers/active-entities-handler.ts` validates `ids`/`pointers` with Joi `min(1)` but **no `max(...)`**. The endpoint is public and large arrays trigger costly `ANY(...)` queries; there is no app-level rate limiting.

## Impact
Unauthenticated resource exhaustion via huge `ids`/`pointers` arrays.

## Fix
Add `.max(N)` (e.g. 1000) to the Joi schema for both fields.

### LINKED ISSUE #1936: Unbounded loop DoS: POST /entities AuthChain index parsing iterates to attacker-controlled max
## Summary
`create-entity-handler.ts` derives `biggestIndex` from the AuthChain form field names (regex over `authChain[N][...]`) and then loops `for (let i = 0; i <= biggestIndex; i++)` (around line 122) with no cap. `N` is fully attacker-controlled via the multipart field names on the public `POST /entities` endpoint.

## Impact
A request with a field like `authChain[2000000000][type]` forces a multi-billion iteration loop, blocking the event loop — unauthenticated CPU DoS (the loop runs before signature validation completes).

## Fix
Cap `biggestIndex` to a small maximum (the AuthChain is normally 2-3 links; reuse the `maxItems: 10` bound from @dcl/schemas) and reject larger indices before looping.

LautaroPetaccio:
## What

Removes `node-fetch` from the content server's source and outbound HTTP path by migrating to native `fetch` and adopting `@dcl/http-server` v2, and bundles two related updates that touch the same files (a `@dcl/catalyst-storage` bump and a round of upload/security hardening).

This is intentionally one PR because the changes are tangled across shared files (`components.ts`, `package.json`, `Environment.ts`, `multipart.ts`).

## Native fetch migration

- Outbound HTTP now uses **`@dcl/fetch-component`** (native `fetch`) via the bumped **`@dcl/traced-fetch-component`**. The direct `@well-known-components/fetch-component` dependency is removed; `node-fetch` becomes a **dev-only** dependency (still pulled transitively by `@dcl/snapshots-fetcher`, pinned to `2.7.0` via `resolutions`).
- Adopt **`@dcl/http-server` v2** (`^2.0.0`), which serves **native `Request`/`Response`**. HTTP context types are now sourced from **`@dcl/core-commons`** (matching v2) instead of `@well-known-components/interfaces`, and the multipart wrapper adapts the request's web `ReadableStream` via `Readable.fromWeb` before piping to busboy.
- `@dcl/crypto` → `^3.7.0` (drops its `node-fetch` dep); `@types/node` → `^24`.
- Integration tests migrated off `node-fetch` to native `fetch` (`res.arrayBuffer()`, a Jest test environment exposing Node 24's web globals, and a `node-fetch`-backed fetcher kept only for the `dcl-catalyst-client` test helper, which still requires it).

## `@dcl/catalyst-storage` 4.6.1

Picks up the decompression-bomb cap and path-traversal containment fixes; wires the new `decompressMaxFileSize` option to a `STORAGE_DECOMPRESS_MAX_FILE_SIZE` env var.

## Security hardening

- `POST /entities` enforces configurable upload size/count limits (**HTTP 413**) instead of buffering unbounded request bodies in memory (`MAX_UPLOAD_FILE_SIZE` / `MAX_UPLOAD_FILE_COUNT`).
- Stop logging auth chains/signatures on deployment errors.
- Escape `LIKE` wildcards in the collection-URN pointer queries.
- Warn at startup when `IGNORE_BLOCKCHAIN_ACCESS_CHECKS` disables access validation.

## Testing

- `tsc -b` clean; eslint clean.
- **285 unit tests** pass.
- **295 integration tests** pass (4 pre-existing skips) — the full suite runs every request through the new native-fetch `@dcl/http-server` v2, covering multipart uploads, content serving (incl. range requests), deployments, and synchronization.

## Notes

- `node-fetch` is not fully eliminated from the dependency tree — `@dcl/snapshots-fetcher` still pulls it (pinned to `2.7.0`), plus the dev-only test-client usage. The server's own source is now `node-fetch`-free, and `@dcl/http-server` is no longer a source of it.

---

## Adopt published `@dcl/` HTTP components + endpoint hardening

Now that `@dcl/http-tracer-component` v2 and `@dcl/schema-validator-component` v1 are published with native-fetch (`@dcl/core-commons`) types:

- **Tracing** — switched from `@well-known-components/http-tracer-component` to **`@dcl/http-tracer-component` v2**, which shares the native-fetch `IHttpServerComponent`, so the boundary cast added earlier in this PR is **removed**.
- **Validation** — the `POST /entities/active` body is now validated by **`@dcl/schema-validator-component`** as route middleware (replacing the inline Joi). ⚠️ The 400 response body for invalid input changes to the validator's `{ ok, message, data }` shape (the integration test was updated accordingly).
- **Version bumps** — `@dcl/http-server` → `^2.0.1`, `@dcl/traced-fetch-component` → `^1.0.6`, `@dcl/fetch-component` → `1.0.1`; added `@dcl/http-tracer-component` `^2.0.0` and `@dcl/schema-validator-component` `^1.0.0`.

### Security fixes

- **Fixes #1935** — `POST /entities/active` now caps `ids`/`pointers` at **1000** entries (via the schema-validator's JSON schema `maxItems`), bounding the costly `ANY(...)` query on this public, unauthenticated endpoint.
- **Fixes #1936** — `POST /entities` now caps the AuthChain index at **10** elements, so a crafted `authChain[<huge>][...]` multipart field name can't drive a large parse loop before signature validation.

### Testing

`tsc` clean, eslint clean, **285 unit + 297 integration** tests pass (incl. new tests for both #1935 and #1936).

### Follow-up hardening (completes #1936)

A deeper review found the AuthChain index cap alone didn't fully close the DoS surface: the multipart wrapper bounded file size/count but left **non-file form fields unbounded** (busboy defaults `fields` to `Infinity`, and every field is accumulated into an in-memory object). A request with a huge number of `authChain[i][...]` fields could exhaust memory/CPU regardless of the per-index cap. Added configurable **`maxFields`/`maxFieldSize`** limits to the wrapper (defaults **100 fields / 100 KB per field**, via `MAX_UPLOAD_FIELD_COUNT` / `MAX_UPLOAD_FIELD_SIZE`), rejecting with HTTP 413. Verified legitimate deployments (~a dozen fields) stay well within the limits.

### Multipart security hardening

A dedicated security pass over the multipart upload path found and fixed three more issues:

- **Stream teardown.** The wrapper piped the request body into busboy with `.pipe()`, which doesn't propagate teardown. A client disconnecting mid-upload left busboy emitting neither `finish` nor `error`, so the `await` never settled — **the handler hung and the parser leaked** (repeatable → resource exhaustion). On a limit rejection the body was only unpiped, never destroyed, so the upload wasn't actually aborted. Switched to `stream/promises` `pipeline`, which tears down every stream on any error and rejects with the originating error.
- **Non-multipart Content-Type → 400, not 500.** busboy throws synchronously on a non-multipart `Content-Type`; that escaped as a 500. Now caught and mapped to `InvalidRequestError` (400).
- **Null-prototype field/file maps** so an attacker-controlled part name (`__proto__`, `constructor`) is stored as a plain key rather than touching the object prototype.

Plus a guard on the required `entityId` field (missing field was a `TypeError`/500 instead of a 400).

**Noted, not changed (trade-offs):** cumulative upload size is still only bounded by `maxFiles × maxFileSize` — a total-body cap belongs at the reverse proxy / LB, or could be added as an explicit `maxTotalSize`. The per-file/field 413 messages echo the (attacker-controlled) file/field name; harmless over JSON but worth sanitizing if these get surfaced to a terminal/log viewer.

### Total upload-size cap (cross-ported from worlds-content-server)

Following a cross-check against the worlds-content-server multipart wrapper, ported its total-size guard here (catalyst buffers files in memory, so it matters more): a cumulative `totalBytes` cap across all files+fields plus an up-front `Content-Length` rejection, both wired through a new, tunable `MAX_UPLOAD_TOTAL_SIZE` (default **2 GiB**). The default is deliberately generous because the validator's size check is *per pointer*, so a legitimate multi-parcel scene can be several GB; it bounds the pathological `maxFiles × maxFileSize` case (~300 GB) without rejecting large estate deployments. Streaming uploads to disk (as WCS does) would remove the in-memory exposure entirely and remains the proper follow-up.

The reverse direction (the `__proto__` null-prototype fix and the `pipeline` stream-teardown fix from this PR) was applied to worlds-content-server in decentraland/worlds-content-server#492.

### Malformed `authChain` JSON → 500 (fixed)

A further pass found that `POST /entities` parses the `authChain` form field with `JSON.parse` and then reads `authChain[0].payload` **before** the try/catch. A crafted field that parses to `[]`, a non-array (`42`, `{}`), or `[null]` made that read throw a `TypeError` → uncaught **500** (plus an error log + error-metric increment) on this public, unauthenticated endpoint. `extractAuthChain` now rejects non-array JSON up front and the eth-address read is null-safe, so these all return a clean 400 (verified with 4 cases). The worlds-content-server deploy handler was checked and is not affected (indexed-only, guarded).

### ⚠️ Consumer-facing note: `POST /entities/active` 400 body shape

Adopting `@dcl/schema-validator-component` changes the **error** body for a 400 from the old Joi shape `{ error: "..." }` to `{ ok: false, message: "Invalid JSON body", data: [...] }`. The **happy path (200) is unchanged.** Consumers that parse the 400 body (catalyst-client, js-sdk-toolchain, unity-explorer, asset-bundle-converter, …) should be aware. If this matters, it can be kept backward-compatible with a thin wrapper that re-maps the shape.

`ensureJsonContentType: false` is kept to preserve the previous lenient Content-Type behavior; tightening it is left to a follow-up.


LautaroPetaccio:
Thanks for the thorough review 🙏 Addressed in `2c5863e2`:

**P1 #1 — auth-chain cap bypassed via JSON path** ✅ Fixed. `extractAuthChain` now applies the same `MAX_AUTH_CHAIN_LENGTH` check on the JSON path (after the `Array.isArray` guard). Added an integration test asserting the `"too long"` message specifically (a bare 400 wouldn't prove the cap fired vs. `AuthChain.validate` rejecting garbage links).

**P1 #2 — `Readable.from` on a web stream** ✅ Fixed → `Readable.fromWeb(...)` with a missing-body guard. Note: it wasn't a hard crash on Node 24 (web `ReadableStream` has `Symbol.asyncIterator`, so `Readable.from` works by accident), but `fromWeb` is the correct/explicit API and propagates cancellation cleanly on `destroy()`.

**P1 #3 — `parseInt` NaN silently disables the cap** ✅ Fixed. All five `MAX_UPLOAD_*` vars go through a helper that throws at startup on a non-numeric/negative value, instead of letting `NaN` reach busboy (which treats it as "no limit").

**P2 #4 — double `request.json()`** ℹ️ Not a bug — the validator reads via `context.request.clone().json()` (component.js:56), so the original body is untouched and the handler's `json()` is safe. Added a comment documenting this so the coupling isn't a surprise.

**P2 #5 — overshoot after `destroy()`** ✅ Hardened. A single `abort()` destroys the parser once and the field/file handlers short-circuit on an `aborted` flag, so no further chunks are buffered after a rejection.

**P2 #6 — missing edge-case tests** ✅ Added: JSON auth-chain length cap, no-body (now a 400, not 500 — see below), cumulative field-size cap (previously only files), and duplicate field/file name (last wins). Bonus find while writing these: a malformed/truncated/empty multipart body made busboy reject with a generic error that surfaced as a **500** — now mapped to **400** (size-limit `PayloadTooLargeError` still passes through as 413).

**P2 #7 — 400 response shape changed for `/entities/active`** 📝 Acknowledged and documented in the PR description. Intentional (adopting the shared `@dcl/schema-validator-component`); the happy path is unchanged. Can revert to the old `{ error }` shape with a thin wrapper if consumers need it — flagging for a call.

**P2 #8 — `ensureJsonContentType: false`** 📝 Kept for backward-compat (previous Joi validation didn't enforce Content-Type either). Agree it's worth tightening; will do in a follow-up rather than change behavior in this PR.

PR #1956, opened 2026-08-03
fix: enforce scene base parcel integrity

LautaroPetaccio:
## Summary

- add defense-in-depth validation for scene pointers, parcels, and base
- run the integrity check in the production content-validator adapter
- reject malformed scene metadata before access validation can authorize it
- preserve order-independent pointer and parcel set comparison

## Rollout note

This intentionally tightens validation for new, replayed, or backfilled scene deployments. Entity pointers must equal the unique canonical metadata.scene.parcels set, and metadata.scene.base must be included in that set. Deployments that were historically accepted without satisfying these invariants will be rejected when replayed or revalidated by an upgraded Catalyst.

## Verification

- build and lint passed
- full unit suite passed
- focused scene integrity suite passed with 7 tests
- different pointer and parcel ordering is covered
- diff checks passed

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@dcl/memory-cache-component": "^2.4.3",
     "@dcl/metrics": "^1.0.1",
     "@dcl/schema-validator-component": "^1.0.2",
-    "@dcl/schemas": "^26.5.0",
+    "@dcl/schemas": "^27.0.0",
     "@dcl/snapshots-fetcher": "11.0.0",
     "@dcl/thegraph-component": "^0.1.2",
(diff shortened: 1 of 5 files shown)


LautaroPetaccio:
Coordinated scene-integrity hardening set:

- decentraland/schemas#437
- decentraland/core-libs#60
- decentraland/catalyst#1956
- decentraland/worlds-content-server#521
- decentraland/sdk-multiplayer-server#120
- decentraland/places#856
- decentraland/discovery-server#105
- decentraland/comms-gatekeeper#285
- decentraland/definitions#1640

Please review merge and deployment ordering across this set before rollout.

PR #1957, opened 2026-08-07
chore: bump @dcl/content-validator to 8.0.0

LautaroPetaccio:
## What

Bumps `@dcl/content-validator` from 7.4.2 to 8.0.0 and removes the local scene base-parcel validation that the new version now performs itself.

## Why

v8.0.0 ships two changes (core-libs [#60](https://github.com/decentraland/core-libs/pull/60)):

- Upgrades the validator's runtime and type dependency to `@dcl/schemas@27`. Catalyst already resolves `@dcl/schemas@27.0.0` at the top level, so the previously nested `schemas@26.2.0` copy under `@dcl/content-validator` simply disappears from the tree.
- Adds `sceneParcelsMatchPointersValidateFn` to the stateless scene validations: a scene's base, metadata parcels, and entity pointers must form the same unique canonical parcel identity. This is the check #1956 added locally through `createSceneBaseAwareAccessValidateFn`, upstreamed.

With the check now running inside `sceneValidateFn` (both in `createValidator` and in the staging subset), the local wrapper is redundant, so this PR deletes `scene-base-validation.ts` and its spec, and unwraps the three access-validator call sites in the adapter.

## Behavior notes

- Validation outcomes are unchanged relative to current `main`: the same deployments are accepted and rejected. Only the failure message text and the point in the pipeline where the check runs differ (stateless scene validation instead of the access step), and like the local wrapper the upstream check is timestamp-independent, so the historical-replay semantics introduced by #1956 are preserved.
- The deep imports under `dist/validations/access/...` used by the adapter are unchanged in v8.

## Testing

- Unit suite: 33 suites / 359 tests passing.
- Integration suite (Postgres testcontainer): 35 suites / 306 tests passing, 4 skipped (pre-existing).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


LautaroPetaccio:
Addressed the P2 in d71bc880: added `test/unit/adapters/content-validator/component.spec.ts`, a smoke test that builds the validator through `createContentValidator` on the ignore-blockchain path (access validator forced to `OK`) and asserts a scene whose pointers don't cover the scene parcels — or whose base is outside them — is still rejected by the upstream stateless validation. The matching-identity case is also covered and passes end-to-end, so the rejections are attributable to the scene-identity check specifically.

